PR #2, opened 2021-08-11
Onset indices draft

ikhomyakov:
This is a very rough numpy based draft of onset_date (onset_indices) function. The main idea is to come up with a vectorized implementation. Please note that this implementation returns all onset dates (indices).
If we decide that this draft is worth further development, we can:

* Debug it to make sure it corresponds to the function documentation: http://iridl.ldeo.columbia.edu/dochelp/Documentation/details/index.html?func=onsetDate
* Extend it to handle multi-dimensional ndarrays along any axis
* Wrap it into xarray API
* Think about how to implement multi-chunk version

### DIFF
--- a/onset_maproom/onset_indices.py
+++ b/onset_maproom/onset_indices.py
@@ -52,8 +52,10 @@ def onset_date(
     returns the time delta of first wet day in that wet spell
     the time delta from first day of daily rain
     """
+    # Find wet days
     wet_day = daily_rain > wet_th
 
+    # Find 1st wet day in wet spells length
     first_wet_day = wet_day * 1
     first_wet_day = (
         first_wet_day.rolling(**{time_coord: wet_spell_length})
@@ -61,15 +63,16 @@ def onset_date(
         .argmax("wsl")
     )
 
+    # Find wet spells
     wet_spell = (
         daily_rain.rolling(**{time_coord: wet_spell_length}).sum() >= wet_spell_th
     ) & (wet_day.rolling(**{time_coord: wet_spell_length}).sum() >= min_rainy_days)
 
+    # Find dry spells following wet spells
     dry_day = ~wet_day
     false_start = (
         dry_day.rolling(**{time_coord: dry_spell_length}).sum() == dry_spell_length
     )
-
     # Note that rolling assigns to the last position of the wet_spell
     false_start_ahead = (
         false_start.rolling(**{time_coord: dry_spell_search})
@@ -78,18 +81,12 @@ def onset_date(
         != 0
     )
 
-    wet_day_within_onset = (wet_day) & (
-        (wet_spell & ~false_start_ahead)
-        .rolling(**{time_coord: wet_spell_length})
-        .sum()
-        .shift(**{time_coord: 1 - wet_spell_length})
-        != 0
-    )
-
+    # Create a mask of 1s and nans where onset conditions are met
     # Turns False/True into nan/1
     onset_mask = (wet_spell & ~false_start_ahead) * 1
     onset_mask = onset_mask.where((wet_spell & ~false_start_ahead))
 
+    # Find onset date (or rather last day of 1st valid wet spell)
     # Note it doesn't matter to use idxmax or idxmin,
     # it finds the first max thus the first onset date since we have only 1s and nans
     # all nans returns nan
@@ -98,8 +95,12 @@ def onset_date(
     #    print(first_wet_day.sel(T=onset_delta))
     #    print(first_wet_day[onset_delta])
     #    print(first_wet_day.where(onset_delta == first_wet_day["T"], drop=True))
+    # Apply offsets to return onset date delta
     onset_delta = (
         onset_delta
+        # offset relative position of first wet day
+        # note it doesn't matter to apply max or min
+        # per construction all values are nan but 1
         - (
             wet_spell_length
             - 1
@@ -107,6 +108,7 @@ def onset_date(
                 dim=time_coord
             )
         ).astype("timedelta64[D]")
+        # delta from 1st day of time series
         - daily_rain[time_coord][0]
     ).rename("onset_delta")
     return onset_delta
--- a/onset_maproom/onset_indices.py
+++ b/onset_maproom/onset_indices.py
@@ -1,4 +1,5 @@
 import numpy as np
+import xarray as xr
 from numpy.lib.stride_tricks import sliding_window_view
 np.random.seed(0)
 
@@ -35,6 +36,62 @@ def onset_indices(
     onset_ids = np.unique(first_rainy[onset_ids] + onset_ids)
     return onset_ids
 
+def onset_index(
+    data, data_flag, wlength,      wthresh,       flagged_thresh, fwlength, false_search, coord_index
+):
+    """Finds the first window of size wlength where data exceeds wthresh,
+    with at least flagged_thresh count of flagged data (by data_flag),
+    not followed by a false_window of size fwlength of unflagged data, for the following false_search indices
+    returns the index of first flagged data in that window
+    (how much wood would a woodchuck chuck if a woodchuck could chuck wood?)
+    """
+    flagged = data > data_flag
+
+    window = (
+        data.rolling(**{coord_index: wlength}).sum() > wthresh
+    ) & (
+        flagged.rolling(**{coord_index: wlength}).sum() > flagged_thresh
+    )
+
+    unflagged = ~flagged
+    false_window = (
+        unflagged.rolling(**{coord_index: fwlength}).sum() == fwlength
+    )
+
+#Note that rolling assigns to the last position of the window
+    false_window_ahead = (
+        false_window.rolling(**{coord_index: false_search}).sum().shift(**{coord_index: false_search*-1}) != 0
+    )
+
+    flagged_within_onset = (
+        flagged
+    ) & (
+        (window & ~false_window_ahead).rolling(**{coord_index: wlength}).sum().shift(**{coord_index: 1-wlength}) != 0
+    )
+
+#Turns False/True into nan/1
+    onset_mask = flagged_within_onset*1
+    onset_mask = onset_mask.where(flagged_within_onset)
+
+#Note it doesn't matter to use idxmax or idxmin, this finds the first max/min thus the first onset date since we have only 1s and nans
+#all nans returns nan
+    onset_id = onset_mask.idxmax(dim=coord_index)
+    return onset_id
+
+def run_test():
+  import pyaconf
+  import os
+  from pathlib import Path
+  import calc
+
+  CONFIG = pyaconf.load(os.environ["CONFIG"])
+  DR_PATH = CONFIG["daily_rainfall_path"]
+  RR_MRG_ZARR = Path(DR_PATH)
+  rr_mrg = calc.read_zarr_data(RR_MRG_ZARR)
+  rr_mrg = rr_mrg.sel(T=slice("2000-01-01", "2000-12-31"))
+  print(
+    onset_index(rr_mrg.precip, 1, 3, 20, 1, 7, 21, "T")
+  )
 
 rain = np.random.rand(1000)
 xs = onset_indices(
--- a/onset_maproom/onset_indices.py
+++ b/onset_maproom/onset_indices.py
@@ -90,6 +90,9 @@ def onset_date(
     # it finds the first max thus the first onset date since we have only 1s and nans
     # all nans returns nan
     onset_delta = onset_mask.idxmax(dim=time_coord)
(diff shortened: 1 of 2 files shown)


ikhomyakov:
I am wondering if there is a better way to compute the first wet day within each window?

    first_rainy = np.apply_along_axis(
        lambda v: np.where(v)[0][0] if len(np.where(v)[0]) != 0 else running_days,
        1,
        sliding_window_view(rainy, window_shape=running_days),
    )

onset_maproom/onset_indices.py
```diff
@@ -0,0 +1,50 @@
...
+    dry_spells = (
+        np.convolve(~rainy, np.ones(dry_days, dtype=int), "valid") == dry_days
+    )
+    dry_spells_ahead = (
+        np.convolve(dry_spells, np.ones(running_days + dry_spell, dtype=int), "valid")
+        != 0
+    )
+    onset_ids = np.where((wet_spells[:n]) & (~dry_spells_ahead))[0]
+    onset_ids = np.unique(first_rainy[onset_ids] + onset_ids)
+    return onset_ids
```
aaron-kaplan, line 36: I think this function should return `onset_ids[0]`, not `onset_ids`. The onset of the rainy season is the *first* day that meets the criteria, not a list of days.

If I'm right about that, then there's no point calculating the first wet day in every window (the step I pointed out might be slow about). You can wait until you get to the end, and then find the first wet day in the *first* qualifying window, which is cheaper.

onset_maproom/onset_indices.py
```diff
@@ -0,0 +1,50 @@
...
+    n = rain.shape[0]
+    rain = np.pad(
+        rain,
+        (0, running_days + dry_spell + dry_days - 2),
+        mode="constant",
+        constant_values=0,
+    )
+    rainy = rain > rainy_day
+    first_rainy = np.apply_along_axis(
+        lambda v: np.where(v)[0][0] if len(np.where(v)[0]) != 0 else running_days,
```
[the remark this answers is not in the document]
ikhomyakov: I dont like it euther. Would be nice to find a fully vectorized form.

onset_maproom/onset_indices.py
ikhomyakov: I think you are right. I've  just been trying to find a more generic vectorized form

remicousin:
Adding the xarray/rolling version. It also takes care of finding the 1st flagged/wet day within the window/wet_spells. It's done by checking if there is a window after a wet day, same logic as checking whether there is a dry spell after a window, making use of the fact that rolling assigns to the last position of the rolling window

onset_maproom/onset_indices.py
```diff
@@ -35,6 +36,62 @@ def onset_indices(
...
+
+#Note that rolling assigns to the last position of the window
+    false_window_ahead = (
+        false_window.rolling(**{coord_index: false_search}).sum().shift(**{coord_index: false_search*-1}) != 0
+    )
+
+    flagged_within_onset = (
+        flagged
+    ) & (
+        (window & ~false_window_ahead).rolling(**{coord_index: wlength}).sum().shift(**{coord_index: 1-wlength}) != 0
```
aaron-kaplan, line 69: I'm having trouble understanding this line.  I understand that `window & ~false_window_ahead` is True on dates that are at the end of a wet spell and not followed soon by a dry spell. I understand that the shift moves the marker from the end of the wet spell to the start. But I don't understand how `rolling` is being used.

The first few values of the rolling sum are always `NaN` because there isn't a full window to look at, and `NaN != 0`, so this expression comes out to True on some dates that aren't anywhere near a wet spell.

onset_maproom/onset_indices.py
```diff
@@ -35,6 +36,62 @@ def onset_indices(
...
+    ) & (
+        flagged.rolling(**{coord_index: wlength}).sum() > flagged_thresh
+    )
+
+    unflagged = ~flagged
+    false_window = (
+        unflagged.rolling(**{coord_index: fwlength}).sum() == fwlength
+    )
+
+#Note that rolling assigns to the last position of the window
```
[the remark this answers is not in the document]
remicousin: fixed

onset_maproom/onset_indices.py
```diff
@@ -35,6 +36,62 @@ def onset_indices(
...
 
+def onset_index(
+    data, data_flag, wlength,      wthresh,       flagged_thresh, fwlength, false_search, coord_index
+):
+    """Finds the first window of size wlength where data exceeds wthresh,
+    with at least flagged_thresh count of flagged data (by data_flag),
+    not followed by a false_window of size fwlength of unflagged data, for the following false_search indices
+    returns the index of first flagged data in that window
+    (how much wood would a woodchuck chuck if a woodchuck could chuck wood?)
+    """
```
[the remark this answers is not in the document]
remicousin: Personally, I don't believe there is another use case for this. Even cessation date, I don't think it can apply. It can inspire a cessation date function and other certainly though. Assuming input is daily data will greatly simplify things. In particular the last step to find the first wet day within the first valid wet spell. I could use arg/idxmin to get the relative position in the group and then use that as an offset to the final result that finds the position of the last day of the wet spell. I actually did that originally until I realized I did assume daily data. What I did didn't work on indices because I ran into cases where idxmin and argmin respectively gave a date and a NaN when applied to the same input. And once idxmin is used, we are in the coordinates domain and no more in indices domain.
Anyway, my point is that generalizing brings constrains that can seriously change the philosophy of the code so I'd suggest either we stick to general, or we go for an onset_date function that assume daily rainfall as input.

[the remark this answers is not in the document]
remicousin: For the little history about "flag." In Ingrid, flag functions return 1 or 0 based on a condition. And in many datasets we are used to manipulate, a flag variable is often a quality control key that can be more subtle than binary, e.g. 'observed', 'estimated', 'filled', interpolated', or more specialized term. So probably flag should not be connoted "bad."

onset_maproom/onset_indices.py
```diff
@@ -35,6 +36,62 @@ def onset_indices(
...
+    flagged = data > data_flag
+
+    window = (
+        data.rolling(**{coord_index: wlength}).sum() > wthresh
```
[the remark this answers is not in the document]
remicousin: correct and fixed

onset_maproom/onset_indices.py
remicousin: Assuming we are looking at a window of 3-day long wet spell, to simplify explanation, the first wet day can be the 3rd, 2nd or 1st day of any of the 3 spells it could be in. So the rolling.sum allows to do a "OR" on this given day being in one of those spells. Then shift moves the marker etc...

As for the 2nd comment, I did get onset dates way away from the beginning of times so I haven't checked but I suspect the 1-wlength shift removes the heading NaNs you are talking about.

remicousin: This could be a lot more simple if assuming input to be daily rainfall

onset_maproom/onset_indices.py
```diff
@@ -0,0 +1,115 @@
...
+    coord_index,
+):
+    """Finds the first window of size wlength where data exceeds wthresh,
+    with at least flagged_thresh count of flagged data (by data_flag),
+    not followed by a false_window of size fwlength of unflagged data,
+    for the following false_search indices
+    returns the index of first flagged data in that window
+    (how much wood would a woodchuck chuck if a woodchuck could chuck wood?)
+    """
+    flagged = data > data_flag
```
aaron-kaplan, line 54: Another `>` that should be `>=`?

onset_maproom/onset_indices.py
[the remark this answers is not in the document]
aaron-kaplan: The groupby function (your other PR) already assumes daily data, so no point making this one more general yet, especially since of the two components, this one is less likely to be reusable.

onset_maproom/onset_indices.py
aaron-kaplan: If assuming daily rainfall is going to allow you to simplify it, then I won't spend time trying to understand this.

Possibly related, although I'm not sure because I still don't understand this entirely... you define a wet spell as a period of length x containing at least y wet days and total rainfall of at least z, and then you find the first wet day of the first wet spell. I wonder if it would be simpler to add to the definition of a wet spell the requirement that it must start with a wet day.

onset_maproom/onset_indices.py
remicousin: not that one because a user might want to be able to define a wet day as anything above 0.

onset_maproom/onset_indices.py
remicousin: That would change things a bit as an earlier non-wet day could contribute to making a wet spell that would otherwise not be reaching the wet spell threshold. That's probably not many cases but possible. We are probably in that situation because that piece came as an afterthought. Once I wrote the algo to find the first wet spell, I was going to make that period of time to be the onset date, but then I was told to pick the 1st wet day within it.

But so... are we agreeing to move away from the general case and assume daily rainfall as input?

onset_maproom/onset_indices.py
aaron-kaplan: Makes sense, and now I see that the Ingrid [docs](https://iridl.ldeo.columbia.edu/dochelp/Documentation/details/index.html?func=onsetDate) say that quite explicitly.

I was misled by a different part of the Ingrid docs:

> running window that totals runningTotals (in prcp units) or more of precipitation

Now I see that this is about the total precip over the window, whereas this line is about the precip for a single day. But isn't it confusing that one precip threshold defines an open interval and the other a closed interval?

onset_maproom/onset_indices.py
aaron-kaplan: I see. You're right, it wouldn't be equivalent.

Yes, let's assume daily rainfall.

onset_maproom/onset_indices.py
remicousin: yes. That is the definition that was given... That is to say the threshold to validate wet to be inclusive...

remicousin:
I pushed a version that assumes daily rainfall.

onset_maproom/onset_indices.py
```diff
@@ -0,0 +1,151 @@
...
+        onset_delta
+        # offset relative position of first wet day
+        # note it doesn't matter to apply max or min
+        # per construction all values are nan but 1
+        - (
+            wet_spell_length
+            - 1
+            - first_wet_day.where(first_wet_day[time_coord] == onset_delta).max(
+                dim=time_coord
+            )
```
aaron-kaplan, line 104: 102-104 seems more complicated than necessary. If we use `argmax` instead of `idxmax` in line 93, then
```
first_wet_day.where(first_wet_day[time_coord] == onset_delta).max(
                dim=time_coord
            )
```
can be rewritten
```
first_wet_day.isel(**{time_coord: onset_delta})
```
or something like that. (It's not quite that simple, because `onset_delta` is also referenced in line 95, but hopefully you get the idea. If not, then skip it, it's not that important right now.)

aaron-kaplan (commented):
Just about there. Only some minor cosmetic comments.

If there are > that should be >= in Igor's implentation, could you fix those before we merge?

aaron-kaplan:
> Just about there. Only some minor cosmetic comments.
> 
> If there are > that should be >= in Igor's implentation, could you fix those before we merge?

Or we can separate the two imlementations onto separate branches. I want to move ahead with the xarray implementation for now, but it would be good to keep track of the numpy version somewhere in case we want to try it in the future.

remicousin:
I am working on these things now. Igor was also suggesting not to merge his branch. So maybe once we're good here, I move the xarray work to another branch and we let that one be.

remicousin:
102-104 seems more complicated than necessary. If we use argmax instead of idxmax in line 93, then

first_wet_day.where(first_wet_day[time_coord] == onset_delta).max(
                dim=time_coord
            )

can be rewritten

first_wet_day.isel(**{time_coord: onset_delta})

or something like that. (It's not quite that simple, because onset_delta is also referenced in line 95, but hopefully you get the idea. If not, then skip it, it's not that important right now.)


Doesn't work. I pushed some prints that illustrate that... idxmax and argmax applied to same input respectively return a date and all nan in the given example. I ran into that previously what causes me to use great detours to avoid argmax...

aaron-kaplan:
OK, here's the reason idxmax works but argmax doesn't. idxmax returns datetime64s, whereas argmax returns ints. datetime64 has a missing value indicator NaT (Not a Time), whereas int has no missing value indicator. So if there's an X, Y point where the variable is NaN for all values of T (e.g. `onset_mask.isel(X=0, Y=0)`), idxmax returns NaT for that point, but there's no appropriate value for argmax to return so it has to fail.

We could probably work around this using a mask, but I don't think it would come out any simpler than what you've done. So put your implementation in a new PR and I'll approve it.

PR #3, opened 2021-08-17
first try at a function to break 1D-daily into 2D for seasons

remicousin:
Also includes seasonal_sum to illustrate how could be used to make "seasonal" functions.

The original plan was to make a generic function that will split 1D daily data into 2D daily data by days and years and select a subset of days defining a season. Thus allowing to perform analysis on the days dimension to obtain yearly series of a seasonal quantity and thus be able to reproduce a lot of popular Ingrid functions ([e.g.](http://iridl.ldeo.columbia.edu/dochelp/Documentation/details/index.html?func=flexseasonalfreqGT) and including onset date).

It turns out that xarray offers a [grouping function](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.groupby.html) and more specifically by [bins](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.groupby_bins.html) that then allow operations by group. This seems nicer to me as there is then no need to create a new dimension, worry about indexing and bins can have different sizes (making leap year management easier). The idea being to give the edges of the season of interest, group, then drop every other group and then that can be fed to any other operator.

So I went with that. Because bins are defined by their edges, I made the inputs for the season be start and end rather than start and length. Now that fits better the groupby_bins function and it's typically easier for a user to think of seasons in these terms. That may mean changing onset date function inputs accordingly, or maybe cover the 2 cases (ie also offer inputs to be start and length).

Now it also turns out that it's not easy to drop groups from a grouped object. Or it was but someone [complained](https://github.com/pydata/xarray/issues/1019) about automatically dropped empty bins so this [PR](https://github.com/pydata/xarray/pull/1027) brings them back automatically and recommends using dropna to actually drop. But one can do dropna only after applying an operator to the grouped data, so that is something to be aware of when creating something like seasonal_sum.

That makes me doubt whether this is a good route or not. The grouping thing is seducing but the dropping business not so much. Other options could be:

- use the pandas versions of those things ([groupby](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.groupby.html), [cut](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.cut.html)) and use [filter](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.filter.html) for dropping;
- go a totally different route, TBD

There is a run_test_season_stuff() to help with testing and one can run calc.py with
CONFIG=config-sample.yaml python calc.py

### DIFF
--- a/onset_maproom/calc.py
+++ b/onset_maproom/calc.py
@@ -95,11 +95,6 @@ def daily_tobegroupedby_season(
     if start_day == 29 and start_month == 2:
         start_day = 1
         start_month = 3
-    end_day2 = end_day
-    end_month2 = end_month
-    if end_day == 29 and end_month == 2:
-        end_day2 = 1
-        end_month2 = 3
     # Find seasons edges
     start_edges = daily_data[time_coord].where(
         lambda x: (x.dt.day == start_day) & (x.dt.month == start_month),
@@ -108,22 +103,22 @@ def daily_tobegroupedby_season(
     if end_day == 29 and end_month == 2:
         end_edges = daily_data[time_coord].where(
             lambda x: (
-                (x + xr.DataArray([1]).astype("timedelta64[D]")).dt.day.squeeze(
+                (x + np.timedelta64(1, "D")).dt.day.squeeze(
                     drop=True
                 )
-                == end_day2
+                == 1 
             )
             & (
-                (x + xr.DataArray([1]).astype("timedelta64[D]")).dt.month.squeeze(
+                (x + np.timedelta64(1, "D")).dt.month.squeeze(
                     drop=True
                 )
-                == end_month2
+                == 3
             ),
             drop=True,
         )
     else:
         end_edges = daily_data[time_coord].where(
-            lambda x: (x.dt.day == end_day2) & (x.dt.month == end_month2),
+            lambda x: (x.dt.day == end_day) & (x.dt.month == end_month),
             drop=True,
         )
     # Drop dates outside very first and very last edges
--- a/onset_maproom/calc.py
+++ b/onset_maproom/calc.py
@@ -111,8 +111,6 @@ def daily_tobegroupedby_season(daily_data, start_day, start_month, end_day, end_
 def seasonal_sum(daily_data, start_day, start_month, end_day, end_month, min_count=None, time_coord="T"):
     """Calculates seasonal totals of daily data in season defined by day-month edges
     """
-    #It turns out that having daily_groupby_season concatenate only every other group is not enough to drop entired the undesired seasons
-    #sum will return NaN for these so need to use dropna to clean up
     grouped_daily_data = daily_tobegroupedby_season(daily_data, start_day, start_month, end_day, end_month)
     seasonal_data = (
       grouped_daily_data[daily_data.name]
--- a/onset_maproom/calc.py
+++ b/onset_maproom/calc.py
@@ -2,15 +2,30 @@ import numpy as np
 import pandas as pd
 import xarray as xr
 
-#Date Reading functions
+# Date Reading functions
+
 
 def read_zarr_data(zarr_path):
     zarr_data = xr.open_zarr(zarr_path)
     return zarr_data
 
-#Growing season functions
 
-def onset_date(daily_rain, early_start_day, early_start_month, search_days, rainy_day, running_days, running_total, min_rainy_days, dry_days, dry_spell, time_coord="T"):
+# Growing season functions
+
+
+def onset_date(
+    daily_rain,
+    early_start_day,
+    early_start_month,
+    search_days,
+    rainy_day,
+    running_days,
+    running_total,
+    min_rainy_days,
+    dry_days,
+    dry_spell,
+    time_coord="T",
+):
     """Function reproducing Ingrid onsetDate function
     http://iridl.ldeo.columbia.edu/dochelp/Documentation/details/index.html?func=onsetDate
     with the exception that:
@@ -19,25 +34,30 @@ def onset_date(daily_rain, early_start_day, early_start_month, search_days, rain
     output is now the timedelta with each year's earlyStart (as opposed to the date itself)
     """
     onset_date = daily_rain[
-      (daily_rain[time_coord].dt.day==early_start_day)
-      &
-      (daily_rain[time_coord].dt.month==early_start_month)
+        (daily_rain[time_coord].dt.day == early_start_day)
+        & (daily_rain[time_coord].dt.month == early_start_month)
     ]
     onset_date = xr.DataArray(
-      data=np.random.randint(0, high=search_days, size=onset_date.shape).astype('timedelta64[D]'),
-      dims=onset_date.dims,
-      coords=onset_date.coords,
-      attrs=dict(
-          description="Onset Date",
-      ),
+        data=np.random.randint(0, high=search_days, size=onset_date.shape).astype(
+            "timedelta64[D]"
+        ),
+        dims=onset_date.dims,
+        coords=onset_date.coords,
+        attrs=dict(
+            description="Onset Date",
+        ),
     )
-   #Tip to get dates from timedelta  early_start_day
-   #  onset_date = onset_date[time_coord] + onset_date  
+    # Tip to get dates from timedelta  early_start_day
+    #  onset_date = onset_date[time_coord] + onset_date
     return onset_date
+
 
 # Time functions
 
-def daily_tobegroupedby_season(daily_data, start_day, start_month, end_day, end_month, time_coord="T"):
+
+def daily_tobegroupedby_season(
+    daily_data, start_day, start_month, end_day, end_month, time_coord="T"
+):
     """Returns dataset ready to be grouped by with:
     the daily data where all days not in season of interest are dropped
     season_starts: an array where the non-dropped days are indexed by the first day of their season -- to use to groupby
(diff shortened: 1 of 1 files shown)


remicousin:
And of course there could be other problems that the ones I laid down in introduction

ikhomyakov:
> That makes me doubt whether this is a good route or not. The grouping thing is seducing but the dropping business not so much. Other options could be:
> 
@remicousin, I also doubt that the grouping is a good approach. Let me look into this more and I will try to update by Monday.

remicousin:
> > That makes me doubt whether this is a good route or not. The grouping thing is seducing but the dropping business not so much. Other options could be:
> 
> @remicousin, I also doubt that the grouping is a good approach. Let me look into this more and I will try to update by Monday.

ok! 

That label-grouping concept seems to be all over the place though... Maybe I am not using it right... or maybe we need a more sophisticated version... or maybe it's not the way to go at all...

aaron-kaplan:
My first thought for extracting a single season would be to use `where` with `drop=True` instead of `groupby_bins`. But if you're planning to generalize beyond the current use case to support multiple seasons, `groupby_bins` might be the way to go. Did you consider and reject the option of using `where`?

onset_maproom/calc.py
```diff
@@ -33,6 +35,80 @@ def onset_date(daily_rain, early_start_day, early_start_month, search_days, rain
...
+# Time functions
+
+def daily_groupby_season(daily_data, start_day, start_month, end_day, end_month, time_coord="T"):
+  """Groups daily data by yearly seasons from day-month edges of season
+  Can then apply functions on each group to create yearly time series of seasonal quantities
+  If starting day-month is 29-Feb, uses 1-Mar.
+  If ending day-month is 29-Feb, uses 1-Mar and triggers the option to open the right edge of the Interval.
+  That means that the last day included in the season will be 29-Feb in leap years and 28-Feb otherwise
+  """
+#Deal with leap year cases
```
aaron-kaplan, line 47: Please indent comment lines to the same level as the code they apply to. These unindented comments make it harder to see where functions and if/then/else blocks start and end.

onset_maproom/calc.py
```diff
@@ -33,6 +35,80 @@ def onset_date(daily_rain, early_start_day, early_start_month, search_days, rain
...
+  )
+#Second pass to recreate the group labels that concat lost
+  daily_groubedby_season = daily_groubedby_season.groupby_bins(time_coord, season_bins, right=right_bool)
+  return daily_groubedby_season
+
+def seasonal_sum(daily_data, start_day, start_month, end_day, end_month, min_count=None, time_coord="T"):
+  """Calculates seasonal totals of daily data in season defined by day-month edges
+  """
+#It turns out that having daily_groupby_season concatenate only every other group is not enough to drop entired the undesired seasons
+#sum will return NaN for these so need to use dropna to clean up
```
aaron-kaplan, line 78: Seems like `dropna` might sometimes drop more than you intended. NaNs in the original data might result in legitimate NaNs for the season of interest, and we shouldn't drop those.  If we're going to stick with `groupby_bins` (rather than `where` as I suggested in another comment), we should try to eliminate the unwanted seasons in a more principled way. I won't spend more time thinking about that until you reply about `where`.

remicousin: Correct: dropna could drop more legitimate NAs. In Ingrid, there are clearly 2 distinct worlds between regularly spaced grids and not; but in Python/Xarray, with the labelling system, it was unclear to me. Now to the where question

aaron-kaplan:
Rémi and I had a convo and decided to try using `where` to drop the dates that are out of season, followed by `groupby_bins` to do the aggregation over each season. There will then be one group per year rather than two, eliminating the complicated and not entirely correct concatenation and `dropna` steps.

remicousin:
Latest commits improves on this. However, can't use groupby_bins, even after dropping days outside seasons, because the empty interval for the wrong seasons is still considered. Thus used groupby according to a group description. If that is the right route, then need to figure out how to give "group" better metadata (which is a bit sad because groupby_bins was creating nice labels -- not that I am sure they could be reused in a meaningful manner, though...)

remicousin:
In this last commit I also attempted to clean up: code formatting; naming conventions... but there might be more to do.

onset_maproom/calc.py
```diff
@@ -33,6 +35,123 @@ def onset_date(daily_rain, early_start_day, early_start_month, search_days, rain
...
+  #It turns out that having daily_groupby_season concatenate only every other group is not enough to drop entired the undesired seasons
+  #sum will return NaN for these so need to use dropna to clean up
+  seasonal_data = (
+    daily_tobegroupedby_season(daily_data, start_day, start_month, end_day, end_month)[daily_data.name]
+    .groupby(daily_tobegroupedby_season(daily_data, start_day, start_month, end_day, end_month)["seasons_starts"])
+    .sum(dim=time_coord, skipna=True, min_count=min_count)
+    .rename({"seasons_starts": time_coord})
+  )
+  seasons_ends = (
+    daily_tobegroupedby_season(daily_data, start_day, start_month, end_day, end_month)["seasons_ends"]
```
aaron-kaplan, line 123: Unless I'm missing something, you're calling `daily_tobegroupedby_season` three times with the same arguments. That slows down the program (it does the same work three times) and increases the chance of typos. Call it once and assign the result to a variable, then use that variable three times.

onset_maproom/calc.py
```diff
@@ -33,6 +35,123 @@ def onset_date(daily_rain, early_start_day, early_start_month, search_days, rain
...
+  daily_data = daily_data.where(seasons_starts, drop=True)
+  #Group by season with groups labeled by seasons_starts
+  daily_tobegroupedby_season = xr.merge([daily_data, seasons_starts, seasons_ends])
+  return daily_tobegroupedby_season
+
+def seasonal_sum(daily_data, start_day, start_month, end_day, end_month, min_count=None, time_coord="T"):
+  """Calculates seasonal totals of daily data in season defined by day-month edges
+  """
+  #It turns out that having daily_groupby_season concatenate only every other group is not enough to drop entired the undesired seasons
+  #sum will return NaN for these so need to use dropna to clean up
```
aaron-kaplan, line 115: Is this comment out of date? I don't see the dropna it refers to.

aaron-kaplan (changes requested):
You're doing some weird shit with array broadcasting that I find difficult to follow. I made it most of the way through but gave up near the end. If you think the result is right, let's move ahead for now. I would like to get the onset date function working, write a comprehensive set of automated tests for it (I will help with that), and then we can refine the implementation with confidence because the tests will tell us if we break anything.

onset_maproom/calc.py
```diff
@@ -33,6 +35,123 @@ def onset_date(daily_rain, early_start_day, early_start_month, search_days, rain
...
+  )
+  daily_data = daily_data.sel(**{time_coord: slice(start_edges[0],end_edges[-1])})
+  start_edges = daily_data[time_coord].where(
+    ((daily_data[time_coord].dt.day==start_day) & (daily_data[time_coord].dt.month==start_month)),
+    drop=True
+  )
+  end_edges = daily_data[time_coord].where(
+    ((daily_data[time_coord].dt.day==end_day2) & (daily_data[time_coord].dt.month==end_month2)),
+    drop=True
+  )
```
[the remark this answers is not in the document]
remicousin: no because first I get the edges that belong to the whole time domain of the data; then I restrict the data to the very first and very last edges; then I regenerate the list of start and end edges. That ensures they are the same size, and that they are "paired in order. That is not a given before restricted the data to the very first and very last edges. e.g. data starts 1 Jan 1981; seasons are (4 Nov - 28 Feb) then 28 Feb 1981 is not the first start_edge: it's discarded.

onset_maproom/calc.py
```diff
@@ -33,6 +35,123 @@ def onset_date(daily_rain, early_start_day, early_start_month, search_days, rain
...
+  #Creates array of edges of the season that will form the bins
+  seasons_edges = xr.concat([start_edges, end_edges], "T_out", join="override")
```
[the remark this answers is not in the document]
remicousin: That's because I need the start and end of each season "paired" against a common dimension. start/end_edges are on different dimensions that are their very values.

onset_maproom/calc.py
```diff
@@ -33,6 +35,123 @@ def onset_date(daily_rain, early_start_day, early_start_month, search_days, rain
...
+    ) & (
+      daily_data[time_coord] < seasons_edges.isel(T_out=1).rename({time_coord: "group"})
+    )
+  else:
+    days_in_season = (
+      daily_data[time_coord] >= seasons_edges.isel(T_out=0).rename({time_coord: "group"})
+    ) & (
+      daily_data[time_coord] <= seasons_edges.isel(T_out=1).rename({time_coord: "group"})
+    )
+  seasons_starts = daily_data[time_coord].where(days_in_season)
```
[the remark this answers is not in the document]
remicousin: it makes a 2D array of dates values that depend on the full time T dimension and the n groups (4 years here). There are date values where the date is within the right group/year and within the seasons edges. And groups has the advantage of being indexed by the day that starts the season. Thus, I can use where to replace, in each groups, the non NaT values by the value of group dimension (the day that starts that season); and then use min (or max really) to squeeze the groups dimension that has no more purpose now and get back to a 1D that depends on T and where all days that are within a season have for value the first day of the season. And I can use that to groupby so that my groupedby dimension is the one I want: the first day of each season, instead of meaningless number 1 to n.
So basically the telescoping from 1D to 2D allows me to treat the groups of days within each season differently, as I need to.

onset_maproom/calc.py
remicousin: is now removed

onset_maproom/calc.py
remicousin: yes. Is my recent change more proper?

onset_maproom/calc.py
```diff
@@ -33,6 +35,123 @@ def onset_date(daily_rain, early_start_day, early_start_month, search_days, rain
...
+  #Deal with leap year cases
+  if start_day == 29 and start_month == 2 :
+    start_day = 1
+    start_month = 3
+  end_day2 = end_day
+  end_month2 = end_month
+  if end_day == 29 and end_month == 2 :
+    end_day2 = 1
+    end_month2 = 3
+  #Drop date outside very first and very last edges -- this ensures we get complete seasons with regards to edges, later on
```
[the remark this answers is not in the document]
remicousin: done

onset_maproom/calc.py
```diff
@@ -2,14 +2,16 @@
 import pandas as pd
 import xarray as xr
 
+#Date Reading functions
+
 def read_zarr_data(zarr_path):
   zarr_data = xr.open_zarr(zarr_path)
```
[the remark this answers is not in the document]
remicousin: switch to 4-space indent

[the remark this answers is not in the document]
aaron-kaplan: I still see a lot of lines that start with six spaces instead of eight.

What editor are you using? It probably has a setting to do this automatically. Black will also do it for you. You'll have to install it in your conda environment first, with `conda install -c conda-forge black`.

onset_maproom/calc.py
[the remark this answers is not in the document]
aaron-kaplan: I see.

If I come back to this code in a few months I'll probably make the same remark, forgetting that we had this conversation. A comment might help head that off.

When you use the same code in multiple places, it's often beneficial to factor it out into a function and call that function each time, rather than copy-pasting the whole thing. When you copy-paste, you introduce the possibility that someone will edit one of them but forget to edit the other.

onset_maproom/calc.py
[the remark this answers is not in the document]
aaron-kaplan: Not positive I understand what you're saying, but what you want is to give `end_edges` the same time coordinate as `start_edges`, it seems simpler/clearer to do something like `end_edges.assign_coords(T=start_edges["T"])`, rather than combining them with `concat` only to split them apart again.

onset_maproom/calc.py
aaron-kaplan: yes.

onset_maproom/calc.py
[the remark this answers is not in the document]
aaron-kaplan: I sort of half understand. Is there documentation that says what it means to use a 2D array to select elements from a 1D array, or is this just obvious to people who are used to working with array broadcasting?

I imagine there's another way to do it that (a) is easier for me to understand and (b) doesn't require a quadratic amount of memory, but we don't need to find it right now. If this works, let's go with it.

onset_maproom/calc.py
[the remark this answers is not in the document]
remicousin: Maybe make this

daily_data[time_coord].where(
    ((daily_data[time_coord].dt.day==start_day) & (daily_data[time_coord].dt.month==start_month)),
    drop=True
  )

a function of daily data, time_coord, start_day and start_month?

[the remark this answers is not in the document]
aaron-kaplan: Instead of rebuilidng the start_edges and end_edges arrays from scratch, what if you did this?
```
    start_edges = start_edges.sel(**{time_coord: slice(start_edges[0],end_edges[-1])})
```
and likewise for `end_edges`? That eliminates the copy-paste, and expresses the intent more clearly (in my subjective opinion).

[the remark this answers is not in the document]
remicousin: I'll try that...

onset_maproom/calc.py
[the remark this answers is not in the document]
remicousin: balcked

onset_maproom/calc.py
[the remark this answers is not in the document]
remicousin: done

onset_maproom/calc.py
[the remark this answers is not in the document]
remicousin: where mentions broadcasting quite a bit. I am 100% sure what it means though so not sure I interpret correctly.
http://xarray.pydata.org/en/stable/generated/xarray.DataArray.where.html#xarray.DataArray.where
https://numpy.org/doc/stable/reference/generated/numpy.where.html#numpy.where

I did that intuitively though...

Another way is to return to the cumsum trick then the group values [1...12...23...34...4...] turn out to be the positions (-1) of the dates I want in their place from start_edges. Is there a way to replace them by transitivity or composition? something like seasons_start = start_edges([1...12...23...34...4...])? I think I tried that but failed but that may just be my python clumsiness. So if you tell me there should be a way, I can try again.

onset_maproom/calc.py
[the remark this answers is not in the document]
remicousin: done

onset_maproom/calc.py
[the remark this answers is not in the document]
remicousin: I am onto something

onset_maproom/calc.py
```diff
@@ -2,51 +2,198 @@
...
+
+# Time functions
+
+
+def daily_tobegroupedby_season(
+    daily_data, start_day, start_month, end_day, end_month, time_coord="T"
+):
+    """Returns dataset ready to be grouped by with:
+    the daily data where all days not in season of interest are dropped
+    season_starts: an array where the non-dropped days are indexed by the first day of their season -- to use to groupby
```
aaron-kaplan, line 63: Today I learned that while black generally enforces a maximum line length of 88 characters, it makes an exception for docstrings. We have to keep lines like this short manually. Your editor might be able to fill paragraphs automatically.

onset_maproom/calc.py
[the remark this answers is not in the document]
remicousin: another way found and implemented in last commit

onset_maproom/calc.py
remicousin: fixed manually. I am using vi on devi

onset_maproom/calc.py
```diff
@@ -2,51 +2,219 @@
...
+        end_day2 = 1
+        end_month2 = 3
+    # Find seasons edges
+    start_edges = daily_data[time_coord].where(
+        (
+            (daily_data[time_coord].dt.day == start_day)
+            & (daily_data[time_coord].dt.month == start_month)
+        ),
+        drop=True,
+    )
```
aaron-kaplan, line 91: A stylistic comment that's not essential; take it or leave it.

Here's some shorthand you could use here and elsewhere to further reduce the amount of copy-paste:
```python
start_edges = daily_data[time_coord].where(
    lambda x: x.dt.day == start_day & x.month = start_month,
    drop=True
)
```

remicousin: how does that work? How does it know that x is daily_data[time_coord]?

aaron-kaplan: First of all,
```
lambda x: blahblah
```
 is equivalent to writing `myfunc` when `myfunc` is defined as
```
def myfunc(x):
    return blahblah
```
i.e. the value of the lambda expression is a function.

The [documentation](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.where.html) for `DataArray.where` says:

> If [the first argument is] a callable [e.g. a function], it must expect this object as its only parameter.

So if `da` is a `DataArray`, then `da.where(lambda x: x > 1)` calls the lambda function with `x` set to `da`, and expects it to return a `DataArray` of booleans.

onset_maproom/calc.py
```diff
@@ -27,11 +27,12 @@ def onset_date(
     time_coord="T",
 ):
     """Function reproducing Ingrid onsetDate function
-    http://iridl.ldeo.columbia.edu/dochelp/Documentation/details/index.html?func=onsetDate
+    http://iridl.ldeo.columbia.edu/dochelp/Documentation/details/index.html
```
aaron-kaplan, line 30: Changing a URL to keep it from going over 88 characters is probably being overzealous.

remicousin: I don't know what is at stake here...

onset_maproom/calc.py
```diff
@@ -2,51 +2,225 @@
...
+    if end_day == 29 and end_month == 2:
+        days_in_season = (
+            daily_data[time_coord] >= start_edges.rename({time_coord: "group"})
+        ) & (daily_data[time_coord] < end_edges.rename({time_coord: "group"}))
+    else:
+        days_in_season = (
+            daily_data[time_coord] >= start_edges.rename({time_coord: "group"})
+        ) & (daily_data[time_coord] <= end_edges.rename({time_coord: "group"}))
+    days_in_season = days_in_season.sum(dim="group")
+    daily_data = daily_data.where(days_in_season == 1, drop=True)
```
aaron-kaplan, line 129: You're dropping any days that are in more than one season? How could that happen?

onset_maproom/calc.py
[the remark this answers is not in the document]
aaron-kaplan: OK, I understand this version!

onset_maproom/calc.py
aaron-kaplan: Just readability. My editor window wasn't wide enough to fit your long docstring lines, so they wrapped around, which disturbs the appearance of the indentation, which makes it harder for me to see the shape of the code blocks. (I assume it's obvious what's at stake when you change the URL.)

aaron-kaplan (approved):
Looks good.

At some point I would like to see if we can express the season starts and ends using the [CF conventions for cell boundaries](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.3/build/ch07.html). Xarray doesn't understand cell bounds, so this isn't relevant to our immediate goals, but other existing tools (e.g. ingrid) do understand cell bounds, and xarray might in the future, so I think we should generate CF-compliant metadata if we can. If you want to look into that now, feel free; if you want to put it off for another time I'm ok with that too.

onset_maproom/calc.py
remicousin: it can't happen. the previous sum turns days_in_season from boolean to 1 and 0 and is used to apply OR basically. From there, days_in_season is no more boolean but 1 or 0 so need to tell it to where, needn't I?

remicousin:
See that I made a last commit as I realized my dealing of picking 29 Feb as end of season was not quite right. I explain in the commit comment. Can explain further if need be. I think now it is square.

For the CF conventions business, shall we address the other PR first to wrap up onset_date function? Once this is merged, I'll rebase the other branch on master and I'll try use what we created here to apply the onset_index function with it to make an onset_date function (for multiple years basically). Then if I have time, I am happy to check back out at CF stuff.

onset_maproom/calc.py
aaron-kaplan: I see. You're not dropping days that are in more than one season, you're dropping the days that are in zero seasons.

onset_maproom/calc.py
```diff
@@ -33,6 +35,80 @@ def onset_date(daily_rain, early_start_day, early_start_month, search_days, rain
 #  onset_date = onset_date[time_coord] + onset_date  
   return onset_date
 
+# Time functions
+
+def daily_groupby_season(daily_data, start_day, start_month, end_day, end_month, time_coord="T"):
+  """Groups daily data by yearly seasons from day-month edges of season
+  Can then apply functions on each group to create yearly time series of seasonal quantities
+  If starting day-month is 29-Feb, uses 1-Mar.
```
aaron-kaplan: I think a more natural interpretation would be that the season starts on Feb 29 in leap years and Mar 1 in non-leap years. Not terribly important; I doubt anyone will actually do this.

remicousin: Currently it's not what I am doing. If a user gives a start of 29 Feb, I make it 1 Mar. I am dealing with the end differently because it makes sense to be able to select the entire month of Feb, so if the end input is 28 Feb, they get seasons that end 28 Feb, if they enter 29 Feb, they get the full month of Feb. I feel like that might be useful especially if later this kind of function dives into a bigger environment where "Feb" selects the full month of February, for instance. Whereas I feel like someone who wants a season that starts with 29 Feb when it exists and 1 Mar when it doesn't, would be a jerk :) But obviously I am biased in those choices I make and maybe it would be more natural to deal with start symmetrically to how I deal with end...

aaron-kaplan: I understand, the comment correctly describes the implementation, but I'm suggesting that the implementation decision is a little bit surprising/asymmetrical.

Apparently I have to "resolve" this conversation before the PR can be merged, so I'm going to do that, even though the issue isn't really resolved. It's just not important enough to block us from moving ahead.

onset_maproom/calc.py
aaron-kaplan: This is another issue that's not resolved, but doesn't need to block us right now. Do you have a todo list somewhere to keep track of issues that we want to carry over after we merge the PR?

onset_maproom/calc.py
aaron-kaplan: Put back the original URL.

http://iridl.ldeo.columbia.edu/dochelp/Documentation/details/index.html?func=onsetDate

onset_maproom/calc.py
remicousin: This is obsolete. We don't need to use dropna anymore because: i) we are not using groupby_bins anymore and ii) we are dropping the daily data out of season before dealing with grouping. Thus if a season is fully missing, it will return NA

remicousin: Is there an issue tracker with gitHub? or how do you keep tracks of to-do list on your other coding projects?

aaron-kaplan: Oh right, we don't use dropna anymore. This is resolved.

aaron-kaplan (approved):

PR #4, opened 2021-09-13
Drewr/ethiopia onset

drewmresnick:
Added an extra tab which displays the interactive plot for the onset algorithm.

My way of generating the plot for dates by month/day instead of the plot automatically plotting each point by day/month/year/ is ...perhaps not the best. However I could not find another solution to utilizing datetimes, as the data type seems to require a year. Any ideas as to a smarter way of doing this would be welcome.

### DIFF
--- a/onset_maproom/maproom.py
+++ b/onset_maproom/maproom.py
@@ -106,13 +106,15 @@ def onset_plot(click_lat_lng, earlyStartDay, earlyStartMonth, searchDays, wetThr
     onsetDate = onsetDays["T"] + onsetDays
     year = pd.DatetimeIndex(onsetDate['T']).year
     onsetMD = onsetDate.dt.strftime("2000-%m-%d").to_dataframe(name="Onset Date")
+    onsetMD['Year'] = year
     graph = px.line(
         data_frame=onsetMD,
-        x=onsetMD.index, 
-        y="Onset Date", 
+        x="Year", 
+        y="Onset Date",
     )
     graph.update_traces(
         mode="markers+lines",
+        hovertemplate='%{y} %{x}'
     )
     graph.update_layout(
         yaxis=dict(tickformat="%b %d"), 
--- a/onset_maproom/maproom.py
+++ b/onset_maproom/maproom.py
@@ -99,20 +99,26 @@ def map_click(click_lat_lng):
 )
 def onset_plot(click_lat_lng, earlyStartDay, earlyStartMonth, searchDays, wetThreshold, runningDays, runningTotal, minRainyDays, dryDays,drySpell):
     lat, lng = get_coords(click_lat_lng)
-    ds = rr_mrg.sel(X=lng, Y=lat, method="nearest")
+    ds = rr_mrg.sel(X=lng, Y=lat, method="nearest", drop=True)
     onsetDays = calc.onset_date(ds.precip, int(earlyStartDay), \
         calc.strftimeb2int(earlyStartMonth), searchDays, \
         wetThreshold, runningDays, runningTotal, minRainyDays, dryDays, drySpell)
     onsetDate = onsetDays["T"] + onsetDays
     year = pd.DatetimeIndex(onsetDate['T']).year
-    onsetMD = onsetDate.dt.strftime("1999-%m-%d").to_dataframe(name="Onset Date")
-    graph = px.line(data_frame=onsetMD,x=onsetMD.index, y="Onset Date")
-    graph.update_traces(mode="markers+lines") # , hovertemplate=None)
+    onsetMD = onsetDate.dt.strftime("2000-%m-%d").to_dataframe(name="Onset Date")
+    graph = px.line(
+        data_frame=onsetMD,
+        x=onsetMD.index, 
+        y="Onset Date", 
+    )
+    graph.update_traces(
+        mode="markers+lines",
+    )
     graph.update_layout(
         yaxis=dict(tickformat="%b %d"), 
         xaxis_title="Year", 
         yaxis_title="Onset Date",
-        title= f"Starting dates of {int(earlyStartDay)} {earlyStartMonth} season {year.min()}-{year.max()} ([{lat}E - {lng}N])"
+        title= f"Starting dates of {int(earlyStartDay)} {earlyStartMonth} season {year.min()}-{year.max()} ({lat}E, {lng}N)"
     )
     return graph
 
--- a/onset_maproom/maproom.py
+++ b/onset_maproom/maproom.py
@@ -105,11 +105,11 @@ def onset_plot(click_lat_lng, earlyStartDay, earlyStartMonth, searchDays, wetThr
         wetThreshold, runningDays, runningTotal, minRainyDays, dryDays, drySpell)
     onsetDate = onset["T"] + onset
     year = pd.DatetimeIndex(onsetDate['T']).year
-    onsetMD = onsetDate.dt.strftime("1999-%m-%d")
-    graph = px.line(x=onsetMD['T'], y=onsetMD)
-    graph.update_traces(mode="markers+lines")
+    onsetMD = onsetDate.dt.strftime("1999-%m-%d").to_dataframe(name="Onset Date")
+    graph = px.line(data_frame=onsetMD,x=onsetMD.index, y="Onset Date")
+    graph.update_traces(mode="markers+lines") # , hovertemplate=None)
     graph.update_layout(
-        yaxis=dict(tickformat="%m-%d"), 
+        yaxis=dict(tickformat="%b %d"), 
         xaxis_title="Year", 
         yaxis_title="Onset Date",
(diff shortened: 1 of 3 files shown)


drewmresnick:
@remicousin  @aaron-kaplan  Also the outputs for the new and old graphs do look different.. however I don't want to make any judgments on the code Rémi wrote before mine is checked for errors as well..

remicousin:
The current function you are using is just producing random numbers in the appropriate range, and units. We are done yet on the actual python onset calculations.

remicousin:
Once you have added onset and its T, you should have full dates, day, month and year. Then you can check what was done for bdfloodsmr where the mouse-over prints a full date.

drewmresnick:
> The current function you are using is just producing random numbers in the appropriate range, and units. We are done yet on the actual python onset calculations.

Oh yes right! I forgot about that.

drewmresnick:
> Once you have added onset and its T, you should have full dates, day, month and year. Then you can check what was done for bdfloodsmr where the mouse-over prints a full date.

@remicousin Are you saying that for the mouseover on the graph you would want it to show the full mm/dd/yyyy instead of just the mm/dd?

remicousin:
That would feel more readable to me. I would just put what corresponds to the y-line that is currently there with the full date. That gives all the information needed.

onset_maproom/maproom.py
```diff
@@ -82,6 +84,52 @@ def map_click(click_lat_lng):
...
+        calc.strftimeb2int(earlyStartMonth), params["searchDays"],\
+        params["wetThreshold"], params["runningDays"], params["runningTotal"],\
+        params["minRainyDays"], params["dryDays"], params["drySpell"])
+    #onset = calc.onset_date(ds.precip, int(earlyStartDay,\
+    #    calc.strftimeb2int(earlyStartMonth), \
+    #    searchDays, wetThreshold, runningDays, runningTotal, \
+    #    minRainyDays, dryDays, drySpell)
+    onsetDate = onset["T"] + onset
+    year = pd.DatetimeIndex(onsetDate['T']).year
+    onsetMD = onsetDate.dt.strftime("1999-%m-%d")
```
remicousin, line 123: I understand the question now. It's probably the way to go. It will never be exact because of leap years but since we are having the actual date pop up on mouse over, the approximation of the y-axis is probably good enough.

onset_maproom/maproom.py
```diff
@@ -82,6 +84,52 @@ def map_click(click_lat_lng):
...
+    graph = px.line(x=onsetMD['T'], y=onsetMD)
```
aaron-kaplan, line 124: Did you mean to use `year` for the x axis? I currently see years like `0.6x10^18".

aaron-kaplan (commented):
This looks great.

Can we customize the popup that appears when you hover over the graph? The axis names "x" and "y" aren't meaningful here.

onset_maproom/maproom.py
```diff
@@ -82,6 +84,52 @@ def map_click(click_lat_lng):
...
+    params = {
+        "earlyStart": str(earlyStartDay) + " " + earlyStartMonth,
+        "searchDays": searchDays,
+        "wetThreshold": wetThreshold,
+        "runningDays": runningDays,
+        "runningTotal": runningTotal,
+        "minRainyDays": minRainyDays,
+        "dryDays": dryDays,
+        "drySpell": drySpell
+    }
```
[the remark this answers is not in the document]
drewmresnick: Yes see my latest push I removed this. It was residual from the code Kevin wrote where he needed to set up a dictionary in his callback to speak to Ingrid. I have removed it.

drewmresnick:
> This looks great.
> 
> Can we customize the popup that appears when you hover over the graph? The axis names "x" and "y" aren't meaningful here.

@aaron-kaplan I am in the process of customizing the popup now. As for the axes.. I am not sure why yours displays them as "x" and "y" - mine shows the axes as "Year" and "Onset Date". Refer to the update_layout function starting at 111 where I update the axes.

onset_maproom/maproom.py
drewmresnick: When I run the app this does not happen.. so I am wondering why there are discrepancies in the way the app is running for the two of us? The graph shows the x axis as integers years~ 1981-2020

aaron-kaplan:
I wrote my comments prior to seeing the two commits you just pushed. The years are displayed correctly now.

onset_maproom/maproom.py
```diff
@@ -82,6 +84,37 @@ def map_click(click_lat_lng):
...
+    Input("wetThreshold", "value"),
+    Input("runningDays", "value"),
+    Input("runningTotal", "value"),
+    Input("minRainyDays", "value"),
+    Input("dryDays", "value"),
+    Input("drySpell", "value"),
+)
+def onset_plot(click_lat_lng, earlyStartDay, earlyStartMonth, searchDays, wetThreshold, runningDays, runningTotal, minRainyDays, dryDays,drySpell):
+    lat, lng = get_coords(click_lat_lng)
+    ds = rr_mrg.sel(X=lng, Y=lat, method="nearest")
```
aaron-kaplan, line 102: Minor suggestion: add `drop=True` to get rid of the X and Y dimensions here. While stepping through the code, I noticed that the `onsetMD` dataframe currently looks like this:
```
                  X       Y  Onset Date
T                                      
1981-06-01  38.7375  9.0375  1999-08-07
1982-06-01  38.7375  9.0375  1999-06-18
1983-06-01  38.7375  9.0375  1999-08-05
```

onset_maproom/maproom.py
```diff
@@ -82,6 +84,37 @@ def map_click(click_lat_lng):
...
+    onsetDays = calc.onset_date(ds.precip, int(earlyStartDay), \
+        calc.strftimeb2int(earlyStartMonth), searchDays, \
+        wetThreshold, runningDays, runningTotal, minRainyDays, dryDays, drySpell)
+    onsetDate = onsetDays["T"] + onsetDays
+    year = pd.DatetimeIndex(onsetDate['T']).year
+    onsetMD = onsetDate.dt.strftime("1999-%m-%d").to_dataframe(name="Onset Date")
```
aaron-kaplan, line 108: Use 2000 instead of 1999 in case there's an onset on Feb 29?

onset_maproom/maproom.py
drewmresnick: Okay. Is there a benefit for doing this other than it making the data more simple?

aaron-kaplan:
We start with an integer day-of-the-year (onsetDays) for each year. These can be compared with each other (day 3 is earlier than day 4), but aren't very human-readable. So we calculate T + onsetDays to get the full datetime (day, month, and year) of each onset. Now they're human-readable, but comparing them with each other is no longer meaningful (every date in 2000 is later than every date in 1999; that's not what we want to know when we compare one year's onset date with another's). So then we replace the year parts with 1999, resulting in dates that are both human-readable and comparable with each other.

Here's an alternative approach that seems simpler: plot the original integer onsetDays, but label the axis with the month and day of each day in 1999.

Both approaches will be slighly wrong in some cases because of leap years, but your approach hides the problem well whereas it will be more visible in mine. That could be an argument for sticking with the way you've done it.

Not necessarily asking for a change here, just offering a possibility in case it triggers more thoughts.

onset_maproom/maproom.py
aaron-kaplan: Probaby no advantage to doing it in general. Ordinarily it takes very little extra space to keep the extra dimensions, and the extra metadata might come in handy sometimes. The problem I noticed only arises when you use `to_dataframe`, which copies the metadata into each row of the table. For a table with many rows, this could waste a significant amount of space. (For this application, where the number of rows is pretty small, it really makes no difference. I probably shouldn't have mentioned it.)

drewmresnick:
> That would feel more readable to me. I would just put what corresponds to the y-line that is currently there with the full date. That gives all the information needed.

@remicousin My latest commit makes it so that the tooltip expresses the date as "%b %d %Y" upon hover

aaron-kaplan (approved):

onset_maproom/maproom.py
```diff
@@ -82,6 +84,45 @@ def map_click(click_lat_lng):
...
+    graph.update_traces(
+        mode="markers+lines",
+        hovertemplate='%{y} %{x}'
+    )
+    graph.update_layout(
+        yaxis=dict(tickformat="%b %d"), 
+        xaxis_title="Year", 
+        yaxis_title="Onset Date",
+        title= f"Starting dates of {int(earlyStartDay)} {earlyStartMonth} season {year.min()}-{year.max()} ({lat}E, {lng}N)"
+    )
```
remicousin, line 124: Which lat/lon is being printed? I would print the result of nearest point, as it is what the data is. Thus, given the data spatial resolution is 0.0375, I would make sure that it prints at 4 digits after decimal point so that there is not ambiguity which data spatial point is being looked at (currently only 2 digits).

remicousin: this final 1 is just a machine error reading/writing. Here 4 as we know the resolution is 0.0375

remicousin (approved):

PR #5, opened 2021-09-15
Use dlcharts charts that pull data from IRI instead of NMA

aaron-kaplan:
I used the dlcharts UI to create copies of the snippets that use our local copy of ENACTS_v7 instead of NMA's copy of some older version. The final version won't use dlcharts at all; this is just to compare against the output of the python reimplementations.

I was hoping this would be faster than the one that pulled from NMA, but it's not. It's still an improvement, because the NMA server is frequently down.

I only modified the graphs that use dlcharts, namely onset date, onset probability of exceedance, and cessation date. The [cessation probability of exceedance chart](https://github.com/iridl/x24-vector/blob/master/onset_maproom/charts.py#L33) uses the NMA server directly, without dlcharts. I started editing that to make it use the IRI server, but I'm not sure it's correct. @remicousin is the rainfall cessation date calculation really supposed to use a temperature climatology, or is was that a copy-paste error?


aaron-kaplan:
There's also the PDF link, which uses dlsnippets. Looking into that now.

aaron-kaplan:
And there's the table, which also pulls from NMA, but I don't think we need to bother making a local version of that because we're planning to get rid of it.

aaron-kaplan:
I made local versions of all the snippets involved in generating the PDF, but it doesn't work because of the authentication issue (this is an LDAP-protected dataset, and dlsnippets isn't compatible with LDAP). I'm not very motivated to try to fix this. We don't need to do a side-by-side comparison of the python-generated PDF with the dlsnippets-generated one, because it's just an aggregation of charts that we can verify individually.

aaron-kaplan:
I converted the cessation probability graph. It works, but takes for...e....ver. Still want @remicousin to tell me what the tmin climatology has to do with rainfall cessation date.

remicousin:
It's not using tmin really as it converts it to a constant of 5 and then pretends that is EvapoTranspiration (ET). I am not sure why it's done this way. This cessation calculation was done in a rush back then, so I can only assume that it is a leftover from a previous calculation attempt or something...
Cessation doesn't rely on rainfall, but on soil moisture, so there is an ultra simplistic estimation of soil moisture that is done for this. And then a cessation algorithm kicks in.
Bottom line is that in the Ingrid code, all
ET
can be replaced by
5
and
SOURCES .Ethiopia .NMA .monthly .climatologies .tmin
   /region get_parameter geoobject
    [X Y]weighted-average
   0.0 mul
   5.0 add
   [T]regridAverage
   2 RECHUNK
   {rainfall ET}ds
by
 {rainfall}ds

drewmresnick (dismissed):
Not sure if you are wanting to wait until the discussion regarding cessation dates is resolved to merge, but I will approve for now regardless.

aaron-kaplan:
> It's not using tmin really as it converts it to a constant of 5 and then pretends that is EvapoTranspiration (ET). I am not sure why it's done this way. This cessation calculation was done in a rush back then, so I can only assume that it is a leftover from a previous calculation attempt or something...
> Cessation doesn't rely on rainfall, but on soil moisture, so there is an ultra simplistic estimation of soil moisture that is done for this. And then a cessation algorithm kicks in.
> Bottom line is that in the Ingrid code, all
> ET
> can be replaced by
> 5
> and
> SOURCES .Ethiopia .NMA .monthly .climatologies .tmin
> /region get_parameter geoobject
> [X Y]weighted-average
> 0.0 mul
> 5.0 add
> [T]regridAverage
> 2 RECHUNK
> {rainfall ET}ds
> by
> {rainfall}ds

Thanks. @drewmresnick that's mostly for your benefit, when you get to rewriting this code in python.

aaron-kaplan:
@drewmresnick could you approve again? I rebased locally and pushed, which made github consider your approval "stale."

remicousin (approved):

drewmresnick (approved):

PR #6, opened 2021-09-16
Display probability threshold

aaron-kaplan:
I accidentally sent this PR [on bitbucket](https://bitbucket.org/iridl/ingrid/pull-requests/41) first. Already merged there, now bringing github back up to date.


aaron-kaplan:
Never mind, doing it this way would result in different merge commits between bitbucket and github. I'll do it a different way.

PR #7, opened 2021-09-17
Add forecasts for DJF 2022 to Madagascar

aaron-kaplan:
Note that the year range was set incorrectly before. We have probably been misinterpreting the results until now. I will send an email to everyone about this with more details.


nitinmagima (approved):
Thanks @aaron-kaplan!

I've updated it on my side.

PR #8, opened 2021-09-20
adding onset_date in xarray

remicousin:
renaming (random) onset_date that will become seasonal_onset_date
I guess for consistency...

### DIFF
--- a/onset_maproom/calc.py
+++ b/onset_maproom/calc.py
@@ -13,7 +13,7 @@ def read_zarr_data(zarr_path):
 # Growing season functions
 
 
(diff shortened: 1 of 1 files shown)


aaron-kaplan (approved):
Not crazy about the name of the function, but it's time to move on!

remicousin:
yeah.... but we have to distinguish the one that just looks at one time stream and the one that will do so for multiple years at once... I don't like it very much but on the other hand the equivalent seasonal_sum makes a lot of sense to me... Not sure what this cognitive glitch is about...

PR #9, opened 2021-09-22
Drewr/probExceedGraph

drewmresnick:
Added a function in calc.py to calculate probabilities.

Added graph and formatting in layout.py and maproom.py


onset_maproom/calc.py
```diff
@@ -268,6 +268,23 @@ def seasonal_sum(
     summed_seasons = xr.merge([seasonal_data, seasons_ends])
     return summed_seasons
 
+def probExceed(onsetMD, earlyStart):
+    onsetMD_df = onsetMD
```
remicousin, line 272: Should we try to make a generic probExceed function that takes an xarray and the coord against which the probExceed is to be computed against, and probably like in Ingrid, the range and step to apply to the binning?

In that case the specifics of the onset date case should be taken away (e.g. earlyStart) and dealt with outside of the function; and should try to use xarray objects.

For instance, I wonder if we could use [quantile](https://xarray.pydata.org/en/stable/generated/xarray.DataArray.quantile.html) (unless there is something more directly appropriate -- I still am not too good at searching existing functions). Then the exercise is mostly to construct q with the right range and step and then do some manips to get percent rather than quantiles, and maybe calculate the complementary (1-x).

remicousin:
I made some more overarching comments for now. I would let Aaron chiming in with regards to generalizing vs speeding up. As for the position of the function, I tried to organize things a bit but it's still messy to I don't think it matters for now...

remicousin:
or maybe it's not worth to create a generic probability of exceeding function and you can just use quantile for this specific case...

onset_maproom/calc.py
aaron-kaplan: I think it's ok to focus on the specific case at hand at first. We can generalize later.

onset_maproom/calc.py
```diff
@@ -268,6 +268,23 @@ def seasonal_sum(
...
+    onsetDatetime = pd.to_datetime(onsetMD_df.onset)
+    onsetDiff = onsetDatetime - earlyStart
+    onsetDiff_df = onsetDiff.to_frame()
+    counts = onsetDiff_df['onset'].value_counts()
+    countsDF = counts.to_frame().sort_index()
+    cumsum = countsDF.cumsum()
+    onset = []
+    for i in onsetDiff_df.T:
+        float = onsetDiff_df.onset[i].total_seconds() / 86400
+        onset.append(float)
```
aaron-kaplan, line 282: Can lines 279-282 be rewritten in one line?
```
onset = onsetDiff_df.onset.dt.total_seconds() / (24 * 60 * 60)
```
When there's a choice between using a python loop and using matrix algebra, the matrix algebra approach is generally preferable . Besides being more succinct, it's much faster. (That said, if you find it easier to think in terms of loops, go ahead and write it that way first. Once you get it right, you can rewrite it in a more efficient way and then verify that the result is the same.)

onset_maproom/maproom.py
```diff
@@ -102,32 +103,48 @@ def map_click(click_lat_lng):
     Input("dryDays", "value"),
     Input("drySpell", "value"),
 )
-def onset_plot(click_lat_lng, earlyStartDay, earlyStartMonth, searchDays, wetThreshold, runningDays, runningTotal, minRainyDays, dryDays,drySpell):
+def onset_plots(click_lat_lng, earlyStartDay, earlyStartMonth, searchDays, wetThreshold, runningDays, runningTotal, minRainyDays, dryDays,drySpell):
```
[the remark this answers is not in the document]
drewmresnick: I think this could be a good idea, I struggled with that variable name personally so something more clear may be more user friendly?

onset_maproom/calc.py
drewmresnick: That is definitely something I can work on at some point once I have this specific case finished.

drewmresnick:
@remicousin @aaron-kaplan The function is quite slow at the moment. 

I tried selecting from the data using lat/lng before calling the function, but that broke the code. Did not spend time investigating why yet. However if it will make the function faster, I can try to debug. 

Although I wonder if the fact that it is using dask arrays means that it is not actually calculating the values until after .sel is run, since we call .values at the end of the line of code? (line 109 in maproom.py)

`    onsetDate = (calc.seasonal_onset_date(rr_mrg.precip, int(earlyStartDay), \
        calc.strftimeb2int(earlyStartMonth), int(searchDays), \
        int(wetThreshold), int(runningDays), int(runningTotal), \
        int(minRainyDays), int(dryDays), int(drySpell), time_coord="T").onset_delta \
        + calc.seasonal_onset_date(rr_mrg.precip, int(earlyStartDay), \
        calc.strftimeb2int(earlyStartMonth), int(searchDays), \
        int(wetThreshold), int(runningDays), int(runningTotal),\
        int(minRainyDays), int(dryDays),\
        int(drySpell), time_coord="T")["T"]).sel(X=lng, Y=lat, method="nearest").values`

I have a very basic understanding of dask so I am not sure how sound my reasoning is on this.

drewmresnick:
I also get this warning:

`eWarning: All-NaN slice encountered
  return func(*(_execute_task(a, cache) for a in args))
/home/drewr/anaconda3/envs/enactsmaproom/lib/python3.9/site-packages/dask/array/reductions.py:960: RuntimeWarning: All-NaN slice encountered
  vals = func(x, axis=arg_axis, keepdims=True)`

remicousin:
> I also get this warning:
> 
> `eWarning: All-NaN slice encountered return func(*(_execute_task(a, cache) for a in args)) /home/drewr/anaconda3/envs/enactsmaproom/lib/python3.9/site-packages/dask/array/reductions.py:960: RuntimeWarning: All-NaN slice encountered vals = func(x, axis=arg_axis, keepdims=True)`

I also get that during my tests. I believe it is from the grid cells out of Ethiopia that are all NaN. I suspected that was ok but @aaron-kaplan may have another opinion on that.

aaron-kaplan:
> The function is quite slow at the moment.

The Ingrid version is also quite slow. Is this worse? In any case, let's first get the whole thing working end-to-end, and worry about optimization later.

I looked into the warning while reviewing Rémi's code. What it's noticing is that there are points on the grid where it can't calculate an onset date because there are no rainfall data. IIRC `.isel(X=0, Y=0)` was one such point. I believe this is normal and expected, so now we need to look into

- What effect does this have on the output? I'm guessing it results in an "onset offset" of NaT for the affected grid points?
- Does our code behave appropriately when this happens? E.g. if someone selects a grid point for which we have no data, do we somehow make it clear what happened, or is there a cryptic error message like "500 server error?"

Once we've assured ourselves that we're handling the condition appropriately, we should [suppress the warning](https://docs.python.org/3/library/warnings.html#warnings.filterwarnings). Otherwise we'll eventually get used to the log being full of warnings, which will lead us to miss other warnings that are actually important.

aaron-kaplan:
Have you compared your graphs with the originals? They look completely different.

aaron-kaplan:
> The Ingrid version is also quite slow. Is this worse? In any case, let's first get the whole thing working end-to-end, and worry about optimization later.

After trying it, I take this back. It's so slow that it's going to be difficult to work on. I already made one suggestion that I think will speed it up by a factor of two. Another thing to try (which is only a short-term workaround, not a real fix) is to add `@lru_cache` to the seasonal_onset_date function declaration. See fbfmaproom/fbfmaproom.py in this repo for examples.

onset_maproom/maproom.py
```diff
@@ -102,32 +104,53 @@ def map_click(click_lat_lng):
...
+        y="probExceed",
+    )
+    probExceed_graph.update_traces(
+        mode="markers+lines",
+        hovertemplate= 'Days since Early Start Date: %{x}'+'<br>Probability: %{y:.0%}'
+    )
+    probExceed_graph.update_layout(
+        yaxis=dict(tickformat=".0%"),
+        yaxis_title="Probability of Exceeding",
+        xaxis_title=f"Onset Date since {earlyStartDay} {earlyStartMonth} [days]"
```
aaron-kaplan, line 151: Suggest "Onset date [days since {earlyStartDay} {earlyStartMonth}]"

aaron-kaplan:
I just noticed that dask is listed in environment.yml but not in environment_linux.yml.

@remicousin when you [added](https://bitbucket.org/iridl/x24-vector/pull-requests/52) dask to environment.yml, you should have regenerated environment_linux.yml.

@drewmresnick you mentioned seeing evidence of dask while debugging, which means you didn't build your conda environment from environment_linux.yml. You should (assuming you're developing on a linux server, not on your laptop).

To review: environment.yml is a list of the packages we use directly, listed without version numbers except when we specifically need a particular version. environment_linux.yml is a snapshot of what conda actually installed, including version numbers, on one particular day. When we create an environment for development or to run on a server, we should use environment_linux.yml. This ensures that we get exactly the same packages every time, except when explicitly decide to update them. This protects us from the situation where something works for one developer but not for another, or works on one server but not on another, because their conda environments were built on different days, resulting in different versions of various packages.

drewmresnick:
> > The Ingrid version is also quite slow. Is this worse? In any case, let's first get the whole thing working end-to-end, and worry about optimization later.
> 
> After trying it, I take this back. It's so slow that it's going to be difficult to work on. I already made one suggestion that I think will speed it up by a factor of two. Another thing to try (which is only a short-term workaround, not a real fix) is to add `@lru_cache` to the seasonal_onset_date function declaration. See fbfmaproom/fbfmaproom.py in this repo for examples.

I will look more into this tomorrow; when I add @lru_cache in the calc.py file above the function declaration it breaks the code with error:

`TypeError: unhashable type: 'DataArray'`

I am not sure why yet, if you have an idea @aaron-kaplan let me know otherwise I will try to debug tomorrow. It is a bit painful to try working on the maproom with it loading so slowly..!

aaron-kaplan:
OK, then forget about lru_cache. I'm working on other ways of speeding it up, will report later.

aaron-kaplan:
I took the liberty of pushing some commits. One brings environment_linux.yml up to date as discussed earlier. The other two make the chart generation much faster, by

- Calling seasonal_onset_date once instead of twice,
- Calling seasonal_onset_date on just the one lat/lon point instead of the entire country,
- Changed the config file to use a copy of the dataset that I have rechunked. The original chunking was one file per day; if you wanted the entire time series for a single grid point, you had to open some 13,000 files and read one data point from each of them. Now it's chunked primarily by geographic area instead of by time.
- Added a call to `.load` which seems to cause it to read the data in a more efficient way.

I think you should now concentrate on validating the calculation, i.e. figuring out why the new charts look completely different from the old ones.

remicousin:
I am playing a little bit with it and it is a bit difficult indeed to get results for both sets of graphs together. One thing that will likely help is that I noticed that there are often cases where the python version doesn't update with new choices (playing with start date) and returns an error message and the last successful graph ( a behavior likely to change at some point). But I also noticed that there is never a time series with missing values, for years the algo doesn't find an onset date. I suspect those 2 are related. That is to say that the graph is not plotted when one of the years (or more) have missing. Fixing this would facilitate comparison as it is hard to get a time series plotted right now. So that give us a bigger population to make comparisons. Maybe it has to do with the fact that missing are going to be NaT and that might reauire special handling for the figure?

remicousin:
Also I am using this to pick meaningful pairs of location and search start. Ethiopia has a very diverse climate and onset dates can be at very different times of the year in different parts of the country

http://iridl.ldeo.columbia.edu/maproom/Global/Climatologies/Select_a_Point.html?bbox=bb%3A31.898%3A2.829%3A48.982%3A17.601%3Abb&region=bb%3A33%3A7.75%3A33.5%3A8.25%3Abb

remicousin:
I am also seeing that the location label for the Ingrid plots are wrong. for instance the default one says something like 38E-8E; 39N-9N. When it should clearly be 38E-39E; 8N-9N. if it is only a labeling issue, that won't explain any graphs difference, but if Ingrid indeed computes onset dates for such big areas average, then it is likely explaining A LOT of the differences.

So I think those 2 issues: have the python produce a time series even where there is missing; and make sure the bounding box entries are not partially interverted for Ingrid, will take us a good deal forward to allow comparisons.

onset_maproom/maproom.py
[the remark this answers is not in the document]
aaron-kaplan: Rémi made this change in his code before merging his last PR, so now we have a mix of earlyStartDay and search_start. @drewmresnick please update your code to match.

remicousin:
I likely won't have a good look at this before 10 days from now. My next 2 half-days of work are going to be in trains and was thinking of dedicating that to email, admin, and rather no-brainer stuff

drewmresnick:
I will spend today and tomorrow looking at the code I have written, based on the comments you both have left above. @aaron-kaplan I will let you know if I get to a point where I am stuck.

onset_maproom/maproom.py
```diff
@@ -102,32 +104,53 @@ def map_click(click_lat_lng):
...
-    onsetMD = onsetDate.dt.strftime("2000-%m-%d").to_dataframe(name="Onset Date")
+    onsetDate = (calc.seasonal_onset_date(rr_mrg.precip, int(earlyStartDay), \
+        calc.strftimeb2int(earlyStartMonth), int(searchDays), \
+        int(wetThreshold), int(runningDays), int(runningTotal), \
+        int(minRainyDays), int(dryDays), int(drySpell), time_coord="T").onset_delta \
+        + calc.seasonal_onset_date(rr_mrg.precip, int(earlyStartDay), \
+        calc.strftimeb2int(earlyStartMonth), int(searchDays), \
+        int(wetThreshold), int(runningDays), int(runningTotal),\
+        int(minRainyDays), int(dryDays),\
+        int(drySpell), time_coord="T")["T"]).sel(X=lng, Y=lat, method="nearest").values
```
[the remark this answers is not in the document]
drewmresnick: Okay. I would like to discuss further with you as well how to handle errors in this case. Specifically, how to alert the user that the point they clicked does not have data. Currently the graphs do not update and you are only returned with a KeyError.

drewmresnick:
> I just noticed that dask is listed in environment.yml but not in environment_linux.yml.
> 
> @remicousin when you [added](https://bitbucket.org/iridl/x24-vector/pull-requests/52) dask to environment.yml, you should have regenerated environment_linux.yml.
> 
> @drewmresnick you mentioned seeing evidence of dask while debugging, which means you didn't build your conda environment from environment_linux.yml. You should (assuming you're developing on a linux server, not on your laptop).
> 
> To review: environment.yml is a list of the packages we use directly, listed without version numbers except when we specifically need a particular version. environment_linux.yml is a snapshot of what conda actually installed, including version numbers, on one particular day. When we create an environment for development or to run on a server, we should use environment_linux.yml. This ensures that we get exactly the same packages every time, except when explicitly decide to update them. This protects us from the situation where something works for one developer but not for another, or works on one server but not on another, because their conda environments were built on different days, resulting in different versions of various packages.

Do you mean by this that you think I installed the env using environment.yml insteal of the linux version?

aaron-kaplan:
> > I just noticed that dask is listed in environment.yml but not in environment_linux.yml.
> > @remicousin when you [added](https://bitbucket.org/iridl/x24-vector/pull-requests/52) dask to environment.yml, you should have regenerated environment_linux.yml.
> > @drewmresnick you mentioned seeing evidence of dask while debugging, which means you didn't build your conda environment from environment_linux.yml. You should (assuming you're developing on a linux server, not on your laptop).
> > To review: environment.yml is a list of the packages we use directly, listed without version numbers except when we specifically need a particular version. environment_linux.yml is a snapshot of what conda actually installed, including version numbers, on one particular day. When we create an environment for development or to run on a server, we should use environment_linux.yml. This ensures that we get exactly the same packages every time, except when explicitly decide to update them. This protects us from the situation where something works for one developer but not for another, or works on one server but not on another, because their conda environments were built on different days, resulting in different versions of various packages.
> 
> @aaron-kaplan Do you mean by this that you think I installed the env using environment.yml insteal of the linux version?

That's my guess, since dask seems to be installed in your environment but dask wasn't included in environment_linux.yml. Does that sound plausible?

drewmresnick:
The commits I just pushed seem to fix the issue of the graphs not updating properly re missing onset dates. However there is still an issue updating graphs when you try to type in a number versus using the up/down arrows for the inputs, which I need to look more into.

drewmresnick:
> The commits I just pushed seem to fix the issue of the graphs not updating properly re missing onset dates. However there is still an issue updating graphs when you try to type in a number versus using the up/down arrows for the inputs, which I need to look more into.

Upon further investigation, it is not the actual input of a number into the text box that throws an error, it is the blank space you create when you delete the previous number. So this is actually not an issue.

As far as I can tell the graphs are updating properly now.

It seems that overall the new alrgorithm is doing a good job at mimicking overall the previous one (although I am not sure how close we expect them to be, as some yearly are off by several days.. but mostly at least within the same month. I will leave this to @remicousin to make a comment on). 

I did notice that the plot for the new onset alg plots years 1981- 2020, whereas the table for the old alg shows dates 1983-2014.

aaron-kaplan:
>
> Upon further investigation, it is not the actual input of a number into
> the text box that throws an error, it is the blank space you create when
> you delete the previous number. So this is actually not an issue.


It's still an issue, just one that has a workaround. If you ran into this,
the users probably will too, and they're going to waste their own time and
ours finding out how to work around it. We should fix it. Not necessarily
in this PR, but keep track of it as something to be addressed. (We need a
place to keep track of these things. How about we try out the github issue
tracker for this project?)

drewmresnick:
@aaron-kaplan I did not reply to your suggestion about using issue tracking for the project, but am open to trying it if you think it will be useful.

onset_maproom/maproom.py
```diff
@@ -102,33 +106,72 @@ def map_click(click_lat_lng):
...
+    else:
+        precip.load()
+    try:
+        onset_delta = calc.seasonal_onset_date(precip, int(search_start_day),
+            calc.strftimeb2int(search_start_month), int(searchDays),
+            int(wetThreshold), int(runningDays), int(runningTotal),
+            int(minRainyDays), int(dryDays), int(drySpell), time_coord="T")
+    except TypeError:
+        errorFig = pgo.Figure().add_annotation(x=2, y=2,text="No Data to Display",font=dict(family="sans serif",size=30,color="crimson"),showarrow=False, yshift=10, xshift=60)
+        alert1 = dbc.Alert("Please ensure all input boxes are filled for the calculation to run.", color="danger", dismissable=True)
```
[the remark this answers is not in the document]
aaron-kaplan: Check out the `debounce` option. That might help prevent unnecessary form submissions.
https://dash.plotly.com/dash-core-components/input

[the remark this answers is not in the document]
aaron-kaplan: Setting `max` might also help prevent some invalid values, like March 392 as a date.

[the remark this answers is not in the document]
aaron-kaplan: Feel free to propose changes to Kevin's code where appropriate.

In the long run, it would be nice to have a Date input type, with date-specific validation and maybe a convenient picker interface. Dash has a date picker class, but it requires you to pick a year as well as month and day. At some point we should probably extend that to support leaving the year out. Let's not focus on that for now, though. Let's pick the low-hanging fruit (if any) on the front end, and make sure that when back-end validation fails, the user gets good feedback and it's not hugely annoying.

aaron-kaplan (commented):
IIUC the only thing left to do here is compare with the original graphs from Ingrid and/or dlcharts. Right? Are you working with @kgraaf to figure that out?

onset_maproom/maproom.py
```diff
@@ -111,11 +112,11 @@ def onset_plots(click_lat_lng, search_start_day, search_start_month, searchDays,
     lat, lng = get_coords(click_lat_lng)
     try:
         precip = rr_mrg.precip.sel(X=lng, Y=lat, method="nearest", tolerance=0.04)
-        isnan = np.isnan(precip)
-        if any(isnan) == True:
+        isnan = np.isnan(precip).sum().sum()
+        if isnan > 0:
```
drewmresnick, line 116: @aaron-kaplan This speeds up the maproom greatly. I also think this could be useful if we decide on some sort of threshold of missing values?

aaron-kaplan: whatever!

aaron-kaplan:
@drewmresnick Kevin is working on the onset maproom now too, but what's currently on master doesn't work, so he can't start his branch from master. We need either to merge this branch, or pick through the changes and figure out which one fixed the problem that's present on master so we can merge that change separately. Given that nobody is using the application yet, I see no reason not to merge this, even if the graphs aren't correct. Unless you have specific reasons not to, please rebase, resolve the conflicts, and then merge.

aaron-kaplan (approved):

remicousin:
I approve so as not to be a blocker as it seems this needs to move forward and remaining problems can be reconsidered in new PRs. (I don't see a button for me to approve though but please go ahead and consider I approved).

aaron-kaplan:
To approve, you go to the "Files changed" tab, then the green "review changes" button on the right.

remicousin (approved):

drewmresnick:
Okay I will do this. I think we were just waiting on comparing the algorithms, which I just finished doing and was planning on talking about during the team meeting. But the maproom itself seems to be working fine, if we are not focused on the performance of the algorithm itself so I think it is worth merging.

PR #10, opened 2021-09-23
Remic/seasonset

remicousin:


aaron-kaplan (commented):
Looks good, only minor stylistic suggestions.

(I would approve now, but under the current settings my approval would be invalidated when you push new commits. Should we relax the setting and trust people to ask for another review if they make non-trival changes after approval?)

remicousin:
I think we can relax settings. All the cosmetics changes are pushed. One thing maybe: the output is now a dataset whereas the dummy case was just an array. This is because I saved seasons_ends to be consistent with seasonal_sum.... but maybe that's not that useful for that case?
either I output an array, or the rest using seasonal_onset_date will need slight editing.

aaron-kaplan:
> the output is now a dataset whereas the dummy case was just an array. 
> either I output an array, or the rest using seasonal_onset_date will need slight editing.

I'll let you and Drew sort out the best sequence for merging your respective changes.

aaron-kaplan (approved):

remicousin:
@drewmresnick , it is ok if I merge now? that is likely going to conflict a bit with what you are doing now. And I am going to be offline at least until Tuesday soon

drewmresnick:
@remicousin I suppose I am a bit unclear as to what the potential conflicts would be. Would it just have to do with the format that the results are in (array versus dataset?) If there is more to it than this, would you have time today to chat over zoom?

remicousin:
no. Nothing else I think. Or maybe just basic "conflict" because we are editing the same file, even though we aren't editing the same locations in the file...

drewmresnick:
Okay this seems manageable to me to work with, and seems like something Aaron would be able to help me on if I got really stuck. I am approving for you to merge.

drewmresnick (approved):

PR #13, opened 2021-09-28
Move logos out of this codebase

aaron-kaplan:
This repo is supposed to provide generic code that can be used to set up maprooms for different countries on different servers. Country-specific details like datasets and icons should be kept somewhere else. In this PR I'm removing just the icons. The immediate impetus for this change is that I don't want to have to rebuild the docker image every time we add a new country.

I will be making a compation pull request to add these icons to the dlconfig repository, which is where we maintain various files that are specific to services that run at the IRI.


aaron-kaplan:
dlconfig PR is here: https://bitbucket.org/iridl/dlconfig/pull-requests/143/deploy-icons-for-fbfmaproom2

PR #14, opened 2021-09-28
Put path to data dir in one place instead of many

aaron-kaplan:
This should make it less of a pain to keep our config files in sync with the sample.


fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -43,6 +43,8 @@ dataframes:
         schema: public
         table: fbf_maproom
 
+data_root: /home/aaron/scratch/fbf-candidate
```
aaron-kaplan, line 46: change to /data/aaron/fbf

PR #15, opened 2021-10-01
Restore outline of selected shape

aaron-kaplan:
The map is supposed to display the shape of the selected region, but we broke that a while ago and didn't notice. This fixes it.


aaron-kaplan:
I figured out why we hadn't noticed this problem: PropTypes are only enforced in debug mode. The boundary is still being drawn correctly in production mode.

PR #20, opened 2021-10-14
tests for daily grouping by season function

remicousin:
I blacked it so the array of daily data (a few days or several years) is not arranged in a very convenient manner...


onset_maproom/tests/test_calc.py
```diff
@@ -4,24 +4,2025 @@
...
+        == xr.DataArray(
+            [
+                "NaT",
+                "2001-03-08T00:00:00.000000000",
+                "NaT",
+                "2003-04-12T00:00:00.000000000",
+                "2004-04-04T00:00:00.000000000",
+            ]
+        )
+    ).all
```
aaron-kaplan, line 1933: Did you actually verify that these results are correct? Even if you did, it's hard for the reader (me) to verify it for himself. I think it would be easier if you used synthetic data.

I'm not saying this test is worthless. It works as a regression test: if one day we change the function in a way that changes its outputs when we didn't intend to, this test can alert us to it. But it doesn't help (as much as it could) to convince us that the function is correct in the first place.

BTW it might be good to have one test that uses (a small amount of) actual data, in addition to ones that use synthetic data, to guard against writing code that is wrong but passes all the tests, because the code and the tests were both based on the same misunderstanding of what the data are supposed to look like.

onset_maproom/tests/test_calc.py
```diff
@@ -4,24 +4,2025 @@
...
+
+    dts = calc.daily_tobegroupedby_season(precip, 29, 11, 29, 2)
+
+    assert dts["T"].size == 461
+    assert dts["group"].size == 5
+    assert (
+        dts.seasons_ends
+        == xr.DataArray(
+            [
+                "2001-02-28T00:00:00.000000000",
```
aaron-kaplan, line 1909: One thing that's very important when writing tests: always perturb the test and see it fail, so you know that you're actually testing what you think you're testing. I tried changing `2001-02-28` in this line to `2001-02-27`, and the test still passes.

In this case, the test passes because you're asserting `ds.all` instead of `ds.all()`. `ds.all` is a function, which is a "truthy" value, so the assertion passes no matter what is in `ds`. You did the same further down. I didn't notice this when reading the code, I only found it after perturbing the setup and seeing that the test still passed.

aaron-kaplan (changes requested):
> I probably want to describe the different scenarios, no?

Yes, comments would be very helpful. Also using succinct but descriptive function names will help when reading the test report, because it lists the names of test functions that failed.

> In the future, will be easier to set up while writing the functions.

Highly encouraged. Some people even recommend writing the test first (described as "test-driven development" or TDD), because it induces to you write more testable code.

remicousin:
Be prepared when searching fmt on the internet...!

onset_maproom/tests/test_calc.py
```diff
@@ -4,24 +4,2025 @@
...
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.451228,
+        0.054314,
+        0.333831,
+        0.637705,
+        6.406705,
+    ]
```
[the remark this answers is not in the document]
remicousin: for now it's using more than one column and won't be balcked anymore.

onset_maproom/tests/test_calc.py
```diff
@@ -4,10 +4,463 @@
 
 import calc
 
-def test_onset():
+
+def set_multi_year_data_sample():
```
aaron-kaplan, line 8: This doesn't set anything. Call it `get_multi_year_data_sample` or just `multi_year_data_sample`.

onset_maproom/tests/test_calc.py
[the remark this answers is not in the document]
aaron-kaplan: This is still big enough that I'd rather it be in a separate file (assuming it doesn't get replaced with synthetic data).

onset_maproom/tests/test_calc.py
```diff
@@ -4,10 +4,463 @@
...
+        dry_spell_length=7,
+        dry_spell_search=21,
+        time_coord="T",
+    )
+    onsets = onsetsds.onset_delta + onsetsds["T"]
+    assert np.isnat((onsets[0]))
+    assert onsets[1] == pd.to_datetime("2001-03-08T00:00:00.000000000")
+    assert np.isnat((onsets[2]))
+    assert onsets[3] == pd.to_datetime("2003-04-12T00:00:00.000000000")
+    assert onsets[4] == pd.to_datetime("2004-04-04T00:00:00.000000000")
```
aaron-kaplan, line 456: Why are you making assertions about individual array elements here, rather than comparing the whole array as you did elsewhere? The disadvantage of doing it this way is that the test stops after the first failed assertion, so we don't get the full picture of what works and what doesn't.

onset_maproom/tests/test_calc.py
```diff
@@ -4,10 +4,463 @@
...
+
+
+def set_precip_sample():
```
aaron-kaplan, line 459: As above regarding the name "set"

onset_maproom/tests/test_calc.py
```diff
@@ -37,3 +496,90 @@ def test_onset():
...
+    precipnoWD = xr.where(
+        (precip["T"] == pd.to_datetime("2000-05-05")),
+        1.1,
+        precip,
+    )
+    precipnoWD = xr.where(
+        (precip["T"] == pd.to_datetime("2000-05-06")),
+        18.7,
+        precipnoWD,
+    )
```
aaron-kaplan, line 575: Indeed this answers my question. (My eye had failed to pick up that 05-05 and 05-06 were different dates.)

I think there's probably a way to change both elements in a single call to `where`. I don't know it off the top of my head.

onset_maproom/tests/test_calc.py
```diff
@@ -4,10 +4,463 @@
...
+                    "2005-02-28T00:00:00.000000000",
+                ],
+                dims=["group"],
+                coords={"group": dts.seasons_ends["group"]},
+            )
+        )
+    ).all()
+
+
+def test_seasonal_onset_date_keeps_returning_same_outputs():
```
aaron-kaplan, line 435: In response to https://github.com/iridl/x24-vector/pull/20#discussion_r729641337 :
> I was not sure what really to test here. 

You are asking the right questions. There are no definitive answers. I agree that this test is fine the way it is, as a canary to alert us to something that changes unexpectedly.

I think this test is a good one to run on real data. The other tests set up a particular scenario and then test for desired behavior in that scenario; those would be better converted to use synthetic data, to make it easier for the reader to undertand the scenario. For this test we're not choosing particular conditions, we're just checking that nothing has changed, so (as you said) it's not clear what synthetic data we would choose.

aaron-kaplan (changes requested):
Ah, this is much easier to understand.

You could reduce the volume of copy-paste by defining this function:
```
def call_onset_date(data):
    calc.onset_date(
        daily_rain=data,
        wet_thresh=1,
        wet_spell_length=3,
        wet_spell_thresh=20,
        min_wet_days=1,
        dry_spell_length=7,
        dry_spell_search=21,
    )
```

onset_maproom/tests/test_calc.py
```diff
@@ -3,11 +3,132 @@
...
+    onsets = onsetsds.onset_delta + onsetsds["T"]
+    assert np.isnat((onsets[0]))
+    assert onsets[1] == pd.to_datetime("2001-03-08T00:00:00.000000000")
+    assert np.isnat((onsets[2]))
+    assert onsets[3] == pd.to_datetime("2003-04-12T00:00:00.000000000")
+    assert onsets[4] == pd.to_datetime("2004-04-04T00:00:00.000000000")
+
+
+def test_seasonal_onset_date():
+    t = pd.date_range(start="2000-01-01", end="2005-02-28", freq="1D")
```
aaron-kaplan, line 72: You start with by constructing a big T variable with five years of daily points, and then you drop most of the points. Why not just start with the more limited variable?
```
t = [
    pd.to_datetime("2000-03-29"),
    pd.to_datetime("2000-03-30"),
...
]
```

onset_maproom/tests/test_calc.py
```diff
@@ -3,11 +3,132 @@
...
+    # this is rr_mrg.sel(T=slice("2000", "2005-02-28")).isel(X=150, Y=150).precip
+    synthetic_precip = xr.DataArray(np.zeros(t.size), dims=["T"], coords={"T": t}) + 1.1
```
aaron-kaplan, line 74: You can add `name="synthetic_precip"` here, rather than letting it use a default name and then renaming it later.

onset_maproom/tests/test_calc.py
aaron-kaplan: I see. I found this: https://stackoverflow.com/questions/10710328/comparing-numpy-arrays-containing-nan

onset_maproom/tests/test_calc.py
remicousin: I don't drop anything. I create wet spells in a 5 year time series.

onset_maproom/tests/test_calc.py
remicousin: xr.where resets the name to None...

remicousin:
This covers what Aaron suggested [here](https://github.com/iridl/x24-vector/pull/19#issuecomment-943496715). Maybe we merge that and I'll give some thought as to whether other tests are welcome.

onset_maproom/tests/test_calc.py
remicousin: I can be more helpful: I create a time series of wet days (1.1 > 1) and then I create wet spells of 3 days * 7mm > 20mm and thus the test tells me that the function finds those synthetic onset dates.

onset_maproom/tests/test_calc.py
```diff
@@ -3,11 +3,144 @@
...
+                    "NaT",
+                    "2001-03-08T00:00:00.000000000",
+                    "NaT",
+                    "2003-04-12T00:00:00.000000000",
+                    "2004-04-04T00:00:00.000000000",
+                ],
+                dims=["T"],
+                coords={"T": onsets["T"]},
+            )
+        ),
```
aaron-kaplan, line 78: I don't see what purpose the DataArray serves here. I think you can construct the same pandas structure directly:
```
pd.to_datetime(
                [
                    "NaT",
                    "2001-03-08T00:00:00.000000000",
                    "NaT",
                    "2003-04-12T00:00:00.000000000",
                    "2004-04-04T00:00:00.000000000",
                ],
)

aaron-kaplan (approved):

onset_maproom/tests/test_calc.py
remicousin: That is mostly driven by the fact that it's unclear to me what the == is testing. Just the values? or also the structure/nature of the array? and if so to what extent?
It does work without the xarray attributes and so made the changes accordingly.
I'm going to merge now, if it lets me.

aaron-kaplan: `NaT == NaT` is false, so you can't use `==` to compare two arrays.

You replaced it with `np.array_equal`. Numpy doesn't support dimension names or coordinate values, so all this compares is the values of the data variable (onset date in this case). If you wanted to compare coordinate values as well, you could have used `pd.testing.assert_frame_equal` or `xr.testing.assert_equal`. According to the doc, `xr.testing.assert_equal` still ignores dimension names.

PR #25, opened 2021-11-05
Macosyml

nitinmagima:
Added macos enviroment


aaron-kaplan (changes requested):
This branch has conflicts, which means you didn't update your local copy of master to the latest version before branching.

Also, please create fbfmaproom/environment_macos.yml, not this file at the top level. Sorry for not being more precise about that.

PR #26, opened 2021-11-05
Macosyml

nitinmagima:
Added macOS environment


aaron-kaplan (changes requested):
Please don't change environment_macos.yml in this PR. Only add fbfmaproom/environment_macos.yml.

aaron-kaplan (changes requested):
Oops, I didn't mean to delete the file, I just meant undo your changes to the file. environment_macos.yml is used by a different project that lives in the same repo.

PR #1, opened 2021-08-11
Second attempt to implement multiple observation datasets

ikhomyakov:
Past discussion and todos [on bitbucket](https://bitbucket.org/iridl/x24-vector/pull-requests/53).

Add NDVI and WRSI columns to the table. Rahel will send us WRSI data. We will need to calculate NDVI spatial averages. Ideally we should render Rain, NDVI and WRSI in a single column with dropbox to switch between Rain, NDVI and WRSI

**REMEMBER TO UPDATE CONFIG IN DLCONFIG WHEN YOU RELEASE**

PR #28, opened 2021-11-15
Remic/sm

remicousin:
This is a function that estimates soil moisture from rainfall, daily, as done in Ethiopia "onset" Maproom, used to then compute a cessation date.

I am not too happy with myself on several accounts:
1/ I can't believe I need so many lines to assign the initial value of SM
2/ I am skeptical that the where within a loop is the most efficient way of doing things... It feels like it needs to scan the whole array with where when the loop is already "restraining" what to look at... That loops back with #1 as I don't seem to know how to assign something without using where (when the dimension name is "unknown" and stored into time_coord)

### DIFF
--- a/onset_maproom/calc.py
+++ b/onset_maproom/calc.py
@@ -38,19 +38,22 @@ def estimate_sm(
         name="soil moisture",
         attrs=dict(description="Soil Moisture", units="mm"),
     )
-    print("rain(0) has a coords T but no more dim T")
-    print(daily_rain.isel({time_coord: 0}))
-    print("thus it seems rain(0) - et broadcasts on et's T")
-    print(daily_rain.isel({time_coord: 0}) - et)
+    # Case et is scalar
+    if np.logical_not(type(et) is xr.core.dataarray.DataArray):
+        et = xr.DataArray(
+            data=[et] * time_coord_size,
+            dims=[time_coord],
+            coords={time_coord: daily_rain[time_coord]}
+        )
     soil_moisture[{time_coord: 0}] = (
-        sminit + daily_rain.isel({time_coord: 0}) - et
+        sminit + daily_rain.isel({time_coord: 0}) - et.isel({time_coord: 0}, missing_dims='ignore')
     ).clip(0, taw)
     # Looping on time_coord
     for i in range(1, time_coord_size):
         soil_moisture[{time_coord: i}] = (
             soil_moisture.isel({time_coord: i - 1})
             + daily_rain.isel({time_coord: i})
-            - et
+            - et.isel({time_coord: i}, missing_dims='ignore')
         ).clip(0, taw)
     return soil_moisture
 
--- a/onset_maproom/calc.py
+++ b/onset_maproom/calc.py
@@ -87,33 +87,6 @@ def onset_date(
     return onset_delta
 
 
-def run_test_onset_date():
-    import pyaconf
-    import os
-    from pathlib import Path
-
-    CONFIG = pyaconf.load(os.environ["CONFIG"])
-    DR_PATH = CONFIG["daily_rainfall_path"]
-    RR_MRG_ZARR = Path(DR_PATH)
-    rr_mrg = read_zarr_data(RR_MRG_ZARR)
-    rr_mrg = rr_mrg.sel(T=slice("2000-01-01", "2000-12-31"))
-    print(
-        onset_date(rr_mrg.precip, 1, 3, 20, 1, 7, 21, time_coord="T")
-        .isel(X=150, Y=150)
-        .values
-    )
-    print(
-        (
-            onset_date(rr_mrg.precip, 1, 3, 20, 1, 7, 21, time_coord="T")
-            + onset_date(rr_mrg.precip, 1, 3, 20, 1, 7, 21, time_coord="T")["T"]
-        )
-        .isel(X=150, Y=150)
-        .values
-    )
-
-
-# run_test_onset_date()
-
 # Time functions
 
 
@@ -265,7 +238,7 @@ def seasonal_onset_date(
             wet_spell_thresh=wet_spell_thresh,
             min_wet_days=min_wet_days,
             dry_spell_length=dry_spell_length,
-            dry_spell_search= dry_spell_search,
+            dry_spell_search=dry_spell_search,
         )
         # This was not needed when applying sum
         .drop_vars(time_coord)
@@ -315,7 +288,7 @@ def run_test_season_onset():
     )
 
 
-#run_test_season_onset()
+# run_test_season_onset()
 
 
 def seasonal_sum(
--- a/onset_maproom/calc.py
+++ b/onset_maproom/calc.py
@@ -38,24 +38,15 @@ def estimate_sm(
         name="soil moisture",
         attrs=dict(description="Soil Moisture", units="mm"),
     )
-    # Case et is scalar
-    if np.logical_not(type(et) is xr.core.dataarray.DataArray):
-        et = xr.DataArray(
-            data=[et] * time_coord_size,
-            dims=[time_coord],
-            coords={time_coord: daily_rain[time_coord]},
-        )
     soil_moisture[{time_coord: 0}] = (
         sminit
-        + daily_rain.isel({time_coord: 0})
-        - et.isel({time_coord: 0}, missing_dims="ignore")
+        + (daily_rain - et).isel({time_coord: 0})
     ).clip(0, taw)
     # Looping on time_coord
     for i in range(1, time_coord_size):
         soil_moisture[{time_coord: i}] = (
             soil_moisture.isel({time_coord: i - 1})
-            + daily_rain.isel({time_coord: i})
-            - et.isel({time_coord: i}, missing_dims="ignore")
+            + (daily_rain - et).isel({time_coord: i})
         ).clip(0, taw)
     return soil_moisture
 
--- a/onset_maproom/calc.py
+++ b/onset_maproom/calc.py
@@ -29,22 +29,16 @@ def estimate_sm(
     with roof and floor respectively at taw and 0 at each time step.
     """
     # Get time_coord info
-    time_coord_index = daily_rain.dims.index(time_coord)
     time_coord_size = daily_rain[time_coord].size
-    # Converts xarray array to numpy array
-    daily_rain_np = daily_rain.to_numpy()
-    # Later need to check if those 2 are arrays or constant
-    et_np = et
-    taw_np = taw
     # Intializing sm
     soil_moisture = [0] * time_coord_size
     soil_moisture[0] = np.maximum(
-        np.minimum(sminit + daily_rain_np[0] - et_np, taw_np), 0
+        np.minimum(sminit + daily_rain.isel({time_coord: 0}) - et, taw), 0
     )
     # Looping on time_coord
     for t in range(1, time_coord_size):
         soil_moisture[t] = np.maximum(
-            np.minimum(soil_moisture[t - 1] + daily_rain_np[t] - et_np, taw_np), 0
+            np.minimum(soil_moisture[t - 1] + daily_rain.isel({time_coord: t}) - et, taw), 0
         )
     # Converts back to xarray
     soil_moisture = xr.DataArray(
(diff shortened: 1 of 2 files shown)


onset_maproom/calc.py
```diff
@@ -13,6 +13,52 @@ def read_zarr_data(zarr_path):
...
+    """
+    # Creating a soil_moisture array
+    soil_moisture = daily_rain * 0
+    # Intializing sm
+    soil_moisture = xr.where(
+        (soil_moisture.coords[time_coord] == soil_moisture.coords[time_coord][0]),
+        (
+            sminit
+            + daily_rain.sel(**{time_coord: daily_rain.coords[time_coord][0]})
+            - et
```
aaron-kaplan, line 39: In your test case `et` is a scalar, so this works, but the comment says that we expect it to be an array with dimension `t`, in which case shouldn't this be something like `et.sel(daily_rain.coords[time_coord][0])`?

aaron-kaplan (changes requested):
Sorry for losing track of this.

xarray/numpy are not well suited to this kind of calculation where each value depends on the previous value. I suggest you rewrite it in plain python without xarray, something like
```
n = len(daily_rain)
soil_moisture = [0] * n
soil_moisture[0] = sminit + daily_rain[0] - et
for i in range(1, n):
    soil_moisture[i] = soil_moisture[i-1] + daily_rain[0] - et
```
(where I'm assuming that `daily_rain` is a python list or maybe a numpy array, but not an xarray DataArray, so you'll have to convert it first). Once you get that working with good test coverage, including the case where `et` varies with time if that's what you intend, we'll see how fast it is. If it's too slow, we can try speeding it up with some tips from [here](https://stackoverflow.com/questions/4407984/is-it-possible-to-vectorize-recursive-calculation-of-a-numpy-array-where-each-el).

remicousin:
Or do I do a numpy 1D version then later wrap it into a numpy/xarray multi-dim version? Like Igor was doing with his first attempt to onset date?

aaron-kaplan:
Ah, I had forgotten about the other dimensions.

Looks like [DataArray.dims()](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.dims.html) gives you an ordered list of dimension names. Then `da.dims().index(time_var)` should give you the index of the time dimension.

I think doing it this way would be faster than doing a 1D version and then wrapping it with `vectorize`. I think that's what Igor's version did, but I'm not sure (and not taking the time to look right now).

remicousin:
ok. I don't think Igor went as far as the vectorizing steps.

Rémi Cousin (he/him/...)
***@***.***
Phone: +1 8456804422
Fax: +1 8456804864
http://iri.columbia.edu
61 route 9W, Monell Building
Palisades, NY 10964

remicousin:
Just a note as I am working on this and to follow through the aborted discussion during the meeting. I am under the impression that 1-D, 2-D and n-D (n > 2) arrays are specific cases in numpy and so I was under the impression that a 1/2-D case could be very different than other n-D ones...

aaron-kaplan:
I’m not aware of this difference. What’s your source?

remicousin:
I am probably not reading documentation well... I treated the simple case where:

1. daily_rain depends only on time_coord
2. et and taw are constants

Later, need to deal with case:

1. daily_rain can have more dims than time_coord

Later, can deal with cases where:

1. et depends on time_coord and other dimensions;
2. taw depends on dimensions other than time_coord

aaron-kaplan:
Is "total available water" standard terminology in the field? As an outsider, I would find it clearer if that parameter had a name that included something like "max" or "capacity."

onset_maproom/calc.py
```diff
@@ -13,6 +13,49 @@ def read_zarr_data(zarr_path):
...
+    total available water and
+    intial soil moisture value, knowing that:
+    sm(t) = sm(t-1) + rain(t) - et(t)
+    with roof and floor respectively at taw and 0 at each time step.
+    """
+    # Get time_coord info
+    time_coord_index = daily_rain.dims.index(time_coord)
+    time_coord_size = daily_rain[time_coord].size
+    # Converts xarray array to numpy array
+    daily_rain_np = daily_rain.to_numpy()
```
aaron-kaplan, line 35: `to_numpy` was introduced in xarray 0.19.0, released in July 2021. `environment_linux.yaml` specifies `xarray=0.18.2`, so this fails for me. But `environment_macos.yaml`, having been built later, specifies `xarray=0.19.0`, so it works for you. We need to (a) find a way to keep linux and macos environments more in sync, but also (b) set up automated test runs in github so that when a developer does something that works under macos but not under linux, they're automatically notified soon after.

aaron-kaplan (commented):
This version served its purpose, which was to be an easy-to-read implementation of a simple case, which we can extend subsequently.

After submitting this review I'll push a few commits, including one that generalizes the function to multiple dimensions.

aaron-kaplan:
Note that there are conflicts with Drew's branch that was merged last week. Please rebase.

onset_maproom/calc.py
remicousin: et should be same structure than daily_rain/soil_moisture. It's a scalar in Eth Maproom case. Shall I try cover the other cases?

aaron-kaplan:
Sure, if it's not hard, and it doesn't seem like it would be.

>

remicousin:
oh you did all the work...
Well I misunderstood what you meant when you said we had to convert to numpy... which was probably tied to the fact that I never figured out that
soil_moisture[{time_coord: 0}] =
would assign soil_moisture for first time step. (Sad...) which in turn made me use 'where' like crazy.

Yes "total available water" is the consecrated term even though not intuitive. Maybe there is a bigger story behind it.

So I guess either that part is done, or I try to cover the case where et is not scalar. Possibly in the future, taw is not scalar either but it's constant in time (typically varies in space and possibly whatever other dimensions (some ensemble member or what-have-you).

remicousin:
I'll try... I am less optimistic now...

remicousin:
After that piece concludes, should be able to move on to cessation date, but will have to lower priority on that.

onset_maproom/calc.py
```diff
@@ -38,17 +38,25 @@ def estimate_sm(
         name="soil moisture",
         attrs=dict(description="Soil Moisture", units="mm"),
     )
-
+    # Case et doesn't depend on time_coord
+    if np.logical_not(type(et) is xr.core.dataarray.DataArray):
```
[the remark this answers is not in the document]
remicousin: I did this because isel({time_coord: i}) applied to et without a time_coord returned an error. But maybe there is a more elegant way to deal with that?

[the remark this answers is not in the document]
aaron-kaplan: Since et and daily_rain have the same dimension name and coordinate values for time, I think you can just do
```
        sminit + daily_rain.isel({time_coord: 0}) - et
```
and it will use the appropriate values from et and ignore the rest.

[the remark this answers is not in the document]
remicousin: didn't think of that... that makes all sense.

But it doesn't seem to behave that way... See last commit where et_has_T test fails and I included prints in the main function that I think describe the behavior.

Maybe there is an isel option to keep the dim?

remicousin:
I pass the tests by treating the case scalar only where I make the scalar an array of size T.
Then I use isel's missing_dims to cover the case where et is an array but doesn't depend on T.

I am not pushing that just yet in case you want to see what I previously pushed, as your solution seemed intuitively right but what I printed intuitively wrong

aaron-kaplan:
> I am not pushing that just yet in case you want to see what I previously pushed, as your solution seemed intuitively right but what I printed intuitively wrong

Go ahead and push it. When I get time to look at it, I can check out the previous commit to try it for myself.

remicousin:
maybe we should just do
(rain - et).isel({time_coord: i})
?
That would work in all cases. But it's still odd that your suggestion behaved as it did

remicousin:
Ok I pushed that last version that doesn't need checking on et object type.

onset_maproom/calc.py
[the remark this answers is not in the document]
aaron-kaplan: The [doc](https://xarray.pydata.org/en/stable/user-guide/computation.html#automatic-alignment) on alignment supports your diagnosis.
> Xarray enforces alignment between index Coordinates (that is, coordinates with the same name as a dimension, marked by *) on objects used in binary operations.

`T` is not an index coordinate in `daily_rain.isel(T=0)`, so it doesn't get aligned.

Your solution (subtract `daily_rain - et` before indexing) is succinct but probably inefficient. It repeats the same subtraction operation between the same two X * Y * T arrays on every iteration of the loop. To avoid that, we could pull out the subtraction operation and do it just once, before entering the loop. Alternatively, it looks like [expand_dims](https://xarray.pydata.org/en/stable/generated/xarray.DataArray.expand_dims.html) could convert the scalar `T` back to an index coordinate.

Sorry, I know it's annoying to wait five days to get feedback. If there are more things to discuss we should schedule time to work through them together.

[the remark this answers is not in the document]
remicousin: no problem. I can't really focus much on this these days.
I pushed the delta solution that seemed the most appropriate to me.

onset_maproom/calc.py
```diff
@@ -13,6 +13,45 @@ def read_zarr_data(zarr_path):
 # Growing season functions
 
 
+def estimate_sm(
```
aaron-kaplan, line 16: Rename `estimate_soil_moisture`, or just `soil_moisture`?

If you expect this function will be useful beyond the cessation date calculation, I think it should have a more self-explanatory name. If it's just an internal detail of the cessation date implementation then I don't care too much.

aaron-kaplan (approved):

onset_maproom/calc.py
remicousin: This is actually the basic algo of the Ingrid water balance. See W(t)= [section](http://iridl.ldeo.columbia.edu/dochelp/Documentation/details/index.html?func=:Water_Balance). The Ingrid water balance just does a lot more stuff to what goes into this basic algo. So it actually makes sense to call this water_balance that could be called by other more elaborated water_balance functions later. Or a same function with options that trigger this or that element of the water_balance?

You may see that the recurrence relationship involved ET (or ETcrop_red) that is actually dependent on SM(t-1) so it means intervention in the loop. Except that actually in Ingrid, I loop only to get SM and ETcrop_red is actually retrieved vectorially after the fact (as it becomes an output of water_balance among several others besides SM). So the loop might be fine as it is.

I guess my point is that it would make sense to name the function water_balance, like Ingrid's, and that it can expand later with the other parts of the algo, that can be triggered with optional arguments (I seem to see that this is a fine thing to do in Xarray & Co); with the slight drawback that some rewriting might be needed to accommodate those options.

Makes sense?

And also in that case that function will likely be used by much more than cessation_date. It will probably be desired by PIs sooner than later...

aaron-kaplan: sounds good.

remicousin: sorry... moving in small steps here. Should I make the output a dataset then of which soil_moisture which be a variable? Assuming that later, there will be more variables output in the dataset created by water_balance function?

aaron-kaplan: Sure, if you anticipate that it will return multiple variables in the future, it would be good to return a Dataset so that won't have to be a breaking change.

Out of curiosity, what other variables?

remicousin: See the list in the ingrid func doc, just before the Notes. Below links to examples. Basically outputs all the different elements of the water balance. You can also look at what Maprooms using this look like (it's slow).

http://iridl.ldeo.columbia.edu/maproom/Water/BAWKP/Agriculture/Bicol_WB_Dry.html?region=irids%3Ahome%3Aremic%3APhilippines%3ABicol%3AsoilTypes%3Agid%40185%3Ads

http://maproom.meteorwanda.gov.rw/maproom/Agriculture/Histo_Agri/WB_BB_MAM.html?region=bb%3A29.56875%3A-2.6812500000000004%3A29.606250000000003%3A-2.6437500000000003%3Abb

onset_maproom/calc.py
```diff
@@ -49,7 +49,9 @@ def estimate_sm(
             soil_moisture.isel({time_coord: i - 1})
             + delta_rain_et.isel({time_coord: i})
         ).clip(0, taw)
-    return soil_moisture
+    water_balance = xr.Dataset().merge(soil_moisture)
+    print(water_balance)
```
aaron-kaplan, line 53: remove the print

aaron-kaplan (approved):

PR #29, opened 2021-11-19
Exclude years with missing data from counts in confusion matrix

aaron-kaplan:


aaron-kaplan:
@nitinmagima this is FYI, but with Igor gone and Kevin not up to speed yet I'm not going to wait for review before merging.

PR #34, opened 2021-12-14
Drewr/outage points

drewmresnick:
working on displaying outage data as point layer on map

### DIFF
--- a/PRISM_maproom/layout.py
+++ b/PRISM_maproom/layout.py
@@ -69,7 +69,7 @@ style_handle = assign(
 
 point_to_layer = assign("""function(feature, latlng, context){
     style = {
-        radius: 8,
+        radius: 5,
         zIndex: 1000,
         color: 'black',
         opacity: 0.7,};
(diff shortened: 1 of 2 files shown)


PRISM_maproom/layout.py
```diff
@@ -66,6 +67,15 @@
 )
 
 
+#point_to_layer = assign("""function(feature, latlng, context){
+#    style = {
+#        radius: 5,
+#        zIndex: 1000,
+#        color: 'black',
+#        opacity: 0.7,};
```
aaron-kaplan, line 75: I finally figured out how to make the CircleMarkers stay on top even after mouseover events. Add to this `style` object the following properties:
```
pane: 'markerPane',
interactive: false
```
`pane` makes it stay on top, and `interactive` makes mousing over the marker behave the same as mousing over the polygon below it.

BTW this object should really be called `options`, not `style`. It's not a list of CSS style properties. See [doc](https://leafletjs.com/reference.html#circlemarker).

remicousin:
Taking over as Drew might be offline for some part of the day.

drewmresnick:
i think i have gotten it to a point that I am satisfied with for today. let me know if you see anything glaring

aaron-kaplan:
If it works then it's good enough for a demo. If you want to merge to
facilitate collaboration with Rémi, feel free. Keep track of my comments
somewhere so you can address them later.

remicousin:
ah ok... I will stop messing around then

remicousin (approved):

drewmresnick:
@remicousin if you prefer to mess around on this branch go ahead, I won't be making any more edits to it now. Just polishing the poster at this point. So it is up to you if you would like to keep playing around on this branch or not.

PR #36, opened 2021-12-17
Remic/finalish

remicousin:


PRISM_maproom/diversity_maproom.py
```diff
@@ -172,17 +171,14 @@ def colorMap(date, myspecies):
...
 )
 def outagePoints(date):
-    dfLoc = dfJoined.loc[dfJoined["date"] == date]
-    dfLoc = dfLoc[["geometry", "eBird.DP.RF"]]
-    dfLoc = dfLoc.rename(columns={"eBird.DP.RF": "diversity"})
     # outage data
     dateSince = datetime.strptime("2010-01-01", "%Y-%m-%d")
     dateSelect = datetime.strptime(date, "%Y-%m-%d")
-    dateDiff = (dateSince - dateSelect).days  # beginning of the day we are looking at
+    dateDiff = (dateSelect - dateSince).days  # beginning of the day we are looking at
```
aaron-kaplan, line 177: How was it not obviously broken before if dateDiff was negative?

remicousin:
I am going to approve and merge this right away as we are running out of time.
First change was actually a mistake that was mis-selecting outage data with respect to week of interest.
2nd change was an improvement to select all events that overlap with the week.
3rd change probably didn't change anything. I was actually trying to do more and plot a time series of outages with the bird data, but didn't get that far. The change was because I was confused by the joining with the whole bird data rather than just the MAtown locations. Because there are several species in there so not clear if that duplicates events also or what... Anyway. As I said didn't get that far...

I was generally confused from the beginning by the indexing of bird data against towns, as they also depend on time and species. But that's another story

PRISM_maproom/layout.py
```diff
@@ -283,6 +290,18 @@ def controls_layout():
                 """
                 )
             ),
+            html.H5("Dataset Sources in the IRI DL"),
+            html.P(
+                dcc.Markdown(
```
aaron-kaplan, line 295: `dcc.Markdown` looks handy. Instead of several `Markdown` interspersed with `H5` and `P`, we could put all this text in a single `Markdown`. (Not before the demo.)

PRISM_maproom/diversity_maproom.py
remicousin: dateDiff can be negative. It's fine if the date is prior to 2010. So no reason to break but actually since the outage data starts in 2013, dateDiff for 2005 was positive and actually selecting outage data in 2015 I guess

remicousin:
I see Aaron is looking into it now. I am going to see the sun for a bit. and merge when I come back.

aaron-kaplan (approved):
Some of these are things I wouldn't choose to change just before a demo, but if you're confident, have at it!

drewmresnick (approved):

PR #38, opened 2022-01-12
Dockerized onset-maproom based on bdfloodsmr

kgraaf:
This PR allows for the onset maproom to be dockerized. One thing I'm a little unclear on though is that mounting the the `/data/` directories doesn't work consistently. 

If I run the application with 

    sudo docker run -p 0.0.0.0:8589:8050 -v /data:/data -v /Data:/Data -it omr

I have to `ls` the relevant directories or execution will fail (complaining they're inaccessible). I'm not sure why that is.


onset_maproom/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN eval "$('/conda/bin/conda' 'shell.bash' 'hook' 2> /dev/null)" && \
+    conda env create -f /build/environment.yml -n app
+
+# mod_wsgi: use pip to compile mod_wsgi from source for the particular versions
+# of apache and python that we're using.
+RUN eval "$('/conda/bin/conda' 'shell.bash' 'hook' 2> /dev/null)" && \
+    conda activate app && \
+    pip install mod_wsgi==4.7.1
+
+RUN yum -y install mesa-libGL
```
kgraaf, line 30: This had to be installed manually and I couldn't figure out which conda package would add it. This might be non-ideal (since it's versioned separately) or it might not matter.

onset_maproom/README.md
```diff
@@ -55,6 +55,13 @@ The widgets module contains a few functions of note:
 * `Number()`: This is a component for a number selector. The first argument is the HTML id,
    the second and third are the lower and upper bound respectively.
 
+# Docker Build Instructions
+
+To build the docker image, we have to use a work around so that pingrid.py will be included correctly, as
+docker doesn't normally allow files above the working directory in the hierarchy to be included
+
+    $ tar -czh . | sudo docker build -t <desired image name> -
```
kgraaf, line 63: symlinking pingrid is somewhat problematic for building dockerfiles. If this workaround is fine, then we can just forge ahead as is

onset_maproom/Dockerfile
aaron-kaplan: This is how we do it in fbfmaproom. https://bitbucket.org/iridl/x24-vector/src/ca47ae24bdbcdb2dbc5d6ba0a9e2f8e54a595936/fbfmaproom/Dockerfile#lines-5:7

onset_maproom/README.md
aaron-kaplan: Here's how we deal with it in fbfmaproom: https://bitbucket.org/iridl/x24-vector/src/ca47ae24bdbcdb2dbc5d6ba0a9e2f8e54a595936/fbfmaproom/Dockerfile#lines-40

onset_maproom/docker/app.wsgi
```diff
@@ -0,0 +1,2 @@
+import maproom
+application = maproom.APP.server
```
aaron-kaplan, line 2: Unrelated to your PR, allow me to vent my frustration at dash for basing their product on flask but refusing to use flask's terminology. The thing flask calls an "app" gets renamed a "server" by dash, and dash invents its own thing which it calls "app".

onset_maproom/README.md
aaron-kaplan: BTW symlinking pingrid is problematic for other reasons too. I had problems with it when setting up pytest. Let's discuss alternatives (but not on this PR).

aaron-kaplan (changes requested):
- [ ] Handle pingrid.py the way we do in fbfmaproom so we don't need to customize the build command.
- [ ] Add a release script. There are two approaches to this that are in use in other projects; let's discuss when we speak.
- [ ] Remove the dlcharts parts of the application.
- [ ] Don't set CONFIG in Dockerfile

I have some other things I'd like to change but after we get it running in production. This is time sensitive.

onset_maproom/README.md
kgraaf: Has this been tested? I tried doing just that and Docker didn't actually resolve the symlink

onset_maproom/Dockerfile
kgraaf: Yeah, no reason to not bundle that with the first yum command

aaron-kaplan:
I don't know what the problem is with the network volume mounts, but this is the way serveral other apps do it, so hopefully it won't be a problem in production.

aaron-kaplan:
Oh, I think I do know what the problem is. You only have this problem with /data, not with /Data, right? On devi /data/* are autofs mounts, i.e. they get mounted on demand, not at boot, and they get unmounted automatically after some idle period.

BTW it's best to add `:ro` to the -v arguments to mount them read-only.

onset_maproom/README.md
aaron-kaplan: Ah, you're right. Yes the fbfmaproom Dockerfile works, but it's written to expect the parent directory as context. That information is embedded in the release script. Your way is better, because it won't force a rebuild when the only changes are in other applications that share the repo.

As I said elsewhere, you need to add a release script, and we'll discuss that on our next call.

onset_maproom/docker/httpd.conf
```diff
@@ -0,0 +1,25 @@
+LoadModule mpm_prefork_module modules/mod_mpm_prefork.so
+LoadModule unixd_module modules/mod_unixd.so
+LoadModule authz_core_module modules/mod_authz_core.so
+LoadModule log_config_module modules/mod_log_config.so
+
+User apache
+Group apache
+Listen 8050
```
aaron-kaplan, line 8: Whoops, missed this until I read the dlconfig PR. Please use port 8000 like all the other apps do.

aaron-kaplan: Has anything else in this file been edited from the BDFloodsMR version?

onset_maproom/docker/httpd.conf
```diff
@@ -0,0 +1,25 @@
...
+
+ErrorLog "/dev/stderr"
+LogLevel warn
+LogFormat "%h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" combined
+CustomLog "/dev/stdout" combined
+
+LoadModule wsgi_module /conda/envs/app/lib/python3.9/site-packages/mod_wsgi/server/mod_wsgi-py39.cpython-39-x86_64-linux-gnu.so
+WSGIPythonHome "/conda/envs/app"
+WSGIScriptAlias / /app/docker/app.wsgi
+WSGIDaemonProcess maproom processes=1 threads=10 maximum-requests=1000 python-path=/app
```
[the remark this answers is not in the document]
aaron-kaplan: I remember. It was because of the caching we do with `@lru_cache` in fbfmaproom. The onset maproom doesn't seem to use `@lru_cache`, so maybe we can safely increase the number of processes here. On the other hand, maybe Igor's `@lru_cache` hack would be a quick and dirty way to get acceptable performance on the raster layer (pretty sure that's what it's used for in fbfmaproom2).

Check that out after we get this app published.

onset_maproom/docker/httpd.conf
kgraaf: nothing besides the name fbfmaproom which doesn't really appear to matter

onset_maproom/README.md
aaron-kaplan, line 63: mention the release script in this section.

kgraaf:
- [ ] remove leading whitespace
- [ ] add docker build method from readme
- [ ] add release script to readme
- [ ] remove rest of legacy ingrid-related code from app
- [ ] remove branch option in release script

aaron-kaplan (approved):

PR #39, opened 2022-01-20
Drewr/PRISMimprovements

drewmresnick:
working on improvements for PRISM maproom. The current task list compiled from previous PR's is as follows

- [ ] Remove copy of widgets.py (move to parents directory and create symlink
- [ ] add error message to plot when no data selected
- [ ] improve on category definition (ctg) for color scale
- [ ] improve SQL query
- [ ] move loading of data from top level into a callback function
- [ ] address centroid warning
- [ ] add error message when date selected is out of range of outage data
- [ ] improve def of colorConditions for choropleth map

### DIFF
--- a/PRISM_maproom/diversity_maproom.py
+++ b/PRISM_maproom/diversity_maproom.py
@@ -152,10 +152,6 @@ def colorMap(date, myspecies):
     ]
     dfLoc["color"] = np.select(colorConditions, colorscale)
     ctg = [str(i) for i in classes]
-    print(ctg)
-    print(type(ctg))
-    print(classes)
-    print(type(classes))
     colorbar = dlx.categorical_colorbar(
         categories=ctg,
         colorscale=colorscale,
--- a/PRISM_maproom/diversity_maproom.py
+++ b/PRISM_maproom/diversity_maproom.py
@@ -89,6 +89,9 @@ def updateCityDD(feature):
 )
 def update_timeSeries(city, myspecies):
     cityString = city
+    if cityString is None:
+        errorFig = pgo.Figure().add_annotation(x=2, y=2,text="No Data to Display. Select a city.",font=dict(family="sans serif",size=30,color="crimson"),showarrow=False, yshift=10, xshift=60)
+        return errorFig
     dfCity = dfJoined.where(dfJoined["species"] == myspecies).loc[
         dfJoined.index == cityString
     ]
@@ -166,6 +169,7 @@ def colorMap(date, myspecies):
 @APP.callback(
     Output("outagePoints", "data"),
     Output("outageTable", "children"),
+    Output("dateWarning", "children"),
     [Input("date_dropdown", "value")],
 )
 def outagePoints(date):
@@ -180,6 +184,9 @@ def outagePoints(date):
         SQL_CONN,
         params={"dateDiff": dateDiff, "dateDiff2": dateDiff2},
     )
+    if outageDF.empty:
+        warning = dbc.Alert("No available outage data for selected date.", color="warning")
+        return None, None, warning
     outageDF["outageCount"] = outageDF["geo_id"].map(outageDF["geo_id"].value_counts())
     outageDF = (
         outageDF.rename(columns={"city_town": "city"}).set_index("city").sort_index()
@@ -208,7 +215,7 @@ def outagePoints(date):
         fixed_rows=dict(headers=True),
     )
     toJSON2 = json.loads(mergedDF.to_json())
-    return toJSON2, outageTable
+    return toJSON2, outageTable, None
 
 
 if __name__ == "__main__":
--- a/PRISM_maproom/diversity_maproom.py
+++ b/PRISM_maproom/diversity_maproom.py
@@ -151,9 +151,11 @@ def colorMap(date, myspecies):
         "#4a148c",
     ]
     dfLoc["color"] = np.select(colorConditions, colorscale)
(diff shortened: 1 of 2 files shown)


remicousin:
I actually noted down some improvements/questions to bring about for this Maproom. Should I list them here or what's the best way to share them? Some are likely trivial and some possibly have a chance to more profoundly affect existing code.

drewmresnick:
> I actually noted down some improvements/questions to bring about for this Maproom. Should I list them here or what's the best way to share them? Some are likely trivial and some possibly have a chance to more profoundly affect existing code.

@remicousin You can list them here if you'd like for now. The bigger discussion would be which tasks you would like me to prioritize with my remaining time as I do not have much time left on this project. So maybe if you add yours we can then take a look together and see where to go from there?

remicousin:
Okey. Well the part I was least comfortable for is that eBird is basically 3D data(time, space, species) and thus I am wondering if pandas is a good candidate to manipulate it. And thus I am not too clear what indexing does in that case, at least until 2 of the 3 dimensions have been selected at a point.

remicousin:
In layout.py, we are reading file(s) to get cities and dates to put in menu. But we are also reading that info from files in the maproom.py but that one is importing layout so we can't import it into layout to get that data. That sounds odd to me. We should be able to read data files once regardless of the set of .py files we are using, shouldn't we? I am encountering this "problem" elsewhere btw. I feel it's likely a trivial problem (like having always a 3rd .py for reading/calculations?) but it needs to be done cleanly as it will be a trivial problem with bigger consequences when we start to read bigger files; or make more complicated maprooms.

remicousin:
The most interesting feature to add to this Maproom to me is to overlay the outage time series with the eBird ones. But we probably want to do the clean-up tasks, and the improvement tasks that teach us something useful rather than  that first.

aaron-kaplan:
> In layout.py, we are reading file(s) to get cities and dates to put in menu. But we are also reading that info from files in the maproom.py but that one is importing layout so we can't import it into layout to get that data.

As I've mentioned before, we need to stop reading data at the top-level of python files. Do it in a function, and call that function from each dash callback that needs the data. If layout.py and maproom.py both need that function, put it in a third file that they both import. (I'm not sure it really makes sense for layout.py to load any data, but maybe.)

remicousin:
ok about functions that read/calculate data. Does it make sense though to have all that happen in callbacks? What if calculations are needed by several callbacks? Does it not make sense to have the common reading/calculation done before and out of callbacks?
This is a typical scenario actually as Maprooms are often a map and a time series of the same analysis respectively further analyzed in time and space.

aaron-kaplan:
> Does it not make sense to have the common reading/calculation done before and out of callbacks?

Your intuition is sound, but loading data when the module is imported is not a good way to implement it, because it has undesirable side effects. For example, if you load the data at import time then every unit test will load the data, even ones that don't use it, which can slow down the test suite by a couple orders of magnitude.

In fbfmaproom, Igor added `@functools.lru_cache` to some functions that read a lot of data. This wraps the function such that the data get read from disk the first time, and the result is stored in memory and returned quickly on subsequent calls.  This approach has its own disadvantages, and won't scale up, but if you need a quick fix it might do for now.

Don't add caching until you've tried without it and seen that it was unacceptably slow.

remicousin:
Just added @kgraaf to this PR. It turns out that Drew doesn't have resources left to keep going with this. It also turns out that @kgraaf also spent all his PRISM allocations. But Kevin's spending on projects timing is not as meaningful since what he does typically contributes to all projects. So, in light of this, I suggest we keep this PR open as a potential source of inspiration for potential work for @kgraaf in general.

Makes sense?

drewmresnick:
> Just added @kgraaf to this PR. It turns out that Drew doesn't have resources left to keep going with this. It also turns out that @kgraaf also spent all his PRISM allocations. But Kevin's spending on projects timing is not as meaningful since what he does typically contributes to all projects. So, in light of this, I suggest we keep this PR open as a potential source of inspiration for potential work for @kgraaf in general.
> 
> Makes sense?

Yes, this PR could also be an opportunity to test out some of the things we are discussing for maprooms with multiple datasets ie working to make interactive points that are sized according to their data values

aaron-kaplan:
This maproom is a project deliverable, right? By when does it need to be published? What else needs to happen before we can publish it?

remicousin:
I think for this NSF project, the notion of deliverable is loose. We already have a poster at AGU about this and that is probably good enough. Thus doing more on this would mostly: tease the desire of PRISM colleagues to keep on working with us in the future; make us look better vis-à-vis NSF; and mostly what I suggested: gives us a sandbox to address more generic things we want to develop of which this Maproom is an instance.

Regardless of all that, I think it is publishable as it is now. We can also publish it in the Summer so that it's part of any final report to NSF. The project ends in October.

aaron-kaplan:
This PR doesn't actually do much yet. @drewmresnick and @remicousin could you just put in a little extra time to finish reviewing the part that's been done, and merge it? Then we can move the remaining todo items somewhere else, e.g. github issues or a TODO file in this directory. 

Long-running branches are a liability. They get out of sync with master and conflicts develop. Keeping changes small and merging them quickly helps us all stay in sync.

remicousin (approved):
Pending changes look good to me. Thus I approve and we'll move the to-do list as Aaron suggested

drewmresnick:
merging current working copy and moving to-do items elsewhere for further possible dev

PR #41, opened 2022-01-26
Drewr/cessDate

drewmresnick:
to create cess date algorithm and add output as additional tab in existing onset/cessation maproom


onset_maproom/cess.py
```diff
@@ -0,0 +1,133 @@
...
+
+    end_month = first_end_date.dt.month.values
+
+    # Apply daily grouping by season
+    grouped_daily_data = calc.daily_tobegroupedby_season(
+        daily_rain, search_start_day, search_start_month, end_day, end_month
+    )
+    # Apply onset_date
+    seasonal_data = (
+        grouped_daily_data #[daily_rain.name]
```
aaron-kaplan, line 98: The code that you commented out in this line is what extracts a single DataArray from the Dataset. Since you commented it out, `cess_date` is getting mapped over a Dataset instead of a DataArray.

onset_maproom/cess.py
```diff
@@ -0,0 +1,133 @@
...
+daily_rain = rr_mrgSub.isel(X=33,Y=3)
+
+
+#onset date calc used in seasonal cess date func
+#first date after START_DAY(1), START_MONTH(9) in SEARCH_DAYS(90) 
+#days when the soil water balance falls below DRY_THRESH(5)mm 
+#for a period of DRY_SPELL_LENGTH(3) days
+def cess_date(daily_rain, dry_thresh, min_dry_days, et, taw, sminit, time_coord="T"):
+    print(daily_rain)
+    daily_rain = daily_rain.to_array()
```
aaron-kaplan, line 33: According to the docs, `to_array` broadcasts all of the variables against each other. That's not what you want. As I said in the other comment, if you want to extract a single variable from a Dataset, just use `ds["varname"]`.

drewmresnick:
@remicousin Can you look at the cess date function I wrote and see if the math I did was correct?

remicousin:
Maybe add some tests so it will be easier to review?

drewmresnick:
> Maybe add some tests so it will be easier to review?

@remicousin Sure I can also work on that. I was hoping just generally for you to check the calculations I did and if they match the calculation for cessation date in the original maproom.

remicousin:
> > Maybe add some tests so it will be easier to review?
> 
> @remicousin Sure I can also work on that. I was hoping just generally for you to check the calculations I did and if they match the calculation for cessation date in the original maproom.

I'll look into results once you addressed the couple issues I pointed out.
Then, generally speaking, I prefer to test in an incremental fashion rather than jump all the way to comparing with the Ingrid function in that case.
What I've started doing is building the tests entries as I write the code I am writing, instead of printing in code or what have you, and I "save" or "freeze" a "print" to become an official test when it seems relevant, or reaching some milestone. This way, the tests are kind of logically what I actually want to verify when writing; and it's not really any additional work because I do it anyways when coding.

drewmresnick:
> > > Maybe add some tests so it will be easier to review?
> > 
> > 
> > @remicousin Sure I can also work on that. I was hoping just generally for you to check the calculations I did and if they match the calculation for cessation date in the original maproom.
> 
> I'll look into results once you addressed the couple issues I pointed out. Then, generally speaking, I prefer to test in an incremental fashion rather than jump all the way to comparing with the Ingrid function in that case. What I've started doing is building the tests entries as I write the code I am writing, instead of printing in code or what have you, and I "save" or "freeze" a "print" to become an official test when it seems relevant, or reaching some milestone. This way, the tests are kind of logically what I actually want to verify when writing; and it's not really any additional work because I do it anyways when coding.

I am not seeing an issues to address other than adding tests? Are there other things you mentioned that I am not seeing?

remicousin (changes requested):

remicousin:
Looks like I had to click "Submit reviews"

onset_maproom/cess.py
```diff
@@ -0,0 +1,129 @@
...
+rr_mrgSub = rr_mrg.sel(T=slice('1981-01-01','1983-12-31'))
+daily_rain = rr_mrgSub.isel(X=200,Y=160)
+
+
+#onset date calc used in seasonal cess date func
+#first date after START_DAY(1), START_MONTH(9) in SEARCH_DAYS(90) 
+#days when the soil water balance falls below DRY_THRESH(5)mm 
+#for a period of DRY_SPELL_LENGTH(3) days
+def cess_date(daily_rain, dry_thresh, min_dry_days, et, taw, sminit, time_coord="T"):
+    water_balance = calc.water_balance(daily_rain, et,taw,sminit,"T")
```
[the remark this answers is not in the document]
drewmresnick: are you saying

change input name so `cess_date(soil_moisture, etc)` instead of the input var being called 'daily_rain' 
or 
do you actually want me to first calculate `soil_moisture` outside of `cess_date()` and then use that output as the input for `cess_date()`

onset_maproom/cess.py
```diff
@@ -0,0 +1,129 @@
...
+    dry_day = water_balance < dry_thresh
+    first_dry_day = dry_day * 1
+    first_dry_day_roll = first_dry_day.rolling(**{time_coord: min_dry_days}).sum() >= min_dry_days
```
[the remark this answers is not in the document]
drewmresnick: Should I call it dry_spell then? Or something else entirely?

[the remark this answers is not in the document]
remicousin: yeah... why not...

onset_maproom/cess.py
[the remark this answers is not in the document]
drewmresnick: @remicousin could you clarify this for me?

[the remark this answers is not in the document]
remicousin: I'll try :) basically both. cess_date should be a function of soil moisture rather than rainfall. Users could use whatever to get soil_moisture to feed into it thus yes, the name of the input should change accordingly. And the arguments that belong to water_balance should disappear from cess_date.
We can talk if easier. I am around, even though there are a bunch of meetings this afternoon.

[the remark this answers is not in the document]
drewmresnick: I think I understand what you are saying!

onset_maproom/cess.py
```diff
@@ -0,0 +1,129 @@
...
+    cess_delta = cess_mask.idxmax(dim=time_coord)
+    cess_delta = (
+        cess_delta
+    # offset relative position of first wet day
+    # note it doesn't matter to apply max or min
+    # per construction all values are nan but 1
+        - (
+            min_dry_days
+            - 1
+            - first_dry_day.where(first_dry_day[time_coord] == cess_delta).max(
```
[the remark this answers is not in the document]
drewmresnick: I just pushed changes however for this, I was having a difficult time subtracting (min_dry_days - 1) from cess_delta; it kept subtracting nanoseconds. My attempts to fix it so far have not been successful except to leave in this line in lieu of a constant 1 

`dry_spell.where(dry_spell[time_coord] == cess_delta).max(
                dim=time_coord)`

Since that will always give a value of 1 anyways..

Not necessarily the smartest fix but for now it worked so I went with it..

onset_maproom/cess.py
```diff
@@ -20,38 +20,36 @@
...
+daily_rain = rr_mrgSub.sel(X="37.75",Y="9", method="nearest", tolerance=0.04)
 
 
 #onset date calc used in seasonal cess date func
 #first date after START_DAY(1), START_MONTH(9) in SEARCH_DAYS(90) 
 #days when the soil water balance falls below DRY_THRESH(5)mm 
 #for a period of DRY_SPELL_LENGTH(3) days
-def cess_date(daily_rain, dry_thresh, min_dry_days, et, taw, sminit, time_coord="T"):
-    water_balance = calc.water_balance(daily_rain, et,taw,sminit,"T")
+def cess_date(water_balance, dry_thresh, min_dry_days, et, taw, sminit, time_coord="T"):
```
remicousin, line 31: let's call the input soil_moisture rather than water_balance... there can be (and will be) many other variables in the water_balance dataset.
Also et, taw, sminit are no more arguments on cess_date since they are arguments of water_balance

onset_maproom/cess.py
```diff
@@ -20,37 +20,36 @@
...
             min_dry_days
-            - 1
-            - dry_spell.where(dry_spell[time_coord] == cess_delta).max(
+            #used this to subtract 1 from min_dry_days bc was having difficulty converting
+            #int/float value to timedelta.. perhaps not permanent but a workaround for now
+            #since the value will always be 1 anyways
+            - dry_spell.where(dry_spell[time_coord] == cess_delta).max( 
                 dim=time_coord
             )
         ).astype("timedelta64[D]")
```
remicousin, line 48: Try this instead then

np.timedelta64(min_dry_days - 1, "D")

remicousin (changes requested):

onset_maproom/cess.py
drewmresnick: Aha!

onset_maproom/calc.py
```diff
@@ -127,6 +130,34 @@ def onset_date(
     return onset_delta
 
 
+def cess_date(daily_rain, dry_thresh, min_dry_days, et, taw, sminit, time_coord="T"):
+    waterBalance = water_balance(daily_rain, et, taw, sminit, "T")
```
aaron-kaplan, line 134: `time_coord`, not `"T"`.

onset_maproom/cess.py
```diff
@@ -0,0 +1,117 @@
+#temporary file for cessation date functions; 
+#these will be moved to the main calc.py file when finished
```
aaron-kaplan, line 2: Remove this file now?

aaron-kaplan (commented):
Is there a meaningful common core that can be factored out of the onset date and cessation date calculations, or are they different enough that they should remain separate?

remicousin:
> Is there a meaningful common core that can be factored out of the onset date and cessation date calculations, or are they different enough that they should remain separate?

Everything that is in cess_date is also in onset_date. But given cess_date is currently 11 lines, I feel like it'd be an overkill to try and factor in.

onset_maproom/cess.py
drewmresnick: I was going to wait for Remi to look over the changes I made the the functions before putting them in the main calc.py file if that's okay. The functions in calc.py are placeholders just for getting the plots working

drewmresnick:
> Is there a meaningful common core that can be factored out of the onset date and cessation date calculations, or are they different enough that they should remain separate?

The calculations they are doing seem pretty separate overall especially given that the inputs for each function will be different (ie start day / month are different inputs so even though both functions do grouping on the data the grouping will be done using different subsets of the data). So I don't know how much could be factored out, unless I am misunderstanding your question. If you're asking just in terms of code that could be pulled out and made into a separate function to be used by both then perhaps this could be done.

aaron-kaplan:
@drewmresnick I didn't mean combine them into a single function that simultaneously calculates both onset date and cessation date. I meant that at a glance the two seem to have a lot in common, so I wondered if it made sense to pull out what they have in common, and make onset and cessation two smaller functions that both call the same common function.

I probably won't take the time to form an opinion on whether it makes sense or not, I just wanted the two of you to think about it.

drewmresnick:
@remicousin If you can look at the calculations for cess date to see if the steps are correct that would be good. So far it does not look like the output plot in the maproom is giving similar outputs to the original maproom.

Also note that I am trying to debug an issue where sometimes upon loading the maproom you get a reindexing error. It does not happen every time but if it does, just refresh the page and it should go away..

onset_maproom/maproom.py
```diff
@@ -105,74 +105,301 @@ def map_click(click_lat_lng):
...
+                showarrow=False,
+                yshift=10,
+                xshift=60,
+            )
+            alert1 = dbc.Alert(
+                "Cess alert: The dataset at the chosen coordinates is empty (NaN). Please choose a different point.",
+                color="danger",
+                dismissable=True,
+            )
+            return errorFig, errorFig, alert1
```
aaron-kaplan, line 291: Lines 277-291 are copy-pasted many times, in this function and the onset one. The only thing that changes is the error message. I would like to factor them out into a function, or maybe two functions (one for the figure and one for the alert). Although, do we really need the alert? Couldn't we just put the full error message in the figure and dispense with the alert?

drewmresnick: Okay. We can get rid of the error message and put it inside the figure, I made them separate so that the user would see it right away instead of only when they are on the correct tab. But it does not really matter.

remicousin:
I pushed some changes to the tests that seem to confirm that cess_date is working fine.

onset_maproom/calc.py
```diff
@@ -293,6 +308,63 @@ def seasonal_onset_date(
     # + seasonal_onset_date.onset_delta
     return seasonal_onset_date
 
+def seasonal_cess_date(
+    search_start_day,
+    search_start_month,
+    search_days,
+    daily_rain,
```
remicousin, line 315: arguments from here should be same as cess_date

onset_maproom/calc.py
```diff
@@ -293,6 +308,63 @@ def seasonal_onset_date(
...
+    )[0] + np.timedelta64(
+        search_days,
+        "D",
+    )
+
+    end_day = first_end_date.dt.day.values
+
+    end_month = first_end_date.dt.month.values
+
+    waterBalance = water_balance(daily_rain, et,taw,sminit,time_coord)
```
remicousin, line 341: that should not be here. soil_moisture is an input

remicousin (changes requested):

onset_maproom/maproom.py
drewmresnick: @aaron-kaplan Is this something you are going to do or should I work on doing this?

onset_maproom/calc.py
```diff
@@ -293,6 +308,59 @@ def seasonal_onset_date(
...
+    end_month = first_end_date.dt.month.values
+
+    # Apply daily grouping by season
+    grouped_daily_data = daily_tobegroupedby_season(
+        soil_moisture, search_start_day, search_start_month, end_day, end_month
+    )
+    # Apply onset_date
+
+    seasonal_data = (
+        grouped_daily_data["soil_moisture"]
```
remicousin, line 345: use seasonal_moisture.name maybe (see seasonal_onset...)? because you can't be sure that the name of the input argument will be "soil_moisture," correct?

onset_maproom/cess.py
remicousin: I've looked only at calc.py recently. Maybe remove that one if now irrelevant and to avoid confusion or duplicating review?

remicousin (changes requested):

onset_maproom/calc.py
```diff
@@ -293,6 +308,59 @@ def seasonal_onset_date(
...
 
+def seasonal_cess_date(
+    search_start_day,
+    search_start_month,
+    search_days,
+    soil_moisture,
+    dry_thresh,
+    min_dry_days,
+    time_coord="T"
+):
```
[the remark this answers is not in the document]
aaron-kaplan: Optional parameters have to come after required parameters. Other than that, it's a matter of taste, but I agree that settling on a standard pattern would help make the library easier to use. I like Rémi's order.

onset_maproom/maproom.py
aaron-kaplan: If you understand what I'm asking for, it would be great if you could do it. If not, it's probably easier for me to do it myself than to explain.

remicousin: I could give it a shot when I return to pyCPT Maproom but that will be only after I push my first PR on water balance. So I suggest that whoever starts working on this among the 3 of us alerts the other 2 so we don't duplicate work.

onset_maproom/calc.py
```diff
@@ -293,6 +308,59 @@ def seasonal_onset_date(
...
+            dry_thresh=dry_thresh,
+            min_dry_days=min_dry_days,
+        )
+        # This was not needed when applying sum
+        .drop_vars(time_coord)
+        .rename({"seasons_starts": time_coord})
+    )
+    # Get the seasons ends
+    seasons_ends = grouped_daily_data["seasons_ends"].rename({"group": time_coord})
+    seasonal_cess_date = xr.merge([seasonal_data.to_dataset(name='cess_delta'), seasons_ends])
```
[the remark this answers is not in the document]
drewmresnick: There was an issue with a naming convention. I just looked into it more and figured out a way to solve it without converting to a ds and renaming it there though which I will push.

onset_maproom/maproom.py
drewmresnick: My understanding is that you are wanting to create a function outside of the callbacks that does the exception handling/error printing, so you would just need to call the function on the data in each callback instead of copy pasting all of the code?

aaron-kaplan: That's the essence of it, yes. It might be a little tricky because "printing" the error in this case doesn't actually mean printing something to the terminal, it means displaying something in a web page.

remicousin:
please write test(s) for seasonal_cess_date function. You can find inspiration with seasonal_onset one(s)

onset_maproom/layout.py
```diff
@@ -183,6 +183,31 @@ def controls_layout():
                 agricultural campaign associated with it.
                 """
             ),
+            Block("Latitude / longitude inputs",
+                dbc.Row(
+                    [
+                        dbc.Col(
+                            dbc.Input(
+                                id = "latInput", 
+                                placeholder="Type latitude...",
```
remicousin, line 192: what about putting the allowed range here (3-15)?
And what about forcing out of range entries back in the range? ie if someone enters 2, we se to 3?
These are open questions and I think should go to another PR if we want to address them as there's already a lot in there.

onset_maproom/layout.py
```diff
@@ -164,97 +183,120 @@ def controls_layout():
...
+           #         id="probability-collapse",
+           #     ),
+           # ),
+            Block(
+                "Onset Date Search Period",
+                Sentence(
+                    "From Early Start date of",
+                    Date("search_start_", 1, "Jun"),
+                    "and within the next",
+                    Number("searchDays", 90, min=0, max=9999), "days",
```
remicousin, line 250: this one text box seems to be stuck to a smaller size that prevent proper reading of the number

remicousin:
I am about to approve as my last comments are either petty, or for a later PR.

If you want to test the results against what Ingrid does, remember that if using Maproom, the data is coarsened in space for performance reason. If you pick an Ingrid URL, you can check that that has been removed to have it done at the original resolution level.

More on layout: The controls sections seem to adapt to screen size even though there is likely improvement to make on that but that's another PR. Also the way it sets up now, the units of a entry number are often either far away or return to the next line. Doesn't seem ideal to me. Also another PR.

onset_maproom/layout.py
remicousin: This made me realize that there are no lat/lon ticking on our map. I wonder how much it matters now that we have the ability to both click and type-in...

remicousin (approved):

onset_maproom/layout.py
drewmresnick: As in, put the allowed range as a text within the input box instead of the 'type latitude...' ?

I can also look into seeing if we can auto-update the box as you say if you think that is better then just getting a red box if it is out of range. I suppose having the lat/lng tix boxes would be very nice in terms is visually choosing coordinates without having to click on the map and wait for the page to reload though.

remicousin: yes, when nothing is entered, show the acceptable range... That seems like a nice place to do it, and that would leave room for other info in the label below the box.
I didn't realize the red boxing when out of range. I remember seeing the red but having the impression it was always red... oh well... UX...

I was just thinking of having ticks which sounds appropriate for any figure in general to have ticks on the axes. Some sort of pop-up showing coordinates where your mouse is may be a useful feature too...

drewmresnick:
@remicousin should I merge this branch and address all of your comments in a new branch? Or just tidy up the smaller comments you added here (fixing the layout and moving the lat/lng range), and then addressing the bigger things ie adding tests and looking into adding tick marks in a new branch?

drewmresnick:
@remicousin @aaron-kaplan Should I merge this branch and open a new PR for bug fixes /adding tests?

aaron-kaplan:
It looks like Rémi already approved, so you can go ahead and merge if you're ready. It's not a consequential decision because nobody is using it yet.

PR #43, opened 2022-02-03
set pnep url to be blank when no forecast is available

kgraaf:
This solves the issue of erroneous forecast raster data being "left behind" when updated to a month for which there is no forecast

### DIFF
--- a/fbfmaproom/fbflayout.py
+++ b/fbfmaproom/fbflayout.py
@@ -358,6 +358,16 @@ def command_layout():
                     "vertical-align": "top",
                 },
             ),
+            dbc.Alert(
+                "No forecast available for this month",
+                color="danger",
+                dismissable=True,
+                is_open=False,
+                id="forecast_warning",
+                style={
+                    "margin-bottom": "8px",
+                },
+            ),
 
         ],
         id="command_panel",
--- a/fbfmaproom/fbflayout.py
+++ b/fbfmaproom/fbflayout.py
@@ -352,21 +352,13 @@ def command_layout():
                 ],
                 style={
                     "position": "relative",
-                    # "width": "1px", # force it to wrap
+                    "width": "1px", # force it to wrap
                     "display": "inline-block",
                     "padding": "10px",
                     "vertical-align": "top",
                 },
             ),
-            html.Div(id="command_msg",
-                     style={
-                         "position": "relative",
-                         "width": "110px",
-                         "display": "inline-block",
-                         "padding": "10px",
-                         "vertical-align": "top",
-                     },
-            ),
+
         ],
         id="command_panel",
         className="info",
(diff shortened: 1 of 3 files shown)


aaron-kaplan (changes requested):
If possible, it would be nice to display a message over the map saying something like "no forecast available for issue month {date}."

Also a minor stylistic comment below.

aaron-kaplan:
If possible, it would be nice to display a message over the map saying something like "no forecast available for issue month {date}."

fbfmaproom/fbfmaproom.py
```diff
@@ -843,9 +843,13 @@ def pnep_tile_url_callback(target_year, issue_month_idx, freq, pathname, season_
...
     target_month0 = season_config["target_month"]
 
-    # Prime the cache before the thundering horde of tile requests
-    select_pnep(country_key, issue_month0, target_month0, target_year, freq).data_array
-    return f"{TILE_PFX}/pnep/{{z}}/{{x}}/{{y}}/{country_key}/{season_id}/{target_year}/{issue_month_idx}/{freq}"
+    try:
+        # Prime the cache before the thundering horde of tile requests
+        select_pnep(country_key, issue_month0, target_month0, target_year, freq).data_array
+        return f"{TILE_PFX}/pnep/{{z}}/{{x}}/{{y}}/{country_key}/{season_id}/{target_year}/{issue_month_idx}/{freq}"
+    except:
```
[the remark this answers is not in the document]
kgraaf: Oh, that's a good point

aaron-kaplan:
I don't love having the message in the control bar. It's hard to notice. Would it be hard to overlay it on the map? That's where I'm looking when I'm looking for the missing forecast.

aaron-kaplan:
If you do manage to overlay the message on the map, try to configure it so that it only appears if the forecast layer is toggled on (visible).

kgraaf:
> I don't love having the message in the control bar. It's hard to notice. Would it be hard to overlay it on the map? That's where I'm looking when I'm looking for the missing forecast.

I can try and see

kgraaf:
Should the error message be placed in a tooltip like the "No Data" message?

aaron-kaplan:
The tooltip isn't visible on initial page load. If you want to make it appear immediately if there's no forecast available, that seems ok.

fbfmaproom/fbflayout.py
```diff
@@ -358,6 +358,16 @@ def command_layout():
...
+            dbc.Alert(
+                "No forecast available for this month",
+                color="danger",
+                dismissable=True,
+                is_open=False,
+                id="forecast_warning",
+                style={
+                    "margin-bottom": "8px",
+                },
+            ),
```
aaron-kaplan, line 370: Does this block interaction until you dismiss it? If so, that seems too heavyweight. Can't we add a label in the center of the map? Maybe you could use a [tooltip](https://leafletjs.com/reference.html#tooltip) with `permanent=true`?

aaron-kaplan: It doesn't block interaction.

aaron-kaplan (approved):

PR #44, opened 2022-02-03
handle condition where no data exists for current location

kgraaf:
Ensures that the table will not show erroneous data

### DIFF
--- a/fbfmaproom/fbflayout.py
+++ b/fbfmaproom/fbflayout.py
@@ -352,21 +352,13 @@ def command_layout():
                 ],
                 style={
                     "position": "relative",
-                    # "width": "1px", # force it to wrap
+                    "width": "1px", # force it to wrap
                     "display": "inline-block",
                     "padding": "10px",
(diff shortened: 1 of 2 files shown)


aaron-kaplan (changes requested):
Same comment as on the other PR: use `except Exception` instead of just `except`, and when there's no data add a message explaining why.

kgraaf:
If you want text to be overlayed on the map, there is already a tool tip that says "no data" when you drag the marker to a location without data. I think that would be redundant to also have an overlay

aaron-kaplan:
Oh yeah, so there is. Please remove the message then, and put back the `width: 1px`, which is there for a reason.

kgraaf:
This should be ready to go

aaron-kaplan (approved):
Looks good. We need to bump the version number in __about__.py before releasing. If the other PR is about to get merged then you can merge this as-is, bump the version in the other one, and then release.

PR #46, opened 2022-02-04
Another bug in trigger check endpoint

aaron-kaplan:
Also removing some dead code, and fixing a test that was broken by the previous PR. (We need CI so that we don't forget to run the tests before merging.)


jturmelle (approved):
Looks good.

aaron-kaplan:
@kgraaf waiting for your review.

kgraaf (approved):

PR #48, opened 2022-02-09
Simplify tiling

aaron-kaplan:
Before we invite others to start generating tiles, I wanted to simplify the tiling API, with two goals: make it easier to use, and encapsulate the use of OpenCV so that we're not tied to it permanently.

~I still want to try to simplify the handling of colormaps, but I couldn't get that done today. I'm submitting what I have so we can merge it quickly and minimize divergence between branches.~

This PR is broken up into logically distinct commits. I recommend reading one commit at a time.

### DIFF
--- a/pingrid.py
+++ b/pingrid.py
@@ -54,7 +54,6 @@ FuncInterp2d = Callable[[Iterable[np.ndarray]], np.ndarray]
 class DataArrayEntry(NamedTuple):
     name: str
     data_array: xr.DataArray
-    interp: Optional[FuncInterp2d]
     min_val: Optional[float]
     max_val: Optional[float]
     colormap: Optional[np.ndarray]
@@ -229,7 +228,7 @@ def produce_bkg_tile(
 
 
 def produce_data_tile(
-    interp: FuncInterp2d,
+    da: xr.DataArray,
     tx: int,
     ty: int,
     tz: int,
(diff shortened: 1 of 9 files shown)


aaron-kaplan:
I pushed more commits. Among other things, I have eliminated the DataArrayEntry class. Now we store the colormap, min, and max as xarray attributes.

aaron-kaplan:
This is exciting... a few commits back I delayed a `transpose` operation that I noticed was causing the entire dataset to be loaded even when we only needed one tile's worth. Now that that's fixed, everything is fast enough that we can get rid of the `@lru_cache` decorators, which were going to cause scaling issues.

pingrid.py
```diff
@@ -121,14 +121,14 @@ def interp_func(output_grids: Iterable[np.ndarray]) -> np.ndarray:
...
-    y = da[dims[0]].values
+def create_interp(da: xr.DataArray) -> FuncInterp2d:
+    x = da["lon"].values
+    y = da["lat"].values
     input_grids = [
         (y[0], y[1] - y[0]),
         (x[0], x[1] - x[0]),
     ]  # require at least 2 points in each spatial dimension, and assuming that the grid is even
-    input_data = da.values
+    input_data = da.transpose("lat", "lon").values
```
kgraaf, line 124: determine whether this line is indeed a no-op (it seems like the lat/lon dimensions should already be in the right order wrt one another and thus this line should do nothing). Also consider reordering dimensions in original file to make XY slices continugous/put T dimension before XY.

kgraaf (approved):

PR #56, opened 2022-02-18
Generate raster tiles with Mercantile and Pillow

kgraaf:
This PR generates raster tiles with Mercantile and Pillow instead of Pingrid. Right now the images are in black and white, I'll need to add the colormap.

Working sample URL http://127.0.0.1:8050/onset-maproom-tiles/onset/7/78/60/1/5/20/3/7/21 (adjust host name and port accordingly of course)

### DIFF
--- a/onset_maproom/calc.py
+++ b/onset_maproom/calc.py
@@ -121,7 +121,6 @@ def onset_date(
                 dim=time_coord
             )
         ).astype("timedelta64[D]")
-        # ).astype(int)
         # delta from 1st day of time series
         - daily_rain[time_coord][0]
     ).rename("onset_delta")
--- a/onset_maproom/calc.py
+++ b/onset_maproom/calc.py
@@ -121,6 +121,7 @@ def onset_date(
                 dim=time_coord
             )
         ).astype("timedelta64[D]")
+        # ).astype(int)
         # delta from 1st day of time series
         - daily_rain[time_coord][0]
     ).rename("onset_delta")
@@ -330,3 +331,8 @@ def probExceed(onsetMD, search_start):
     cumsum["Days"] = onset_unique
     cumsum["probExceed"] = 1 - cumsum.onset / cumsum.onset[-1]
     return cumsum
+
+
+def days(da):
+    """convert time delta into number of days"""
+    return da / np.timedelta64(1, 'D')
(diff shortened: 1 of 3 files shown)


kgraaf:
Currently this is very slow (approximately 30 sec per tile). However, I have not tested it with rechunked data yet.

aaron-kaplan (commented):
[Edit: some of this is wrong. See later comment.]

A square tile in Mercator space corresponds to a non-rectangular region in the equirectangular projection. Instead of finding that non-rectangular region, you're finding its rectangular bounding box, fetching the data for that bounding box, and then calling `resize` on it to squash it into a square. That might look ok superficially, in the sense that pixels appear on the part of the map where we expect to see pixels, but it's not the right pixels.

Since Ethiopia is close to the equator, where the Mercator projection isn't very different from the equirectangular, it might actually be good enough, but if you tried it on Norway I don't think it would be close.

Rather than mapping the entire tile to equirectangular space, pingrid.py:produce_data_tile maps each pixel of the Mercator tile to the centroid of its projection into equirectangular space, and then interpolates to get values at those points.

A week ago I'm not sure I would have understood that what you're doing is wrong, or at least it would have taken me a lot longer. What changed since then is that I spent hours this week reading pingrid.py and cross-checking with the wikipedia page and some other resources on the Mercator projection. It was very educational.

BTW I'm doing a similar experiment on the fbfmaproom but with matplotlib, which has map projection functions, instead of pillow. fbfmaproom is a better place for this kind of experiment because there's no expensive calculation involved, it's just loading data from disk and displaying it.

aaron-kaplan:
I think this shows that there are good prospects for replacing much of pingrid with existing libraries, but that's not what we're working on right now. Since we managed to encapsulate all the cv2 stuff, we can go ahead and use it for now without tying ourselves to it forever. There are other demands that are more pressing.

aaron-kaplan:
Part of my earlier comment was wrong. A square in Mercator space projects to a rectangle in equirectangular space. The problem is just that `resize` will stretch all the pixels equally, when in fact the projection back into Mercator should stretch pixels at the north end more than those at the south end. (Right?)

remicousin:
I would need to do some research on the 2 projections and resize to try and answer that question, is that is what you seek. let me know

aaron-kaplan:
@remicousin don't put any time into it, at least for now. I think our preferred solution is to switch to the equirectangular projection, like Ingrid, which would make this point moot. It remains to be seen how difficult that will be.

PR #57, opened 2022-02-18
Python dependency updates

kgraaf:
Updates onset maproom to work correctly with newer version of the bootstrap components. Also fixes deprecation warnings.

- Downgraded environment to Python 3.9 to fix rasterio error
- Upgraded environment to v1 of Dash Bootstrap Components
- Fixed deprecation warnings by switching to importing from dash instead of dedicated packages (should have identical behavior)
- Switched `no_gutters` parameter to using bootstrap `g-0` class
- Switched deprecated `bs_size` parameter to `size`
- Removed `FormGroup` component and code designed to support it
- Added padding that was eliminated by removal of `FormGroup`s

### DIFF
--- a/onset_maproom/environment_linux.yml
+++ b/onset_maproom/environment_linux.yml
@@ -11,22 +11,28 @@ dependencies:
   - asciitree=0.3.3=py_2
   - attrs=21.4.0=pyhd8ed1ab_0
   - blosc=1.21.0=h9c3ff4c_0
+  - bokeh=2.4.2=py39hf3d152e_0
   - boost-cpp=1.74.0=h6cacc03_7
-  - brotli-python=1.0.9=py310h122e73d_6
+  - brotli-python=1.0.9=py39he80948d_6
   - bzip2=1.0.8=h7f98852_4
   - c-ares=1.18.1=h7f98852_0
   - ca-certificates=2021.10.8=ha878542_0
   - cairo=1.16.0=ha00ac49_1009
+  - certifi=2021.10.8=py39hf3d152e_1
   - cfitsio=4.0.0=h9a35b8e_0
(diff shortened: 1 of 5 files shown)


onset_maproom/environment_linux.yml
```diff
@@ -1,4 +1,4 @@
-name: enactsmaproom
+name: onset
```
aaron-kaplan, line 1: I would like to continue to call this enactsmaproom. There are eventually dozens of ENACTS maprooms, and I don't want them each to have their own conda environment.

aaron-kaplan:
When you send a PR that isn't ready to merge, it would be helpful if you explain what state it's in and what is left to do. When we spoke earlier today, you mentioned that there was something that's still not working. Has that been resolved since? If not, what's the problem? You also mentioned that you can't build the environment from environment_linux.yaml. I just tried and got an error on rasterio. Is that what you ran into?

aaron-kaplan:
I grepped the error that came up when compiling rasterio and found https://github.com/rasterio/rasterio/issues/2333
Sounds like we should back off to python 3.9 until this gets sorted out.

aaron-kaplan:
> that aside this repo is (or was) ready to merge

OK, try recreating the environment with python 3.9 and let me know if it solves the problem.

kgraaf:
I'm doing that right now

kgraaf:
I added a précis of these changes to the original comment. It should be ready to merge but you are welcome to independently test if so desired.

aaron-kaplan (approved):

aaron-kaplan:
@remicousin @drewmresnick the conda environment has been updated. Please rebase any open branches and rebuild your environment.

PR #58, opened 2022-02-22
Updated fbf packages

kgraaf:
removed offending code that prevented bootstrap components upgraded and pinned necessary versions of pandas and shapely

Two packages had to be pinned to specific (old versions) so that the environment could be safely added to:

1. Pandas:

    Newer versions of Pandas do not appear to work correctly with queuepool. In particular, how that package forms SQL query objects needs to be updated. Hopefully the solution here is to either send Igor a issue or to migrate to a different connection pooling library    
2. Shapely:

    Shapely changed how looping over geometries works in the newer version and Pingrid relies on the old functionality. It is unclear to me whether using the .geoms property has the same behavior as the "built in" iterator the old version had. As we may migrate way from Pingrid and we're not going to upgrade it right now and possibly waste the work or effort.

The dash bootstrap components in v1 dropped support for hide_arrow and innerClassName. I suspect the reason why the former was used is that trying to consistently style the tooltip arrows is extremely difficult and I couldn't figure out. You can get the upwards pointing arrow to work but now the other directions. The latter removal is fairly easy to remedy with CSS selectors.  However, the display is still subtly different.

Before:

tooltip-old

After:

tooltip-new

(Note that the lack of the logos is not related to this change, I just don't have that set up in my dev environment, but it isn't a regression)
 

As a brief aside, the tooltips that hover over the table are not actually bootstrap tooltips, they're built in functionality of dash_table with an entirely different structure. As such, trying to get style consistency between the two is extremely difficult and fraught. I tried but couldn't succeed. Also, while I thought that dash_table allowed you to set a class for them this turns out to not be true after all, so even styling them is difficult (and fragile, as I would guess because I don't see any indication that the specific CSS classes they happen to use are necessarily intended to be stable across versions of the library)

### DIFF
--- a/fbfmaproom/assets/style.css
+++ b/fbfmaproom/assets/style.css
@@ -42,11 +42,18 @@ tr {
     background-color: #ff6961;
 }
 
-.tooltiptext {
-  color: black;
-  position: relative;
-  background-color: rgba(228,228,228, 1);
-  max-width: 300px;
-  min-width: 300px;
-  font-size: 110%;
+.tooltiptext { 
+    opacity: 1 !important;
+}
+
+.tooltiptext .tooltip-inner {
+    color: black;
+    background-color: #E4E4E4;
+    max-width: 300px;
+    min-width: 300px;
+    font-size: 110%;
+}
+
+.tooltiptext .tooltip-arrow {
+    display: none;
 }
--- a/fbfmaproom/environment.yml
+++ b/fbfmaproom/environment.yml
@@ -3,15 +3,16 @@ channels:
   - conda-forge
   - defaults
 dependencies:
+  - python=3.9
   - cftime
   - dash
   - flask
   - numpy
   - opencv
-  - pandas
+  - pandas=1.3
   - psycopg2
   - rasterio
-  - shapely
+  - shapely=1.7
   - xarray
   - yaml
   - zarr
--- a/fbfmaproom/environment.yml
+++ b/fbfmaproom/environment.yml
@@ -3,15 +3,28 @@ channels:
   - conda-forge
   - defaults
 dependencies:
+
+    # Conda still appears to have issues surrounding Python 3.10
   - python=3.9
   - cftime
   - dash
   - flask
   - numpy
   - opencv
+
+    # Newer versions of Pandas do not appear to work correctly with queuepool.
+    # In particular, how that package forms SQL query objects needs to be updated. 
+    # Hopefully the solution here is to either send Igor a issue or to migrate to
+    # a different connection pooling library
   - pandas=1.3
   - psycopg2
   - rasterio
+
+    # Shapely changed how looping over geometries works in the newer version and
+    # Pingrid relies on the old functionality. It is unclear to me whether using
+    # the .geoms property has the same behavior as the "built in" iterator the
+    # old version had. As we may migrate way from Pingrid and we're not going to
+    # upgrade it right now and possibly waste the work or effort.
   - shapely=1.7
   - xarray
   - yaml
--- a/fbfmaproom/environment.yml
+++ b/fbfmaproom/environment.yml
@@ -4,7 +4,8 @@ channels:
   - defaults
 dependencies:
 
-    # Conda still appears to have issues surrounding Python 3.10
+    # rasterio build fails with python 3.10
+    # https://github.com/rasterio/rasterio/issues/2333
   - python=3.9
   - cftime
   - dash
--- a/fbfmaproom/environment_linux.yml
+++ b/fbfmaproom/environment_linux.yml
@@ -7,157 +7,171 @@ dependencies:
   - _openmp_mutex=4.5=1_gnu
   - affine=2.3.0=py_0
   - alsa-lib=1.2.3=h516909a_0
+  - aom=3.2.0=h9c3ff4c_2
   - asciitree=0.3.3=py_2
-  - attrs=21.2.0=pyhd8ed1ab_0
+  - attrs=21.4.0=pyhd8ed1ab_0
   - boost-cpp=1.74.0=h312852a_4
-  - brotli-python=1.0.9=py39he80948d_5
+  - brotli-python=1.0.9=py39he80948d_6
   - bzip2=1.0.8=h7f98852_4
-  - c-ares=1.17.1=h7f98852_1
-  - ca-certificates=2021.10.26=h06a4308_2
+  - c-ares=1.18.1=h7f98852_0
+  - ca-certificates=2021.10.8=ha878542_0
   - cairo=1.16.0=h6cf1ce9_1008
-  - certifi=2021.10.8=py39h06a4308_0
+  - certifi=2021.10.8=py39hf3d152e_1
   - cfitsio=3.470=hb418390_7
-  - cftime=1.5.0=py39h6323ea4_0
-  - click=7.1.2=pyh9f0ad1d_0
+  - cftime=1.5.2=py39hce5d2b2_0
+  - click=8.0.4=py39hf3d152e_0
   - click-plugins=1.1.1=py_0
-  - cligj=0.7.2=pyhd8ed1ab_0
-  - curl=7.77.0=hea6ffbf_0
-  - dash=1.20.0=pyhd8ed1ab_0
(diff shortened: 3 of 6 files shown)


aaron-kaplan:
- [ ] Need to consolidate metadata in all the zarrs before upgrading xarray.

kgraaf:
- [ ] make control bar and table tooltips consistent in styling
- [ ] add notes on packages that had to be pinned

aaron-kaplan:
After doing some reading, I think the disadvantages of consolidated metadata outweigh the benefits for us. The benefit is supposed to be that it makes `open_zarr` faster, but for the datasets we're using `open_zarr` is already essentially instantaneous. The disadvantage is the potential for inconsistency. The .zmetadata file is a centralized collection of duplicate copies of the metadata from the individual variables that make up the dataset. In a setup where the zarr store will be updated periodically, there is a risk that we update metadata in a variable but fail to update the consolidated metadata, either because we forget, or because the program crashes between steps. This is yet another potential cause of the frustrating "why isn't the new forecast appearing" problem, and I would rather avoid it.

@kgraaf please add `consolidated=False` to all calls to `open_zarr`, so that it will continue to work with non-consolidated zarr stores as it did with xarray 0.19.

kgraaf:
@aaron-kaplan this should be ready to go

aaron-kaplan:
Whenever we change something that affects the appearance of the application, let's get in the habit of attaching a screenshot to the PR. Although in this case it might be tricky to get a screenshot of a tooltip?

kgraaf:
Just added screenshots to the original comment

fbfmaproom/assets/style.css
```diff
@@ -42,11 +42,18 @@ tr {
...
+    opacity: 1 !important;
+}
+
+.tooltiptext .tooltip-inner {
+    color: black;
+    background-color: #E4E4E4;
+    max-width: 300px;
+    min-width: 300px;
+    font-size: 110%;
+}
```
aaron-kaplan, line 55: How about we use the same style for both `tooltip-inner` and `dash-table-tooltip`?

kgraaf: I can try. It won't really be perfectly consistent because like I said they don't use the same structure. For example, the dash_table tooltips have a border and an entirely different shape.

aaron-kaplan: I see your comment in the edited PR description about not knowing if the DataTable CSS classes are intended to be part of the stable API. It's a good point, but I still think we should try to keep the two consistent. When we get around to writing front end tests, maybe we could write a test that verifies that the table tooltips have the expected classes?

aaron-kaplan: Don't spend all day on it. I'm just curious what happens if you change the selector on this style from `.tooltip-inner` to `.tooltip-inner, .dash-table-tooltip`. Then we'll move on, whether that works or not.

aaron-kaplan (approved):

fbfmaproom/assets/style.css
kgraaf: Those are all good ideas. I'll try it in just a bit, I'm just going to push the table demo because the only thing it's "missing" is the updated environment and it'll be easy enough to rebase in terms of that

aaron-kaplan:
If you haven't already, please test on the data in /data/aaron/fbf, to verify that `consolidated=False` does in fact suppress the consolidated metadata warning.

kgraaf:
> If you haven't already, please test on the data in /data/aaron/fbf, to verify that `consolidated=False` does in fact suppress the consolidated metadata warning.

I believe I already did but I will again just to be sure

fbfmaproom/assets/style.css
kgraaf: Sorry, forgot to add this friday: So, the CSS class is `dash-tooltip` not `dash-table-tooltip` but in any case it doesn't work. Should we just call this done then? I tried a couple of other things too Friday afternoon and this morning and neither worked. At the very least in terms of consistency the table is no worse than it was before.

aaron-kaplan: Yes, let's move on. Out of curiosity, do you know why what you tried didn't work?

FWIW, I inspected the DOM in the browser and saw an element with the class `dash-table-tooltip`.

aaron-kaplan (changes requested):
There was a todo item "add notes on packages that had to be pinned." You checked that item off, but I can't find the notes. Did you forget to push them?

kgraaf:
> There was a todo item "add notes on packages that had to be pinned." You checked that item off, but I can't find the notes. Did you forget to push them?

I added them to the first comment on this PR

fbfmaproom/assets/style.css
kgraaf: I'm wondering what you were looking at then, I'm not entirely sure. I had to use the debugger to "freeze" the DOM to even inspect the tooltip. I did a lot of research on how to style these tooltips and the answers are conflicting and none worked. I think the bootstrap tooltip CSS just don't really work with the dash table ones because the structure is very different.

aaron-kaplan:
> I added them to the first comment on this PR

Ah, right. Please summarize them as comments in environment.yml where the pinning happens.

kgraaf:
Ok cool. Will do

kgraaf:
Wait, I got this wrong. It's a bug with rasterio not Conda. Also, incidentally, GH is saying you requested changes but I can't even see what this is referring to

aaron-kaplan:
> GH is saying you requested changes but I can't even see what this is referring to

That's attached to my comment of 3:17pm asking where the explanation of the pinned packages was.

aaron-kaplan (approved):

kgraaf:
Oh ok cool. The UI was just a little confusing

aaron-kaplan:
Remember to run the tests before you merge. `conda install pytest` and then `python -m pytest`, in the fbfmaproom directory and conda environment.

kgraaf:
OK Thanks for the reminder!

kgraaf:
Ok tests ran but I did have to prefix your command with `CONFIG=fbfmaproom-sample.yaml`. I assume that's right. I also updated the package import to not give the deprecation warnings.

aaron-kaplan:
> I did have to prefix your command with `CONFIG=fbfmaproom-sample.yaml`. 

Yes, right.

Ordinarily I like to deploy changes as soon as they're merged, but I think this might be a particularly sensitive time for the project, so I will check with Dan before we deploy.

aaron-kaplan:
Oops, forgot to update environment_macos.yaml. I have an old macbook I got from Jeff for things like this, will take care of it in a follow-up PR.

kgraaf:
OK thank you! I've been wondering how we should do that since I have no means to

fbfmaproom/environment.yml
```diff
@@ -3,15 +3,30 @@ channels:
...
   - dash
   - flask
   - numpy
   - opencv
-  - pandas
+
+    # Newer versions of Pandas do not appear to work correctly with queuepool.
+    # In particular, how that package forms SQL query objects needs to be updated. 
+    # Hopefully the solution here is to either send Igor a issue or to migrate to
+    # a different connection pooling library
```
aaron-kaplan, line 19: More precisely, new versions of `pandas.read_sql` and `pandas.read_sql_query` no longer accept a DBAPI object as the connection. Apparently it worked by coincidence, not by design. A change made last July [made it an error](https://github.com/pandas-dev/pandas/pull/42546/files#diff-9268174bfb15f08ef2267375665a85fecf201999902542f6fc9c0d3fadfb4553R742), but people [complained](https://github.com/pandas-dev/pandas/issues/45416), so they [changed it to a warning](https://github.com/pandas-dev/pandas/pull/45496/files#diff-9268174bfb15f08ef2267375665a85fecf201999902542f6fc9c0d3fadfb4553R747).

The solution appears to be to use psycopg2 via SQLAlchemy rather than directly. I think that's a good move regardless; it's the community standard.

aaron-kaplan: And once we have SQLAlchemy as a dependency, there's no cost to also using its QueuePool instead of queuepool, which should solve the problem of applications hanging when the database server restarts.

PR #61, opened 2022-03-02
combine summary and table table into one

kgraaf:
This is not complete yet as the styling is not correct but I wanted to just show what I have so far.


fbfmaproom/fbfmaproom.py
```diff
@@ -310,6 +310,25 @@ def generate_tables(
     # that ever actually happen?
     return main_presentation_df, summary_presentation_df, prob_thresh
 
+def merge_tables(summary, table):
```
kgraaf, line 313: the way I wrote this is probably pretty pedantic but I wanted to remain agnostic as to which columns need to be filled in. Note that the pandas `concat()` method will *not* work if the set of columns are not identical between every dataframe being concatenated.

aaron-kaplan: inner means keep only the columns that exist in both; outer means keep all columns that exist in either. So I would think that an outer join was what you want. In any case, there's no harm in being explicit.

aaron-kaplan:
Looks good. Thanks for sharing.

fbfmaproom/fbfmaproom.py
kgraaf: Are you sure? Outer is the default per the instructions but when I don't modify the two tables to have the same sets of columns it throws and exception

aaron-kaplan: What's the exact error message you get when you try an outer join without adding the missing columns first? It seems to work in this simple case:

```
>>> df1 = pd.DataFrame({'foo': [1, 2], 'bar': [3, 6]})
>>> df2 = pd.DataFrame({'foo': [4, 8, 12]})
>>> pd.concat([df1, df2])
   foo  bar
0    1  3.0
1    2  6.0
0    4  NaN
1    8  NaN
2   12  NaN
```

fbfmaproom/fbflayout.py
```diff
@@ -403,87 +403,28 @@ def table_layout():
             dcc.Loading(
                 [
                     table.DataTable(
-                        id="summary",
+                        id="table",
                         page_action="none",
                         style_table={
-                            "height": "auto",
-                            "overflowY": "scroll",
```
kgraaf, line 410: The overflowY property ends up being unnecessary (whether because of fixed rows or just period) and if added introduces undesirable visual artifacts

fbfmaproom/fbflayout.py
```diff
@@ -403,87 +403,28 @@ def table_layout():
...
+                            "height": "900px",
```
kgraaf, line 409: The height of the table needs to be increased since it encompasses both the header and the table proper

aaron-kaplan: Yeah, I remember looking into that and not finding a solution. This is one of the things we planned to address in switching to a new layout.

kgraaf:
This should be mostly ready to go save for one issue I cannot figure out: for some reason these changes (or perhaps others? I'm not sure what version is deployed) removes some padding that looks bad in my opionion. I can't figure out what would be causing that though. 

before:



after:

fbfmaproom/fbflayout.py
```diff
@@ -403,32 +403,33 @@ def table_layout():
...
                             },
                             {
-                                "selector": "tr:not(last-child)",
-                                "rule": "font-weight: bold; background-color: rgb(255, 255, 255); color: rgb(251, 101, 57);",
-                            },
-                            {
-                                "selector": "tr:last-child",
-                                "rule": "font-weight: bold; background-color: rgb(241, 241, 241); color: black",
+                                "selector": "div.dash-fixed-row tr:nth-child(7)",
+                                "rule": "border-bottom: 2px solid rgb(150, 150, 150)",
```
kgraaf, line 420: change to last-child

fbfmaproom/fbflayout.py
```diff
@@ -403,32 +403,33 @@ def table_layout():
...
                             },
                         ],
                         style_cell={
                             "whiteSpace": "normal",
                             "height": "auto",
                             "textAlign": "center",
                             "border": "1px solid rgb(150, 150, 150)",
-                            "background-color": "rgba(255, 255, 255, 0)",
+                            "backgroundColor": "rgb(248, 248, 248)",
```
kgraaf, line 428: mistake, remove

fbfmaproom/fbflayout.py
kgraaf: Take a look at dash-spreadsheet dash-freeze-top

fbfmaproom/fbflayout.py
```diff
@@ -403,32 +403,33 @@ def table_layout():
...
                         page_action="none",
                         style_table={
-                            "height": "auto",
-                            "overflowY": "scroll",
+                            "height": "900px",
                         },
                         css=[
                             {
+                                # attempting to be more specific introduces a bizarre visual artifact
+                                # "selector": "div.dash-fixed-row tr:first-child",
```
aaron-kaplan, line 414: I tried uncommenting this line and commenting the one below, and I don't see any difference. Could you describe/screenshot what you're seeing?

It did put my browser tab into an infinite loop at one point, when I was resizing the window. I don't know if that's related to the problem you saw or not.

aaron-kaplan: OK, I tried it in firefox and the weird behavior is more pronounced.

The reason `div.dash-fixed-row` doesn't match is clear to me--no parent of the row in question has that class. The div that matches that pattern is the summary table (the fixed rows), whereas you're trying to hide the first non-fixed row.

fbfmaproom/fbflayout.py
kgraaf: This is actually correct. I reworked the CSS logic slightly. This is a product of the summary and the data tables being combined, where the gray is the default and the header rows have different background colors

fbfmaproom/fbfmaproom.py
kgraaf: I'm not sure why this behavior doesn't work correctly for combining the summary and data tables. I think we should just keep it as is because the code to work around is fairly minimal.

fbfmaproom/fbflayout.py
kgraaf: No that's incorrect. The first row of the fixed portion is the "built-in" header. DataTable prescribes that the first fixed row be the header which is formed from (in this case) the pandas column labels. There's no way to override this behavior, and we want the header to be the *last* row of the fixed portion. As such we hide the "true" header and insert our own. 

The problem is that if the first row of the fixed portion (that is to say the aforementioned "true" header) *and* the first row of the non-fixed/scrolling portion are not hidden, we get the weird visual artifacts I described.

aaron-kaplan: I see. I thought you were complaining that you tried a more specific selector and it didn't match for some reason that you couldn't figure out.

Now that I understand the situation, I think the comment could be clearer. e.g. "Hide the first fixed row and the first data row to work around glitchy browser behavior." And link to a github issue or forum discussion if you have one.

kgraaf: When this line is uncommented and the one below it is commented. Only the "true" header is hidden. That would be desirable and would make duplicating the first row of the data portion unnecessary. Except it produces this result: 


The left border is missing and, while I can't show this in a static image, while scrolling the data portion, the header part will become unaligned by a single pixel.

kgraaf: Ok, I'll update the comment. I couldn't find any forum or github discussion of this unfortunately.

fbfmaproom/fbflayout.py
kgraaf: [Here's a discussion of this issue.](https://community.plotly.com/t/setting-datatable-max-height-when-using-fixed-headers/26417) It appears to be a known problem.

fbfmaproom/fbfmaproom.py
aaron-kaplan: I just commented out lines 315-325 and it worked fine for Ethiopia, all three issue months. Can you share a configuration in which the bug manifests itself?

fbfmaproom/fbflayout.py
kgraaf: I was able to increase the size by simply setting the height and max height to the same value.

fbfmaproom/fbfmaproom.py
kgraaf: Huh. I just did the same and it worked fine for me too. Maybe it was a bug in a previous version of pandas? (Recall when I first was developing this it was with the environment that had the issues). I guess I'll remove the lines then since I can confirm they definitely aren't needed.

kgraaf: Alright, I deleted those lines

kgraaf:
This PR should be good to go

fbfmaproom/fbfmaproom.py
aaron-kaplan: That seems plausible (pandas bug in previous version).

aaron-kaplan (approved):

aaron-kaplan:
The version that's in production is still using the old version of
bootstrap. I've been holding off on deploying any new versions because
there were important workshops planned and I didn't want to disturb things.
This is just a guess, but the dash-bootstrap upgrade seems a likely reason
for this kind of change.

PR #62, opened 2022-03-08
Revamp exceptions raised for client-side errors; log more errors

aaron-kaplan:
When we fixed the problem of stale data appearing in the table or the map, we suppressed some potential errors that should be logged. Raising a more specific exception in the cases where we actually intended to suppress the message, so we can give them special treatment.

### DIFF
--- a/pingrid.py
+++ b/pingrid.py
@@ -655,32 +655,47 @@ def ring_shapely_to_leaflet(ring):
 # Flask utils
 
 
-class InvalidRequest(Exception):
-    def __init__(self, message):
-        super().__init__()
+class ClientSideError(Exception):
+    def __init__(self, message, status):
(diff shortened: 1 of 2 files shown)


pingrid.py
```diff
@@ -655,32 +655,47 @@ def ring_shapely_to_leaflet(ring):
 # Flask utils
 
 
-class InvalidRequest(Exception):
-    def __init__(self, message):
-        super().__init__()
```
aaron-kaplan, line 660: Because we didn't pass the message along to the superclass constructor, that message didn't get logged along with the stack trace.

aaron-kaplan:
Merging prior to review because Kevin is on vacation. @kgraaf please look it over when you get back.

PR #63, opened 2022-03-11
Remic/watbal

remicousin:
This is a first PR that sets up a suite of functions for soil plant water balance, and its series of tests. The goal is to reproduce Ingrid's ([code](https://bitbucket.org/iridl/ingrid/src/master/src/water_balance.tex), [doc](http://iridl.ldeo.columbia.edu/dochelp/Documentation/details/index.html?func=:Water_Balance)). Ingrid's is monolithic whereas this is broken down into multiple functions, with the ambition that the suite could grow with other ways to compute same parts, as water balance can become a lot more sophisticated. Some alternatives were already implemented:

1. 2 ways to compute runoff
2. now planting date (thus Kc thus ET crop) can be inferred within the algorithm

The master function is soil_plant_water_balance. It probably aligns with the earlier water_balance function but I felt safer to let water_balance alone given the amount of changed, or rather additions. Right now it is thought for doing one season at a time (ie only one planting_date in time (see #8 below and After this PR section for more)).
Eunjin: I don't know your degree of familiarity with Python and Xarray, so please review as you see fit. It would be good though that you review the big picture; whether some things should become functions because they could be replaced by something else; in general whether I've not closed doors to increased sophistication, the way it is coded now. I also know these things are hard to predict but if there are obvious cases, let know.

Some things that may requite attention in this PR, or later:

1. proper units management (for functions, but also for a bunch of constants used hard-coded in functions, or as arguments)
2. Eunjin: I see more can be done with respect to the solar radiation, but I understand that all that belongs to the Penman case, correct? 
3. Eunjin: I see you have a case where rho depends on Crop, and ET crop, but in your code it seemed very much hard-coded and specific so I didn't try to generalize that. I didn't look at the FAO reference about it. Is it where I would find the general case?
4. in weekly_api_runoff, I use dropna after the rolling to get rid of the first days that become NaN because of rolling. I don't have a sense whether I should do that here inside the function, or leave it and let other functions/users choose to dropna or not. It's necessary though somewhere in the overall algo since we are having a recurrence relationship, if we start with NaNs, we'll get NaNs all the way through
5. I notices that empty and empty_like actually initialized with values so I used full_like with fill value of np.nan instead in the creation of the variables that need be defined before being filled in by the loop. FYI
6. Eunjin: Walter didn't get back to us really so the API categories thresholds and polynomial coefficients are hard-coded. They are default arguments the runoff functions though, so at least can be input with something else
7. After I created all that could be created before the loop, and before I start loop business, I broadcast all variables in water_balance dataset. This is because I didn't get broadcasting to automatically work otherwise in the loop business. It's possible that my broadcasting is too... broad and it could be done more as really needed according to options being triggered. I don't have a sense how much it matters, efficiently wise...
8. Planting Date is of same nature as onset/cessation_date which is that it has a T grid that from which the variable expressed in timedeltas refers to. Right now there is only one planting date (see more below in After this PR section). I realized when I added planting date variable (p_d) to the water_balance dataset, that I had to change its dim name not to be T, otherwise it would be meaninglessly extended over the other variables T. So that makes me think that there is maybe a flaw in the mental conception on how those _date variables should be... to be discussed
9. et_crop is built/saved inside the loop, whereas et_crop_red and drain, outside the loop. It could be either way for the 3 variables. Any sense if a way should be better?
10. There is a lot of engineering around the T dim: sel, expand, squeeze... that seemed necessary to deal with a loop that involves i and i-1 points (so need to squeeze), and because all my sub-functions have T as argument (so need to expand), so that they can run independently vectorially against T as well as point-wise in a loop. 
11. I tested my runoff function with Eunjin's, but that's about all I did for individual functions. It might be possible though to test comparable cases between soil_plant_water_balance and Eunjin's entire code as well. Should try that?

After this PR:

The main next step is to run the water balance over multiple seasons. Originally, I thought I could do that using the daily_groupby facility but now I realize that likely won't work because the logic of groupby is to then reduce the groups to a value each; whereas here, we want to keep the daily resolution, we just need not to run the water balance on all T points... only for those where we have a Crop planted (and slightly before for spinning-off simulation, slightly after possibly).

Now I think this should be managed via planting date, which will be tricky as it can be an input or estimated during the run. Then we might not be able to escape running the simulation for all T points, or maybe we can but not desirable to drop such points. I am not sure anymore. This is all about performance since a growing season is typically less than half a year, depending on crops.

Then after that major change, I can see if there is time to address things not addressed by this PR. And then run the wat bal for Senegal historically and make a Historical wat bal Maproom in Python.


remicousin:
Who else would it make sense to add to this PR?
Also can't attach Eunjin's .py. Shall I just commit it?

aaron-kaplan:
> Who else would it make sense to add to this PR?

Eunjin. I know you already initiated that, just reminding you.

> Also can't attach Eunjin's .py. Shall I just commit it?

Hmm, I guess so, but I don't want to cause confusion about whether it's used in the application or not. How about we put it in a subdirectory called `doc` and add a comment at the top explaining what it is?

aaron-kaplan:
BTW there are conflicts on this branch.

aaron-kaplan:
I often do both (rebase before and after). I suppose the feasibility of that approach depends in part on how comfortable you are with the rebase tool, and I know where you stand on that :-).

Agro-Climate:
Regarding #2, yes solar radiation is related to ET estimation only.

Agro-Climate:
Regarding #3, I agree that for now we just use the same fixed parameters from Uruguay. Let's see if we can get a chance to discuss these details with Watler and others later. (So for later response. I haven't yet pulled out the new scripts, but will do it soon).

aaron-kaplan:
Have you looked into using and/or extending existing libraries instead of writing our own? E.g. [xclim](https://xclim.readthedocs.io/en/stable/indices.html?highlight=water%20balance#xclim.indices.water_budget), [PCSE](https://pcse.readthedocs.io/en/master/index.html), github topic [soil-water-balance](https://github.com/topics/soil-water-balance).

Sorry, I know after you've written the code is not the optimal time to suggest this. Not sure why it didn't occur to me before.

If there's one or more that seems to provide the right functionality, I can be involved in evaluating the quality of the software.

aaron-kaplan:
What is `drain`? Is it "drain" as in "drainage"? If so, what's the difference between drainage and runoff? Is it an abbreviation for "daily rain"? If so, use daily_rain consistently. Is it "delta rain"?

remicousin:
[PastedGraphic-2_RC.pdf](https://github.com/iridl/python-maprooms/files/8266960/PastedGraphic-2_RC.pdf)

remicousin:
I am attaching a  graphic that illustrates the water balance "bucket" model. Currently we don't have Capillary Rise, Deep Percolation nor Irrigation. But the rest is basically it.

remicousin:
> Have you looked into using and/or extending existing libraries instead of writing our own? E.g. [xclim](https://xclim.readthedocs.io/en/stable/indices.html?highlight=water%20balance#xclim.indices.water_budget), [PCSE](https://pcse.readthedocs.io/en/master/index.html), github topic [soil-water-balance](https://github.com/topics/soil-water-balance).
> 
> Sorry, I know after you've written the code is not the optimal time to suggest this. Not sure why it didn't occur to me before.
> 
> If there's one or more that seems to provide the right functionality, I can be involved in evaluating the quality of the software.

I haven't. Maybe @Agro-Climate has? How do you suggest we proceed with regards to that?

remicousin:
Yes I've just passed on my mac and on devi. In both cases
34 passed, 63 warnings
(interestingly took 5s on Mac but 12s on devi)

remicousin:
There are 2 asserts for each of these. I can't see from your pasting which fail. In each test, one assert is a "reproducibility" test where "expected" values were gotten from running the very test and the other test is a sanity check test. In with_rho, et_crop and et_crop_red must be equal where SM = TAW (60) because then Ks = 1.   In with_et_crop, et_crop equals et_crop_red always because rho=None, so Ks is not applied, or rather is set to 1.
np.allclose() is involved and maybe that is sensitive to environment?

aaron-kaplan:
Please run `conda env export > environment_remi.yaml` on devi, after activating the environment you use for this project, and send me the output. If I can rebuild that environment and the tests pass for me in it, then we'll know it's a library version problem. If not, we'll investigate other possibilities.

aaron-kaplan:
BTW I would like us not to get in the habit of ignoring warnings. When there's just a few of them it's easy to say "yeah, I know about that one, it's fine," but if we do that habitually, eventually there will be hundreds of them and we won't notice when one that's actually significant arises.

There's almost always a way to suppress a warning. Sometimes the warning is about a deprecated way of doing things, and we can get rid of it by switching to the new, sanctioned way of doing it. Occasionally there's nothing for us to fix, and in those cases we can explicitly suppress the warning, once we've investigated it and determined that it's not a problem. If you need help with this, send specifics.

remicousin:
> Please run `conda env export > environment_remi.yaml` on devi, after activating the environment you use for this project, and send me the output. If I can rebuild that environment and the tests pass for me in it, then we'll know it's a library version problem. If not, we'll investigate other possibilities.

remicousin:
I don't know how I closed this...

onset_maproom/calc.py
```diff
@@ -1,16 +1,295 @@
...
+        planting_delta + np.timedelta64(1, "D") - soil_moisture[time_coord][0]
+    ).rename("planting_delta")
+    planting_delta.attrs = dict(description="Planting Date")
+    return planting_delta
+
+
+def kc_interpolation(planting_date, kc_params, time_coord="T"):
+    """Interpolates Crop Cultivar values against time_coord
+    from Planting Date and according to Kc deltas
+    kc_params are the starting, ending and inflexion points
```
aaron-kaplan, line 202: inflection is misspelled

onset_maproom/tests/test_calc.py
```diff
@@ -78,6 +202,254 @@ def test_water_balance_et_has_T():
...
+    )
+    wat_bal = calc.soil_plant_water_balance(
+        precip_sample(),
+        calc.hargreaves_et_ref(temp_avg, temp_amp, ra),
+        60,
+        10,
+        kc_params=kc_params,
+        rho=None,
+        runoff=calc.weekly_api_runoff(precip_sample()),
+    )
```
aaron-kaplan, line 364: In this test (which is the only one I've looked closely at so far), `daily_rain` and `runoff` are 1D (T is the only dimension) whereas `et` is 2D (T, Y). Is this a realistic scenario? Why would some of these parameters vary by location and others not? It all works out because of a call to `broacast(water_balance)` in the main function. I'm wondering whether that `broacast` is doing something useful, or if it's just obscuring mismatched data shapes when it would be more useful to bring it to the programmer's attention.

aaron-kaplan (commented):
There's a lot here, and I haven't made it through all of it, but I'm sending what I have so far.

Smaller, more incremental PRs are often easier to review. It's not always possible to work that way, but in this case it seems like it might have been.

onset_maproom/calc.py
```diff
@@ -53,6 +332,228 @@ def water_balance(
...
+    water_balance["et_crop"] = water_balance.et_crop.copy()
+    # Give time dimension to sminit
+    sminit0 = xr.full_like(
+        water_balance.soil_moisture
+        .isel({time_coord: 0}).expand_dims(dim=time_coord),
+        sminit,
+    )
+    sminit0 = sminit0.assign_coords(
+        {time_coord: sminit0[time_coord] - np.timedelta64(1, "D")}
+    )
```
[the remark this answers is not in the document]
aaron-kaplan: I'm rereading this comment that I made last week (but didn't submit until yesterday), and it doesn't make sense to me. I suspect `T=10` was a typo, but I'm not sure what I meant to type. I still think the `isel` is a bit confusing, since you're not actually selecting values from `soil_moisture`, only from its time dimension. Maybe something like
```
t0 = water_balance['soil_moisture']['T'][0] - np.timedelta64(1, 'D')
sminit0 = xr.DataArray(coords=dict(T=[t0]), data=[sminit]))
```

onset_maproom/tests/test_calc.py
```diff
@@ -5,6 +5,129 @@
...
+            ra.isel(T=0, Y=0),
+        ]
+        < [
+            ra.isel(T=-1, Y=-1),
+            ra.isel(T=0, Y=0),
+            ra.isel(T=-1, Y=0),
+            ra.isel(T=-1, Y=0),
+            ra.isel(T=-1, Y=0),
+        ]
+    ).all()
```
aaron-kaplan, line 80: I haven't attempted to understand what you're testing here, but I strongly suspect that this code doesn't do what you meant. The things you're applying the `<` operator to are python lists, not numpy or xarray arrays, and the `<` operator on lists works completely diffferently. For example,
```
>>> [1, 2] < [2, 1]
True
```
Happy to go into more detail if you'd like, but the bottom line is don't try to do array math with lists.

onset_maproom/calc.py
```diff
@@ -53,6 +332,228 @@ def water_balance(
...
+    # Create or Initialize Kc
+    if kc_params is None:
+        kc0 = xr.ones_like(water_balance.et)
+    else:
+        if p_d is not None:
+            kc0 = kc_interpolation(p_d, kc_params, time_coord=time_coord)
+        else:
+            p_d_find = planting_date(
+                sminit0, sm_threshold, time_coord=time_coord
+            ).expand_dims(dim=time_coord)
```
aaron-kaplan, line 448: In the PR description you said,

> Planting Date is of same nature as onset/cessation_date which is that it has a T grid that from which the variable expressed in timedeltas refers to. Right now there is only one planting date (see more below in After this PR section). I realized when I added planting date variable (p_d) to the water_balance dataset, that I had to change its dim name not to be T, otherwise it would be meaninglessly extended over the other variables T. So that makes me think that there is maybe a flaw in the mental conception on how those _date variables should be... to be discussed

But here you're using `time_coord` as the dimension name. Has this problem been resolved?

I don't actually understand why the planting date array has a time dimension at all. It seems like it should only vary by X and Y.

onset_maproom/calc.py
```diff
@@ -53,6 +332,228 @@ def water_balance(
...
+        water_balance.soil_moisture[{time_coord: i}] = (
+            water_balance.soil_moisture.isel({time_coord: i - 1}, drop=True)
+            + water_balance.peffective.isel({time_coord: i}, drop=True)
+            - reduce_crop_evapotranspiration(
+                water_balance.et_crop
+                .isel({time_coord: i}).expand_dims(dim=time_coord),
+                ks,
+            ).squeeze(time_coord)
+        ).clip(0, taw)
+    # Save planting date if computed
```
aaron-kaplan, line 510: I would suggest that it's better to save it regardless of whether it was computed or provided as input. Being inconsistent about which variables are included could lead to errors when people assume that something they see in an example output will always be present.

onset_maproom/calc.py
aaron-kaplan: The reference point of the timedelta could be stored as an attribute rather than a dimension.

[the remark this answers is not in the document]
aaron-kaplan: I think this relates to point 9 in your description:

> et_crop is built/saved inside the loop, whereas et_crop_red and drain, outside the loop. It could be either way for the 3 variables. Any sense if a way should be better?

I think you're actually calculating et_crop_red and drain twice each, once inside the loop for use in the stepwise calculation and then again at the end for saving into the result. The second time seems redundant, which makes the program both slower and more likely to be wrong (if we modify the calculation in one place and forget to modify the other).

aaron-kaplan:
How is the multi-year version going to work? Will we have new initial soil moisture measurements for the beginning of each season, or do we use the estimated balance from the previous year to start the new year? Hopefully the former, because in the latter case errors would accumulate, and the calculation would be harder to implement efficiently.

Assuming we have new initial conditions for each year, it might be best to decompose year and day-of-year into separate dimensions. If that works out, we might consider rewriting the onset date calculation that way too. I must confess that I never really got my head around daily_tobegroupedby_season.

onset_maproom/calc.py
```diff
@@ -53,6 +332,228 @@ def water_balance(
...
+    peffective = (daily_rain - water_balance.runoff).rename("peffective")
+    peffective.attrs = dict(description="Effective Precipitation", units="mm")
+    water_balance = water_balance.merge(peffective)
+    # Additional vars and coords
+    et = (et * xr.ones_like(water_balance.peffective)).rename("et")
+    water_balance = water_balance.merge(et)
+    if kc_params is not None:
+        for adim in kc_params.dims:
+            for thedims in water_balance.dims:
+                if adim != thedims and adim != "kc_periods":
```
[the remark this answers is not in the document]
remicousin: adim and thedims are both strings while kc_param.dims and water_balance.dims are both lists of strings.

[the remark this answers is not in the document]
aaron-kaplan: Oh, yeah. Then why is it named "thedims", plural?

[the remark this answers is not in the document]
remicousin: will fix that

onset_maproom/calc.py
```diff
@@ -53,6 +332,228 @@ def water_balance(
...
+    soil_moisture.attrs = dict(description="Soil Moisture", units="mm")
+    water_balance = water_balance.merge(soil_moisture)
+    et_crop = xr.full_like(
+        water_balance.peffective, np.nan
+    ).rename("et_crop")
+    et_crop.attrs = dict(description="Crop Evapotranspiration", units="mm")
+    water_balance = water_balance.merge(et_crop)
+    (water_balance,) = xr.broadcast(water_balance)
+    water_balance["soil_moisture"] = water_balance.soil_moisture.copy()
+    water_balance["et_crop"] = water_balance.et_crop.copy()
```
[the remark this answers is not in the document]
remicousin: if not copy, I get error:
ValueError: Assignment destination is a view.  Do you want to .copy() array first?
which I read that broadcast makes "views" then need to "copy" to use again. I am definitely not on top of that. I believe you commented on broadcasting at some point. So we can check that back later. If no more broadcasting, no more copy.

onset_maproom/config-sample.yaml
```diff
@@ -4,7 +4,15 @@ core_path: /onset-maproom
 admin_path: /onset-maproom-admin
 tile_path: /onset-maproom-tiles
 daily_rainfall_path: /data/aaron/ethiopia-rain-rechunked
+daily_tmin_path: /data/remic/mydatafiles/ANACIM/anacim_v3/ALL_block2/tn_mrg_daily
+daily_tmax_path: /data/remic/mydatafiles/ANACIM/anacim_v3/ALL_block2/tx_mrg_daily
```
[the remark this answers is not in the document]
remicousin: They are used by enactstmaxtozarr.py and enactstmintozarr.py
It's similar problem discussed in other PR that we have data curation, calculus and marpooms all in same place.

onset_maproom/enactstmintozarr.py
```diff
@@ -0,0 +1,39 @@
...
+
+tmin_mrg = xr.open_mfdataset(
+  TMIN_MRG_FILE,
+  preprocess = set_up_dims,
+  parallel=False
+)
+
+tmin_mrg.to_zarr(
+  store = TMIN_MRG_ZARR_PATH
+)
```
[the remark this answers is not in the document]
remicousin: I would tend to think: 1 dataset - 1 script because scripts could be part of the data curation system. Can we agree on what we are doing next on that front before I make changes there?

onset_maproom/enactstmaxtozarr.py
```diff
@@ -0,0 +1,39 @@
...
+      int(xda.encoding["source"].rpartition("/")[2].partition("tmax_mrg_")[2].partition("_")[0][4:6]),
+      int(xda.encoding["source"].rpartition("/")[2].partition("tmax_mrg_")[2].partition("_")[0][6:8])
+    )])
+    xda = xda.rename({'Lon': 'X','Lat': 'Y'})
+    return xda
+
+tmax_mrg = xr.open_mfdataset(
+  TMAX_MRG_FILE,
+  preprocess = set_up_dims,
+  parallel=False
```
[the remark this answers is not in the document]
aaron-kaplan: Yes. Using dask gives us out-of-core processing, regardless of whether parallel processing is enabled or not.

aaron-kaplan:
Regarding the tests that failed for me but passed for you, when I came back the following Monday they passed for me too. I don't know what changed :-(.

remicousin:
> Regarding the tests that failed for me but passed for you, when I came back the following Monday they passed for me too. I don't know what changed :-(.

Spring!

onset_maproom/calc.py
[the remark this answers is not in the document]
remicousin: For planting_date and kc_interpolation, I believe your comment is correct. The current end result looks like this because I tried to reconcile:
1/ one loop fit all (which I thought was a goal at some point);
2/ modularity by having planting_date and kc_interpolation be "stand alone" functions that could be replaced by another.
As of now, the only other option I can think of is to have cases trigger different loops. Some cases will have planting_date and kc outside the loop, and some outside. Those inside the loop will be their own function, those outside will be the vectorized version. The case of planting_date is less trivial because it can vary in the other dimensions so you can have null in some locations and non-null (thus "savable" elsewhere).

The et_crop_red/drain is partially true. They are not computing per say in the loops but could trivially add subsets to "save" them there. I set up both cases (inside/outside the loop) as it's not clear to me what's better. I have a sense that the least in the loop, the better, but maybe that is not the right rule of thumbs.

What do you think of all that?

onset_maproom/tests/test_calc.py
remicousin: Yes, it is a realistic scenario as basically et can depend on crops. Calling it Y here is copy-and-past laziness leading to confusion, sorry. Typically, the climate variables will depend on space (X/Y, station, shape) and T (or S/L) (and potentially ensemble member or this kind of stuff). Then the soil dependent data (TAW) would depend on space only. Then the crop dependent data (that is kc_params and thus et_crop and beyond) could depend on "crop." A user could choose kc_params=None and input an et that is already crop dependent.

This being said, it is possible that broadcast is unnecessary. Now I am wondering if all my expands and squeezes deal with the problems that made me broadcast. Or maybe I just need to add a few. I do get a bunch of "shape mismatch" error when commenting out broadcasting.

onset_maproom/calc.py
remicousin: Yes so I think that one last sentence suggests you understood what I meant. in onset/cessation, we used a dimension to store the reference time point. And it is typically the same day of the year, year after year. So here I used the same model because at some point (next step after this PR), we'll need a planting date that varies with years too. Planting date can vary with years, obviously if it's estimated within the lo; but also if the user gives it as input. For instance the user could use onset_date to estimate planting dates and feed that as input. Or they could have independent planting date data from somewhere else.

onset_maproom/calc.py
remicousin: Can do. There is however the case where kc_params = None where there is no planting date and Kc is always 1. Should I still save p_d as null or something?

onset_maproom/calc.py
```diff
@@ -53,6 +332,228 @@ def water_balance(
...
+    ).squeeze(time_coord)
+    water_balance.soil_moisture[{time_coord: 0}] = (
+        sminit
+        + water_balance.peffective.isel({time_coord: 0})
+        - reduce_crop_evapotranspiration(
+            water_balance.et_crop
+            .isel({time_coord: 0}).expand_dims(dim=time_coord),
+            ks,
+        ).squeeze(time_coord)
+    ).clip(0, taw)
```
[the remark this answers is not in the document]
remicousin: Correct. Since this came up twice,  I will wait for your reaction to my reaction and then act accordingly.

remicousin:
> There's a lot here, and I haven't made it through all of it, but I'm sending what I have so far.
> 
> Smaller, more incremental PRs are often easier to review. It's not always possible to work that way, but in this case it seems like it might have been.

It was hard for me to stop before I get a comprehensive soil_plant_water_balance function... And the cases kind of emerged to be dealt with altogether. I don't know... Maybe it's my strategy to writing that is wrong from the get-go. I will try to be more careful with that when I know it's going to be a big piece. And here I knew.
Or I was waiting to be blocked but that didn't happen.

onset_maproom/calc.py
```diff
@@ -1,16 +1,295 @@
...
+# Data Reading functions
 
 
 def read_zarr_data(zarr_path):
     zarr_data = xr.open_zarr(zarr_path)
     return zarr_data
 
 
-# Growing season functions
+# Water Balance functions
```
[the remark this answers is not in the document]
remicousin: I split it off into an agronomy.py and its test file. I didn't do it for test data because some are used by both. Seemed more arbitrary to pick which test data goes where (as surely we don't want to duplicate, do we?). Let me know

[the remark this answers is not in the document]
remicousin: And actually I wanted to rebase and test that the older water_balance function is used by the app from agronomy rather than calc... but forgot to do that...

onset_maproom/config-sample.yaml
```diff
@@ -4,7 +4,15 @@ core_path: /onset-maproom
...
 rr_mrg_nc_path: /Data/data23/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/
 rr_mrg_zarr_path: /data/remic/mydatafiles/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/rr_mrg_ALL/
+tmin_mrg_nc_path: /Data/data23/ANACIM/anacim_v3/ALL_block2/tn_mrg_daily/
+tmin_mrg_zarr_path: /data/remic/mydatafiles/ANACIM/anacim_v3/ALL_block2/tn_mrg_daily/
+tmax_mrg_nc_path: /Data/data23/ANACIM/anacim_v3/ALL_block2/tx_mrg_daily/
+tmax_mrg_zarr_path: /data/remic/mydatafiles/ANACIM/anacim_v3/ALL_block2/tx_mrg_daily/
+nc_path: /data/remic/mydatafiles/soilgrids/GYGA_Results_ll/Aggregated_overERZD/
```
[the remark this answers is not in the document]
remicousin: I am not sure what to document. The nc_path are where the data is read from; the zarr_path where the data is written.

remicousin:
> > I must say that I have another take on code readability. I find harder to read code that has a bunch of small functions invoked because it makes you go out and about to read the definition of that function and back...
> 
> This is true. It's only helpful for understanding if you can find a name for the function that explains what it does clearly, succinctly, and precisely enough that the reader doesn't have to go read the definition in order to understand what it does. Sometimes that's hard.
> 
> The real problem for me is that when a single function goes on for three full pages, I have a hard time keeping track of everything that has happened. Having a succinct summary that breaks it down into components makes it easier for me. If you don't like splitting it into smaller functions, another way to give me that help would be to write a summary in a comment at the top of the function, and then add more comments in-line as the code progresses telling me which step of that summary happens where.
> 
> Indeed if we can figure out a way to get more of the squeezing/expanding and broadcasting to happen automatically, that would help with readability too.

I will get back to this once we figure the broadcasting business and other currently unresolved problems. Then will try and do some code rationalization, via functions definitions and/or better commenting. So basically once ready to be approved.

remicousin:
I am just going to put here a brief recap of outstanding issues. They all are detailed on conversation above, sometimes with open resolution pending:

- is broadcasting needed (thus copy)?
- clean up conversion scripts (rationalize + document)
- should test_data be split into 2 files?
- keep trying to have only 1 loop? or move to loops per cases? (this will have an impact on p_d/kc computation)
- "save" all variables as loop(s) go (no vectorized recomputation after the loop(s)
- Establish a strategy for multiple years and merge this PR in a state that is compatible with that
- Always save p_d? (there is a case where there is no p_d)
- improve overall code readability (small functions + commenting)

onset_maproom/tests/test_agronomy.py
```diff
@@ -317,19 +317,27 @@ def test_soil_plant_water_balance_with_et_crop():
...
+
+       [[60.        ,  3.27655007]],
+
+       [[59.68250804,  1.27869775]],
+
+       [[60.        , 13.15848481]],
+
+       [[60.        , 14.09643733]],
+
+       [[60.        , 14.15885817]]]
```
aaron-kaplan, line 330: Was it black that added all the blank lines?

remicousin: no... sorry... It's just a copy and past from terminal, from printing the result to be checked in test for reproducibility. I will make nicer.

onset_maproom/calc.py
remicousin: oh yes i see... if years is a dimension from splitting, then the reference day-month can be a constant attribute. As things are now, the 2 are combined on a coord.

Splitting into 2 dimensions will need a bit more subtlety than managing the leap years because it will also need to cover the case where a "year" is not a "calendar year" because the season of interest overlaps 2 "calendar years." I guess that is why the grouping was appealing because it doesn't require to have a workable 2D array for years/doy; you just need the mechanism to "tag" the days you care about.

But I am game with either way.

I can clarify the T name business. At the very end, I merge p_d in the dataset to be part of the outputs, because its dimension is called T, the resulting p_d is expanded over the whole daily T of the other variables (with I guess many NaNs if I recall). That is why I renamed T of p_d just before merging.

So yeah.. it's funny that the very first things we had to start coding are the ones that are the most subtle with managing time dimensions...

Right now, I don't have a strong feeling for which route to take. If there is going to be reduction, groupby seems more straightforward, if not, moving to 2D seems more relevant, with all the subtleties mentioned (and maybe more).

aaron-kaplan: In the FBF application the season months (e.g. DJF) are a constant, and we have a year dimension. We define the year to be the year of the midpoint of the season, so T=2000 refers to the season that runs Dec 1999 to Feb 2000 (whose midpoint is Jan 15 2000). (I think it would have been more convenient if we had used the start or end of the season instead of the midpoint.) Seems like we could do the same here.

We can try the 2D approach first for this case (in a subsequent PR, not this one), where groupby isn't applicable, and then we'll see how we like it and decide whether we want to rewrite onset date to use that approach instead of groupby.

Thanks for explaning the dimension name business, I understand now.

onset_maproom/calc.py
aaron-kaplan: The two options that come to mind are to omit it in that case, or to make it an array of NaNs with the appropriate shape. Given my ignorance of how it will be used, I don't know which would be better. Probably not a very consequential decision.

onset_maproom/config-sample.yaml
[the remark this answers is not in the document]
aaron-kaplan: I think I was asking about the provenance of /data/remic/mydatafiles/soilgrids/GYGA_Results_ll/Aggregated_overERZD/. I assume the other datasets are sufficiently documented by their catalog entries.

onset_maproom/config-sample.yaml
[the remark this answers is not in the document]
aaron-kaplan: `daily_tmax_path` does not occur in `enactstmaxtozarr.py`. It uses `tmax_mrg_nc_path` and `tmax_mrg_zarr_path`. Likewise for tmin.

onset_maproom/config-sample.yaml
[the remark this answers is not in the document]
remicousin: ah got it... GYGA is acutally in the SOURCES:
http://iridl.ldeo.columbia.edu/SOURCES/.SoilGrids/

onset_maproom/agronomy.py
```diff
@@ -0,0 +1,534 @@
...
+    peffective.attrs = dict(description="Effective Precipitation", units="mm")
+    water_balance = water_balance.merge(peffective)
+    # Additional vars and coords
+    et = (et * xr.ones_like(water_balance.peffective)).rename("et")
+    water_balance = water_balance.merge(et)
+    if kc_params is not None:
+        for adim in kc_params.dims:
+            for thedim in water_balance.dims:
+                if adim != thedim and adim != "kc_periods":
+                    water_balance[adim] = kc_params[adim]
```
aaron-kaplan, line 386: The variable name issue has been resolved, but I still don't understand these nested loops. As far as I can see, the following is equivalent (and shorter):
```
for adim in kc_params.dims:
    if adim != "kc_periods":
        water_balance[adim] = kc_params[adim]
```

aaron-kaplan: Ah, maybe what you intended was
```
for adim in kc_params.dims:
    if adim not in water_balance.dims and adim != "kc_periods":
        water_balance[adim] = kc_params[adim]
```

onset_maproom/agronomy.py
```diff
@@ -0,0 +1,534 @@
...
+    if p_d is not None:
+        for adim in p_d.dims:
+            for thedim in water_balance.dims:
+                if adim != thedim and adim != time_coord:
```
aaron-kaplan, line 390: same comment as above.

onset_maproom/calc.py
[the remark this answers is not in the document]
aaron-kaplan: If this still isn't clear, I can try expressing myself in code rather than in English.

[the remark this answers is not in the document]
aaron-kaplan: Regarding et_crop_red and drain, I'm not sure I understand your point of view. Yes, if you could calculate it outside of the loop, and not inside, that would be better, but that's not an option, because each step's soil moisture depends on the previous step's ET and drainage. The best feasible option is to do it inside the loop. Doing it inside the loop and then again after the loop exits is clearly worse than that. Maybe I'm still missing something.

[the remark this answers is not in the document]
remicousin: ok. So we can have planting_date, when estimated, being "saved" step after step of the loop; got back to previous step rather than origin of times; use where or something to search for the new time step where it's not found yet. That seems workable. Maybe only because planting_date depends only on conditions in one day (as opposed to onset_date for instance the looks at at n-day window). But that is probably going to be modul-able later on if such a scenario arise.

After I've done that I'll have a look at Kc in the same spirit.

For et_crop_red/drain, indeed, et_crop_red is literally computed in the loop by reduce_crop_evapotranspiration. Drainage is not as things stand (the clip filter destroys the Drainage information) but it's a trivial matter to add to the loop to save it before clipping. So that is what I meant, that in that very case, Drainage is not computed inside the loop (but close). I don't think you are missing something, I think it's me who is being extremely literal in trying to keep things outside the loop.

So I'll work on all that now (ish) unless you raise a flag reading that conversation.

onset_maproom/agronomy.py
aaron-kaplan: It's not just more succinct, it's correct whereas your version is not.
```
d1 = dict(a=1, b=2)
d2 = dict(a=3)
for k1 in d1.keys():
    for k2 in d1.keys():
        if k1 != k2:
            d2[k1] = d1[k1]
print(d2)
# prints {'a': 1, 'b': 2}

d1 = dict(a=1, b=2)
d2 = dict(a=3)
for k1 in d1.keys():
    if k1 not in d2:
        d2[k1] = d1[k1]
print(d2)
# prints {'a': 3, 'b': 2}
```

onset_maproom/calc.py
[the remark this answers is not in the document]
aaron-kaplan: No red flag.

onset_maproom/config-sample.yaml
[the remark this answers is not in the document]
aaron-kaplan: So is the file in /data/remic/mydatafiles/soilgrids/GYGA_Results_ll/Aggregated_overERZD/ downloaded from Ingrid? If so, add the Ingrid URL as a comment here.

Also the variable name `nc_path` could be made more precise. There are mutliple netcdf datasets involved in this project.

onset_maproom/calc.py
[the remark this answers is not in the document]
aaron-kaplan: This conversation is resolved as far as I'm concerned, but I'll leave it open because your last comment is a reminder to yourself. Feel free to resolve when it's no longer needed.

onset_maproom/tests/test_agronomy.py
```diff
@@ -354,9 +354,9 @@ def test_soil_plant_water_balance_with_et_crop_pd_none():
 
     assert wat_bal.p_d == pd.Timedelta(days=4)
     assert (
-        wat_bal.et.isel(T=slice(0, 4)) == wat_bal.et_crop.isel(T=slice(0, 4))
+        wat_bal.et.isel(T=slice(0, 3)) == wat_bal.et_crop.isel(T=slice(0, 3))
     ).all()
-    assert (wat_bal.et.isel(T=5) != wat_bal.et_crop.isel(T=5)).any()
+    assert (wat_bal.et.isel(T=4) != wat_bal.et_crop.isel(T=4)).any()
```
aaron-kaplan, line 359: Why did this test have to be changed? The change to how planting date is calculated was a performance improvement that wasn't supposed to affect the result.

onset_maproom/calc.py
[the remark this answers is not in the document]
aaron-kaplan: With the latest changes, `test_soil_plant_water_balance_with_et_crop_pd_none` has gone from 1.6s to 1s, which is a significant speedup but stll slower than I would expect given the size of the data.

Do you anticipate that this function will be used in maprooms, or only in offline/batch scripts? If in maprooms, I think we need to continue working on performance (I can help). If it's only going to be run in batches, maybe it doesn't matter.

onset_maproom/config-sample.yaml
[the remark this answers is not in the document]
aaron-kaplan: I suggest you remove them and add back when they're needed, but I won't insist on it. Resolve this conversation whenever you're ready to.

onset_maproom/tests/test_calc.py
aaron-kaplan: Did you want to rename `Y`? Resolve this conversation if you're finished with it.

onset_maproom/agronomy.py
```diff
@@ -0,0 +1,535 @@
...
+        if kc_params is not None and p_d is None:
+            p_d_iter = planting_date(
+                water_balance.soil_moisture
+                .isel({time_coord: i - 1}).expand_dims(dim=time_coord),
+                sm_threshold,
+                time_coord=time_coord,
+            )
+            # it's desired and useful to update Kc if
+            # p_d was NaT and is now found
+            kc_updates = np.isnat(p_d_find) and not np.isnat(p_d_iter)
```
aaron-kaplan, line 490: While investigating the performance issue, I happened to print the value of `kc_updates` here and was surprised to find that it's a Boolean scalar. I think you intended for it to be an array of Booleans, since `p_d_find` and `p_d_iter` are both arrays. The problem is that `and` is a normal python operator, not an array operator.

The solution to the immediate problem is to use `&` instead of `and`. But the fact that this bug went unnoticed raises bigger issues. First of all, why do the tests pass despite this bug? Can we add a test case that would have caught it? Second, why are `p_d_find` and `p_d_iter` arrays anyway? Is that so we can handle multiple years? But we're not handling multiple years yet. Can we simplify this for now now by making them scalars instead?

onset_maproom/agronomy.py
```diff
@@ -0,0 +1,535 @@
...
+            p_d_find = p_d_find.where(
+                lambda x: not np.isnat(x),
```
aaron-kaplan, line 492: Haven't confirmed this one with a print, bu i believe there's a simliar problem here. `not` is a standard python operator, not an xarray/numpy elementwise operator. Should use `~` instead.

How did you figure out that `where` accepts a function (lambda expression) as its first argument? The [doc](https://xarray.pydata.org/en/stable/generated/xarray.where.html) doesn't mention that.

onset_maproom/agronomy.py
```diff
@@ -0,0 +1,535 @@
...
+                other = p_d_iter + p_d_iter[time_coord] - p_d_find[time_coord]
+            )
+            kc = kc.where(
+                not kc_updates,
```
aaron-kaplan, line 496: Should be `~kc_updates`. Again, the fact that this went unnoticed probably means that the tests are insufficient.

onset_maproom/agronomy.py
aaron-kaplan: whoops, and the `not` should be replaced with `~`.

aaron-kaplan: I tried making the changes I suggested here and below (replacing `and` with `&`, `not` with `~`), and it doesn't work. I may have misunderstood the intent. Unless you know how to fix it, maybe we should work through this together.

onset_maproom/tests/test_agronomy.py
remicousin: Yes, that is on my radar, I've done some preliminary tests but chose to push nonetheless for code review.

First, I thought I had screwed up the comparisons, as you might have noticed I am struggling often with those. That wasn't the case. Then, there is something else I need to review, maybe once we are closer to merging. There are several T grids involved in this whole thing. First, there is the T of rain, tmin, tmax (thus et) that we can assume are the same. Second, runoff is derived from a rolling, so it's T is shorter than rain's. I need to drop the heading T points since we are doing a loop, if we start with NaN, we'll just propagate NaNs. Third, there is Kc, who doesn't care about the other Ts as kc_interpolation creates it from planting date until the last day of the growing season.

Now, everything is initialized as full_like of peffective (ie taking into account the days dropped by rolling) then merged. So everybody should be on same T but Kc. You might have noticed that I select on T by location for all, but not for Kc that I select by label, using location of soil_moisture's T.

So I suspect that something was (or is still) off with those selections. I've checked the 5 tests (by printing results) before pushing and all seemed coherent. So possibly it was wrong before the push. But I'd like to give another thorough look at this when some other ongoing changes are validated.

Now I think I also need to double check the cases where the simulation is run outside the growing season as I am not confident this is covered anymore...

onset_maproom/calc.py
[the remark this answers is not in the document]
remicousin: Historically, the initial desire was to run the simulation via a Maproom that controls some of the parametric inputs; but in Ingrid, it wasn't possible (too long to run) so we ran several simulations of interest once and then Maprooms crunched the daily outputs of that ([e.g.](http://maproom.meteorwanda.gov.rw/maproom/Agriculture/index.html#tabs-2) each of these has a different Kc+planting). I think on a short-term, PIs can be amenable to similar set up. In longer term, they might push for being able to run the simulation via a Maproom.

There was also always a desire to have Monitoring Maprooms using such simulation, but it never happened because there never was real-time temperature data. On short-term, that is likely going to remain so, but on mid-term, they might start producing those real-time ENACTS temperature. But the performance issue is a lesser problem then because you'd want to run like 1 year of simulation. Unless PIs want to compare real-time with long-term (e.g. anomalies), in which case, we are back to first paragraph...

So in short, my best guess is that performance issues could be addressed in a later PR, given Jim said this is going to be a deliverable for AICCRA but not ACToday; and I can get a better sense of what functionality he seeks in the meantime. Unless continuing to work on performance is going to shake the code quite a bit, in which case maybe it's better to do it now?

onset_maproom/agronomy.py
remicousin: Did you notice I was terrible with Python comparisons? :)
Let me have a look for myself. In the meantime, one piece of information for you: p_d/_find/_iter can vary like soil_moisture if they are estimated inside the function, so space at least. This is probably an initialization problem because initial soil moisture is sminit that has been a constant/scalar in all the tests. In the code at some point I bothered to give it a time dimension, I guess I should broadcast it against soil_moisture dims but T. And cover the cases where sminit is scalar or xarray (because here again, someone could have an sminit that depends on space already, or just want to test several initial conditions with some made up dim...)

Maybe I should just concat sminit to soil_moisture and keep going with soil_moisture... That's going to complicate the positioning selection mentioned earlier. I'll give it a thought, if you have ideas, I'll take them.

remicousin: I wonder now if that explains the shift by 1 day of the results of the tests mentioned earlier...

onset_maproom/config-sample.yaml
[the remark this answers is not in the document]
remicousin: Resolved this with a commit removing those lines. This is all going to have to be revised when rebasing on master anyway since we now have 2 config files

onset_maproom/calc.py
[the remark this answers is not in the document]
aaron-kaplan: It's easier to work on performance when there's good test coverage (so you can verify that your performance changes haven't broken it), so let's not worry about it for now.

onset_maproom/agronomy.py
remicousin: The not is fixed...

ah... errr.... intuition? Should I revisit?

onset_maproom/agronomy.py
remicousin: I am going to mark this resolved but also add note to think of more tests. I did a bunch of prints and the behavior expected with respect to planting date, et and et_crop seemed appropriate. It may mean that it was wrong before. But I figure I would make best use of my time not to figure that out, but rather to think of more tests

remicousin:
Ok. so now left on my to-do list, hopefully before merging (and assuming the recent pushes are fine):

- how to figure whether broadcasting is needed or not?
- more tests on Kc, planting_date and related
- clean up / readability
- rebase and consequences

And then left for later PRs:
- case when there is no more Kc values passed the growing season
- multiple years (the 2 entries might go together)

remicousin:
So I believe broadcast is the most substantive item whereas the others are cosmetics or trivial.

If I comment out the broadcasting part, I get this type of error when running the tests, when I do a first initilization for index 0 of T:

at:


    water_balance.et_crop[{time_coord: 0}] = crop_evapotranspiration(
        water_balance.et.isel({time_coord: 0}).expand_dims(dim=time_coord),
        kc0,
        time_coord=time_coord,
    ).squeeze(time_coord)

ValueError: shape mismatch: value array of shape (1, 2) could not be broadcast to indexing result with 0 dimensions

and that is because the left side of the = has a dimension size of 0, whereas the other side has a shape (1, 2) (for Y and X dimensions, squeezing having removed the singleton T).

If is remove the squeeze, I get the following error at the same location (see further down). I can't interpret it very well but basically I understand that squeezing fixes it by removing one of the 2 T so there is no more conflict.

Then broadcast fixes the shape mismatch problem above.

And so I am not sure how to go about it. Maybe if there is a way to assign a value to
water_balance.et_crop[{time_coord: 0}]
without having that drop the T dim, that would be a start.
I am also thinking that if there is not broadcasting, I should not select by positioning but rather by label as I can't be too sure then that all variables would start at the same T.


E               IndexError: dimension coordinate 'T' conflicts between indexed and indexing objects:
E               <xarray.DataArray 'T' (T: 1)>
E               array(['2000-05-07T00:00:00.000000000'], dtype='datetime64[ns]')
E               Coordinates:
E                 * T        (T) datetime64[ns] 2000-05-07
E               vs.
E               <xarray.Variable ()>
E               array('2000-05-07T00:00:00.000000000', dtype='datetime64[ns]')

aaron-kaplan:
I commented out the `.broadcast` and ran the tests. One of the tests that fails is `test_soil_plant_water_balance_with_hargreaves`. This test sets up a scenario where `water_balance.et_crop` has only one dimension, T, whereas `crop_evapotranspiration()` returns an array with two dimensions, T and Y. You then try to populate a 0d (scalar) slice of `et_crop` with a 1d slice of the `crop_evapotranspiration` result. This fails because the shapes don't match. `broadcast` makes the error go away by broadcasting `et_crop` across Y, making it 2d.

Was `et_crop` intende to vary over space? If so, why was it initially created as a 1d array?

remicousin:
I've stirred away from early variable declarations and broadcasting all together. No more full_like stuff and broadcast. I let the initialization calculation naturally give a shape to the variable, then add the time dimension. Now the outputs have only the dimensions they are meant to have. There are cases where outputs are also constant in time by design but my impression is that it may not be worth discriminating those cases.

I also revisited all the tests as it seems I keep having weaknesses in selection and testing which, combined or not, had had some irrelevant or wrong tests.

onset_maproom/agronomy.py
```diff
@@ -367,123 +368,93 @@ def soil_plant_water_balance(
     reduced (or actual) evapotranspiration,
     drainage
     """
-    # Start water balance ds with runoff
+    # Runoff
     if runoff is None:
-        runoff = xr.zeros_like(daily_rain).rename("runoff")
-        runoff.attrs = dict(description="Runoff", units="mm")
-    water_balance = xr.Dataset().merge(runoff)
+        runoff = xr.DataArray(np.zeros(1)).squeeze("dim_0")
```
aaron-kaplan, line 373: Is this different from saying `runoff = 0`? I thought xarray was capable of broadcasting a number to a DataArray when it needs to. (Similarly elsewhere for other constants.)

remicousin: In those cases, it's not a broadcasting matter, but just to be able to apply xarray functions (.i/sel, .rename, merge, .dims, .where) to the variable regardless whether it's a number or already an xarray. Either I do it like this, or I make cases each time they (runoff, kc, et) are used. Or maybe there is another path?
The runoff case is simpler as it's not interfering with the loop so it's xarray nature is only useful to merge it to the dataset output at the end. kc and et are more involved.

remicousin: Any chance to look further into this, @aaron-kaplan ?

aaron-kaplan: That makes sense.

I think you can create the 0-d DataArray more succinctly:
```
runoff = xr.DataArray(0.)
```

onset_maproom/agronomy.py
```diff
@@ -367,123 +368,93 @@ def soil_plant_water_balance(
...
-    sminit0 = sminit + xr.zeros_like(
-        water_balance["soil_moisture"].isel({time_coord: 0})
-        .expand_dims(dim=time_coord).assign_coords({time_coord: [t0.values]})
+    t0 = peffective[time_coord][0]
+    tinit = t0 - np.timedelta64(1, 'D')
+    sminit = sminit + xr.zeros_like(
+        peffective.isel({time_coord: 0})
+        .expand_dims(dim=time_coord)
+        .assign_coords({time_coord: [tinit.values]})
     )
```
aaron-kaplan, line 383: Sometimes you multiply by `ones_like(...)`, sometimes you add `zeros_like(...)`. Do you have some difference in meaniing in mind when you choose one over the other? Are they both equivalent to some invocation of `broadcast`? I think using `broacast` might make the intent clearer... although I know how to use `ones_like/zeros_like` but I don't know how to use `broadcast` without looking it up, so maybe not.

aaron-kaplan (approved):
There's still a lot I haven't made it through, but I had already given up on that, and the latest changes look good.

onset_maproom/agronomy.py
remicousin: just pushed a commit to simplify creation of constant arrays

onset_maproom/agronomy.py
remicousin: So there is one case left of * ones_like and another of + zeros_like. I don't think there is a reason to use one or the other, other than the thought of the moment.
And I am not using broadcast because in both those cases, I don't know whether what I am adding those 0s or multiplying those 1s to, whether they are a constant or a DataArray already.

remicousin:
I still see problems. I am tempted to start over, making small steps/PR. I feel the efforts to cover all cases, to generalize, and to extract code into functions so that the whole thing can be modulated are colliding into a poor result. Plus in some cases the generalized functions are vectorial in time (which makes sense as their own independent thing) but since the main water balance function is a loop on time, this generalization is kind of silly.
It might not be a huge amount of work as many things have been written, but the end result may make more sense.

remicousin:
And if so, before I get started, we need to have a plan A on how to deal with multiple years. Ingrid runs over the whole period of time for which there data, which is not great because we typically care about one season (at worst half the year) each year. We could still do that if it makes some sense otherwise.
If not, I was planning on using my groupby thing but you seemed inclined to rather use some sort of grid splitting.
The year-to-year variability in the code will come from the planting_date, from which the Kc values start to evolve. planting_date will be either prescribed as one single date (for all years), or a an array with yearly time dimension (and I guess possibly other dimensions if someone wants to apply different planting date in space, or have "scenarii' of planting dates for instance), or be "discovered" by the code in the loop.
That's the whole management of Kc and planting date that I am not too happy with in the present code and make me think that a fresh restart, going step by step, would be better.
In which case, in might need a quick call to discuss the multi-year strategy.

aaron-kaplan:
For monthly or dekadal data I have a feeling that splitting the T grid would make things easier, but I don't know how we would do it for daily data. The day-of-year dimension would have to have length 365 or 366, and either way some of the years won't fit. If you're counting runs of wet or dry days in a window that spans Feb 28/Mar 1 (or across day 365/0 if we go with a numerical day-of-year), having an extra day or a missing day in the T dimension will throw off the count. 

Hmm, the iteration over days of the year is a python loop. Maybe we could make the day-of-year dimension size 366, and when the loop gets to the leap day it has special logic that makes it do one thing for the leap years and something else for the non-leap years.

In this particular case, it's not a terribly consequential decision, because the choice of T representation is an internal implementation detail that isn't visible to the caller (the output only has a year dimension, regardless of whether we split the T grid internally or not). You can try it and see how it goes, or you can stick with the current approach and we'll be free to revisit it another time.

aaron-kaplan:
(And if you want to talk, I'm open to it.)

remicousin:
Actually the output of the water balance is daily. It's just that we don't
really care about what's happening of the growing season, modeled by Kc.
What makes me recall that my groupby thing is pointless here because it
does mean to reduce the daily values, and have eventually yearly output.

But so I guess you're right. I can start something that is thought for a
stream of days covering one season, then we can see how to deal with
multiple years. Or we discuss this some time next week, to check on the
pieces that will vary with years and so that I don't engage in a dead end.

What needs to be able to vary with years is the planting date. And then if
we decide to run only chunks of days rather than continuously, a new SM
initialization.

PR #65, opened 2022-03-17
Rebuild environment_linux.yml

aaron-kaplan:
The one Kevin built a couple weeks ago is now no longer installable--the solver reports conflicts. We have yet to figure out why, but I think something must have changed upstream, like a package we were using got tagged "broken" and thus the solver no longer considers it valid. This lack of reproducibility is a major concern and must be addressed, but that's not what I'm doing in this PR. I'm just working around the problem so we can keep moving on the enactsmaproom project.

### DIFF
--- a/onset_maproom/environment_linux.yml
+++ b/onset_maproom/environment_linux.yml
@@ -7,7 +7,6 @@ dependencies:
   - _openmp_mutex=4.5=1_gnu
   - affine=2.3.0=py_0
   - alsa-lib=1.2.3=h516909a_0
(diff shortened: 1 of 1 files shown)


remicousin:
I am approving without doing anything. If you'd like me to test those changes on things have been working on, let me know

remicousin (approved):

aaron-kaplan:
I held off merging this because I thought it was causing tests on Rémi's water balance branch to fail, but those tests pass for Rémi even with this environment, so the problem must be elsewhere.

PR #66, opened 2022-03-18
Remic/whenprepare

remicousin:
As I am moving to making a similar but different "onset" maproom for Senegal, I followed Aaron first ideas to structure the enacts maprooms source code and make the app configurable per country (or institution). Note that it's not fully working ideally now so this PR is about discussing and implementing the right things.

N.B.:

1. I didn't get the logo to work in spite of different referencing configurations;
2. I am passing none to display attribute to trigger or untrigger parts of the Maprooms. In this case: Senegal doesn't have cessation business. This worked well for the cessation controls not to show up but it still displays the tab for the cessation graphs. If that is just a matter of using more proper css, cool. This is convenient because it means the configuration involves only layout, but I am thinking that it may not be great because I assume that at app start, it still is doing cessation calculations for Senegal, correct? Discuss...
3. Will probably need a more generic text for the top section of the text, then there is likely going to be quite of tailoring in text needed. That can make big config files but I guess that is ok if they are well written and thus navigable.
4. Ah and also see README on how to run the app. Relative imports got my eyebrows to pop up so... I implemented whatever made me less sweat.


aaron-kaplan:
I looked into the icon business, and it seems not trivial to resolve under the current organization. I think we should separate this into two distinct parts: 1. generalizing the onset maproom for use in Senegal, 2. changing the directory structure to accommodate multiple maprooms in a single project. You can just focus on 1 and I will do 2.

It's probably easiest to do one and then the other, rather than working on them simultaneously and then trying to reconcile. We can do it in whichever order you prefer.

enacts/config-anacim-senegal.yaml
```diff
@@ -0,0 +1,39 @@
...
+listen_port: 8063
+#Jeff set up host 0.0.0.0 port 8000 to run app on  http://shortfin01.iri.columbia.edu:8000/onset-maproom/ for demos
+
+# ENACTSvZARR
+daily_rainfall_path: /data/remic/mydatafiles/ANACIM/anacim_v3/ALL_block2/rr_mrg_daily
+lat0: 12.225
+latf: 16.8
+lon0: -17.5875
+lonf: -11.2125
+lonlatres: 0.0375
```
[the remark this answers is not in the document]
remicousin: That is very fine with me but I think some day you said "no data reading in layout.py" and then since layout is called by maproom, layout can not call maproom where data reading is done.
Which proposition to follow?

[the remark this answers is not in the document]
aaron-kaplan: True--I don't want data files to be read when a module is imported, because that has negative consequences like making tests run slower, and the layout function is called when maproom.py is imported, _ergo_ we shouldn't read data files in the layout. Any information we want to get from the dataset should be removed from the layout, which is static, and set by a callback instead.

(To be clear, datasets are already being read when the maproom module is imported, but that's something I want to change.)

aaron-kaplan:
> Are you suggesting that I revert the changes I made with regards to # 2?

Yes, that's what I'm suggesting. If that's too hard/annoying for some reason, we can discuss alternatives.

enacts/config-anacim-senegal.yaml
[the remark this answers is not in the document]
remicousin: I am running into one problem with this which is the precision problem. For instance Y[0] returns:
12.224999999999998
instead of 
12.225
and I am not sure how to automatically deal with that... That's a challenge for the edges and resolution...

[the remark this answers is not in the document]
remicousin: looks like np.around is going to help

remicousin:
Ok so we can now talk about configuring the same Maprooms for 2 different countries and slightly varying purposes in this PR.
And I addressed the 1st suggestion, ie to use callbacks to retrieve mostly spatial info from data rather than from config file.

enacts/config-anacim-senegal.yaml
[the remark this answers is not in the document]
aaron-kaplan: What happens if you just use the value (12.224999999999998) that you get from the dataset?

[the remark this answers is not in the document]
aaron-kaplan: I see, it's about displaying a value in the interface. `np.around` looks like the right solution.

onset_maproom/config-nma-ethiopia.yaml
```diff
@@ -1,13 +1,31 @@
...
+institution: NMA
+
+# Climate and Ag
+
+# Onset_and_Cessation
+onset_and_cessation_title: Historical Onset and Cessation Date
+default_search_month: Jun
+default_running_days: 5
+default_min_rainy_days: 3
+isonoff_cess_date_hist: inline-block
```
aaron-kaplan, line 31: This variable's name is misleading. From the name, I expected it to take a Boolean value, but it seems it's actually a string to be used as the value of an HTML `display` attribute. This could be fixed in either direction, but I recommend making it a Boolean, and letting the code fill in either `none` or `inline-block` depending on that Boolean value. That gives us the flexibility to change the mechanism by which components get shown/hidde, without having to change all the countries' config files.

onset_maproom/layout.py
```diff
@@ -410,6 +398,7 @@ def results_layout():
                     dbc.Spinner(dcc.Graph(id="probExceed_cess")),
                 ],
                 label="Cessation Date",
+                style={"display": CONFIG["isonoff_cess_date_hist"]}
```
aaron-kaplan, line 401: In the browser dev tools I can see that the cessation date graph is being fetched from the backend, even though it will never be displayed. Let's use a different mechanism for hiding it (and hide the tab while we're at it).

The first thing that comes to mind is to give the `dbc.Tabs` element no children in the layout, and set the children in a callback.

aaron-kaplan (changes requested):

onset_maproom/layout.py
remicousin: I have a solution in the last couple commits. It inflates a bit the repeat codes for trying and alerts for graphs that we said we wanted to rationalize but didn't get to.
I am not sure I need to create alert figures for the cast no tab... Can I just return None?

aaron-kaplan: I'm not sure if returning None will work in this particular context. Try returning None or `{}`. If you really have to return a Figure, you can probably do it in one line instead of ten--skip the call to `add_annotation` and just return a Figure object with no content.

aaron-kaplan: Aside from that simplification, this approach seems fine. Another approach would be to generate the tab(s) in a callback rather than in the static layout, so that in the Senegal case the cessation tab is simply not there, rather than being present with display: none. I don't know offhand what the relative advantages/disadvantages would be.

remicousin: yeah... if this maproom is going to get more complex in the future, with more options that can be triggered on/off, maybe it's easier to manage the layout if all possible elements are in there. Don't know... thinking out loud...

onset_maproom/maproom.py
```diff
@@ -58,21 +59,38 @@ def toggle_navbar_collapse(n, is_open):
...
+    lat_max = str(np.around((rr_mrg.Y[-1]+lat_res/2).values, decimals=10))
+    lon_res = np.around((rr_mrg.X[1]-rr_mrg.X[0]).values, decimals=10)
+    lon_min = str(np.around((rr_mrg.X[0]-lon_res/2).values, decimals=10))
+    lon_max = str(np.around((rr_mrg.X[-1]+lon_res/2).values, decimals=10))
+    lat_label = lat_min+" to "+lat_max+" by "+str(lat_res)+"˚"
+    lon_label = lon_min+" to "+lon_max+" by "+str(lon_res)+"˚"
+    return center_of_the_map, lat_min, lat_max, lon_min, lon_max, lat_label, lon_label
+
+
+    Output("latInput", "placeholder"),
```
aaron-kaplan, line 85: ?

onset_maproom/maproom.py
```diff
@@ -58,21 +59,38 @@ def toggle_navbar_collapse(n, is_open):
...
+    Output("map", "center"),
+    Output("latInput", "min"),
+    Output("latInput", "max"),
+    Output("lngInput", "min"),
+    Output("lngInput", "max"),
+    Output("latLab", "children"),
+    Output("lonLab", "children"),
+    Input("submitLatLng","n_clicks"),
+)
+def initialize(toto):
```
aaron-kaplan, line 72: It was I who suggested that we find the center coordinates in a callback rather than in the layout, so that we could get it from the dataset instead of duplicating the information in the config file. Here we see a disadvantage of doing it that way: a callback must have inputs, so you were forced to declare an arbitrary one even though this function doesn't really depend on it. Now this callback will be called unnecessarily each time the user submits the form.

I'm thinking of a couple of alternative setups, but it would be easier to implement them than to explain them. This is fine for now, and I will try alternatives as part of making a multi-page ENACTS maproom setup.

aaron-kaplan (approved):
@remicousin are you ready to merge this or did you have further changes planned?

aaron-kaplan:
I approved and then un-approved because I don't think the unnecessary callback to fetch the cessation date graph has been eliminated yet. There's also a line at maproom.py:85 that doesn't make sense.

remicousin:
sorry about that...

I am not sure what you are thinking for the cessation callback. In the Senegal case, where it's turned off, the callback is called on opening of app or if changing map location and goes into the "do nothing" case. Then it can never be called by the other controls it's meant to respond to since they are not displayed thus not updatable.

Well I could tell that by making prints. I don't know what to do with the browser dev' tools about it.

aaron-kaplan:
> I am not sure what you are thinking for the cessation callback. 

My main concern was to avoid unnecessary calculation on the server, and I had failed to notice that you added a "do nothing" case in the callback. This should be fine.

FWIW, my previous suggestion was to generate (or not generate, depending on config) the cessation date `dbc.Tab` in a callback instead of making it static in the layout.

aaron-kaplan (approved):
Please use the "squash and merge" option when merging. A lot of the changes you originally had on this branch were later reverted, which could lead to spurious conflicts when people rebase other branches onto this one after it's merged. "Squashing" reduces the branch's history to a single commit, eliminating the intermediate steps that you took along the way.

remicousin:
I didn't ignore your other suggestion but you didn't seem to prefer one over the other in your comment. The rationale that made me pick one is wrong (not mentioned here). The rationale to keep it that way (better readability for layout file I'd say) is weak.

aaron-kaplan:
I repeated the suggestion only because you said you didn't know what I was thinking. What you did is equally valid.

PR #68, opened 2022-03-25
fix scrolling artifact

kgraaf:
Naively adding borders to the DataTable causes a strange artifact in some circumstances where the fixed and non-fixed portion fall out of alignment by a single pixel. Currently I have a very convoluted/esoteric workaround that involves duplicating the first row of the data. This PR eliminates that in favor of a more straightforward solution that applies a left border to the container element that contains both the fixed and non fixed portions of the table.

### DIFF
--- a/fbfmaproom/fbflayout.py
+++ b/fbfmaproom/fbflayout.py
@@ -409,11 +409,21 @@ def table_layout():
                             "height": "650px",
                             "maxHeight": "650px",
                         },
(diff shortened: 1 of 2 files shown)


aaron-kaplan:
How did you come up with this?

aaron-kaplan (approved):

kgraaf:
Vague intuition that's a product of CSS-induced brain damage

PR #73, opened 2022-04-05
Add worst_pnep column to csv download if freq is specified

aaron-kaplan:
Requested by Souha.

### DIFF
--- a/fbfmaproom/tests/test_fbfmaproom.py
+++ b/fbfmaproom/tests/test_fbfmaproom.py
@@ -296,13 +296,12 @@ def test_download_table_one_freq():
             '&geom_key=ET05'
             '&freq=30'
         )
-    #print(resp.data)
     assert resp.status_code == 200
     assert resp.mimetype == "text/csv"
     csv_file = io.StringIO(resp.get_data(as_text=True))
     df = pd.read_csv(csv_file)
     assert list(df.columns) == [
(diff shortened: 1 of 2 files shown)


fbfmaproom/tests/test_fbfmaproom.py
```diff
@@ -296,20 +296,21 @@ def test_download_table_one_freq():
...
-       'time', 'bad_year', 'obs', 'enso_state', 'pnep_30',
+        'time', 'bad_year', 'obs', 'enso_state', 'worst_pnep', 'pnep_30',
     ]
     onerow = df[df["time"] == "2019-04-16"]
     assert len(onerow) == 1
     assert onerow["bad_year"].values[0] == 0.0
     assert np.isclose(onerow["obs"].values[0], 3902.611)
     assert np.isclose(onerow["pnep_30"].values[0], 33.700)
     assert onerow["enso_state"].values[0] == "El Niño"
+    assert onerow["worst_pnep"].values[0] == 0
```
kgraaf, line 312: Why would the zeroth row always be 0?

aaron-kaplan: There's only one row in `onerow`, and it's the row for 2019-04-16 (see line 306). That year happens not to be one of the n worst years (whatever n is in this case).

PR #74, opened 2022-04-05
fixed niger commune table name

kgraaf:
table was added with wrong name to database originally, fixed and updated config

### DIFF
--- a/fbfmaproom/README.md
+++ b/fbfmaproom/README.md
@@ -1,0 +1,27 @@
+# FbF Maproom
+
+## Adding a foreign table
+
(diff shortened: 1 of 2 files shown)


aaron-kaplan (approved):

kgraaf:
@aaron-kaplan I went ahead and added the instructions we discussed to this PR. Let me know if those look good.

kgraaf:
> Instead of using generic/placeholder host, dbname, and local and foreign user names, could you specify the names we actually use?

Ok, I wasn't sure how sensitive that was, but I suppose besides passwords it's all public

kgraaf:
done

aaron-kaplan (approved):

PR #75, opened 2022-04-07
Fix download_table bug introduced in last release

aaron-kaplan:
The good news is that the tests caught this. The bad news is that apparently I forgot to run the tests. I really want to run them in CI, but currently a lot of the tests require the prod database, which won't be available in a CI environment. We need a test data fixture.


aaron-kaplan:
... apparently we both forgot to run the tests!

aaron-kaplan:
@kgraaf no new functionality here, just a bug fix that's needed to fix a test that's failing on master. Please review promptly.

kgraaf (approved):

PR #77, opened 2022-04-08
Remic/fauxdep

remicousin:
Ignore all commits but last. See Description for purpose


remicousin:
This is ready for review.

It turns out this is NOT the function needed for the Senegal project so its review is NOT urgent.

Rebasing was done. Environment update is necessary to use the component bottleneck.

Current warnings come from previous work so I didn't want to mess around with those here. I'll fix those in another PR. Maybe right now.

There is probably no point looking at the history of commits but just the final result. The function and algorithm are described/commented where defined.

Feel free to add other appropriate reviewers.

remicousin:
I am confident that the function is correct. I tested the different scenarios for the heading and tailing days to be either wet or dry ones which for some reason was were I was less confident.

aaron-kaplan:
Sorry, I just merged a branch that also changed environment.yml, so now there are conflicts and you need to rebase again. Also, when you change environment.yml you need to update environment_linux.yml.... and then test the application carefully, including parts you didn't touch, because recreating the environment may result in library updates that break things. Kevin can help with this if it doesn't get done before I leave on vacation.

remicousin:
okey. Can you write down instructions somewhere on the generation of environment_linux.yml? It seems that it's something that we may have to do frequently enough to be worth writing down, and sparse enough for folks like me to forgot how to.

onset_maproom/environment.yml
```diff
@@ -13,12 +13,14 @@ dependencies:
   - opencv
   - psycopg2
   - rasterio
-  - scipy
```
aaron-kaplan, line 16: Did you intend to remove this? Seems like it might have happened accidentally when you rebased.

remicousin: No: I added it when I though I was going to use interpolate_na (or something like that) and then realized bafill was more appropriate, and the one requires bottleneck rather then scipy.

remicousin:
We might want to merge this if there is no red flags. The function is not going to be used by anything anyways.

onset_maproom/calc.py
```diff
@@ -51,6 +51,37 @@ def water_balance(
...
+    Assign cumulated dry days there, Nan elsewhere
+    Propagate backwards and the 0s at the tail
+    And that is going to be the offset
+    Apply offset that is correct for all days followed by dry days
+    Eventually reset days followed by wet days to 0
+    """
+    # Find dry days
+    dry_day = ~(daily_rain > wet_thresh) * 1
+    # Cumul dry days backwards and shift back to get the count to exclude day of
+    dry_spell_length = dry_day.reindex({time_coord: dry_day[time_coord][::-1]}).cumsum(
```
aaron-kaplan, line 71: Give this variable a better name. It's only an intermediate step on the path to calculating dry spell length, one of many, and the others all have meaningful names.

aaron-kaplan (approved):
I have a feeling the double `reindex` will be expensive if we ever run this on big datasets. That's ok, we can explore alternative implementations when and if it becomes a bottleneck.

Despite your extensive comments, I didn't understand the calculation until I stepped through one of the tests and printed out the intermediate results at each step. This suggests that it might be helpful (for me, at least) to walk us through an example line-by-line in the comments.

onset_maproom/calc.py
remicousin: I am pretty conservative on naming inflation after 10+ years of rolling stacked objects... I see I can change my ways...

remicousin:
I resolved the recent inputs and added doc. I will rebase now and see what happens...

remicousin:
How am I supposed to resolve the conflicts on environment_linux.yml? My understanding is that probably none are correct until I finish rebasing the rest, rebuild the resulting environment and re-test all, correct? And then generate a new environment_linux.yml

aaron-kaplan:
Right, resolve the conflicts in environment.yml and then generate environment_linux.yml from the result.

aaron-kaplan:
Looks good, but you should walk through all the functionality in a browser to make sure the library upgrades didn't break anything. I don't think it's important to test all the configurations for both countries, just one.

remicousin:
ok, let me do that now.

remicousin:
I was working on my Mac but now I need be on devi to test serving apps. Since I pushed --force-with-lease, I need to fetch and then rebase remic/fauxdep on origin/remic/fauxdep. That seems to have me run into same conflicts as the master rebasing I did on my Mac a couple of hours ago. Doesn't that seem right to you or I am engaging in the wrong workflow...?

aaron-kaplan:
Assuming you haven't done any work on devi that you forgot to push, just
```
git fetch
git checkout remic/fauxdep
git reset --hard origin/remic/fauxdep
```
which discards whatever you had in that branch on devi before.

remicousin:
Ethiopia Onset Maproom behaved. This last commit is on the docstring documentation that I forgot to test after rebasing. So I think we are all good to merge.

aaron-kaplan:
I reapprove.

PR #78, opened 2022-04-12
Install dash-bootstrap-components and dash-leaflet from conda-forge instead of PyPI

aaron-kaplan:
For some reason, pip pulled in old versions of dash_core_components and dash_table, libraries that have been folded into dash and are no longer needed as separate components.

Removing environment_macos.yaml because nobody has been maintaining it. I hope to switch to conda-lock soon, which should help with managing environments for multiple platforms.

Unrelated to this change, a new version of flask/werkzeug has come out since we last built the environment and our application isn't compatible with it for some reason. Pinning it until we figure out the problem.


aaron-kaplan:
Added a second commit that does the same for the onset maproom. While I was at it, also removed unused dependencies dash_daq and dash-extensions, and I removed flask from the list of explicit dependencies becuase it's already a dependency of dash.

fbfmaproom/environment_linux.yml
```diff
@@ -181,19 +185,12 @@ dependencies:
...
-    - dash-html-components==2.0.0
-    - dash-leaflet==0.1.23
-    - dash-table==5.0.0
-    - geobuf==1.1.1
-    - protobuf==3.19.4
     - pyaconf==0.7.1
     - pyyaml==6.0
     - queuepool==1.3.1
-prefix: /home/aaron/scratch/miniconda3/envs/fbfmaproom2
+prefix: /home/aaron/scratch/miniconda3/envs/dashtest
```
aaron-kaplan, line 196: @kgraaf commented that he usually deletes this line. My response is that it doesn't seem to make a difference, so I'd just as soon not bother with that manual step every time we update anything.

kgraaf (approved):

fbfmaproom/environment_linux.yml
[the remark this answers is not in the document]
aaron-kaplan: I commented about this same topic, and you resolved the comment. Then 15 minutes later you opened this comment. Was this a delayed appearance of a comment you made earlier today? Weird.

fbfmaproom/environment_linux.yml
```diff
@@ -1,44 +1,47 @@
-name: fbfmaproom2
+name: dashtest
```
aaron-kaplan, line 1: Oops, I unwittingly changed the name of the environment here, and the container build depended on it. I fixed that directly on master--now we use `conda env create -n app` so this line is ignored.

PR #79, opened 2022-04-12
Use processes instead of threads in onset apache config

aaron-kaplan:
We don't get any real parallelism from threads, probably due to the GIL. We changed to processes for the dev configuration in PR #70, but I forgot to do the same for the prod configuration.


kgraaf:
Do you want me to test it? It looks good to me, though I am curious why processes actually get you parallelism but threads do not

aaron-kaplan:
If you know how to test it, suggestions are welcome, but I'm ok merging it
before testing.

kgraaf:
If you already ran it in the dev config that's probably fine then

kgraaf (approved):

PR #81, opened 2022-04-13
SQL updates for new shapefiles

kgraaf:


aaron-kaplan (approved):

aaron-kaplan:
IIRC we're waiting to merge this until the new vulnerability data are ready, because the old vuln uses the old shapes and the new vuln uses the new shapes.

kgraaf:
@aaron-kaplan That is correct

PR #82, opened 2022-04-13
find length of longest spell

remicousin:
This new function has immediate vocation to be applied to a rolling to find longest dry spell in a window of n days following a given day.
Will be usable for wet spell, or hot/cold spells, or really whatever data/flagging/dimension/step.

And then to feed into stuff like rolling but also daily_groupby, etc.

Please feel free to add reviewers.

### DIFF
--- a/onset_maproom/calc.py
+++ b/onset_maproom/calc.py
@@ -51,23 +51,23 @@ def water_balance(
     return water_balance
 
 
-def longest_run_length(flagged_data, coord):
-    """ Find the length of the longest run of flagged (0/1) data along a coordinate.
+def longest_run_length(flagged_data, dim):
+    """ Find the length of the longest run of flagged (0/1) data along a dimension.
     
     A run is a series of 1s not interrupted by 0s.
-    The result is expressed in the units of `coord`, that is assumed evenly spaced.
+    The result is expressed in the units of `dim`, that is assumed evenly spaced.
     
     Parameters
     ----------
     flagged_data : DataArray
         Array of flagged data (0s or 1s)
-    coord : str
-        coordinate of `flagged_data` along which to search for runs
+    dim : str
+        dimension of `flagged_data` along which to search for runs
         
     Returns
     -------
     DataArray
-        Array of length of longest run along `coord`
+        Array of length of longest run along `dim`
         
     See Also
     --------
@@ -82,7 +82,7 @@ def longest_run_length(flagged_data, coord):
     Cumulative flags, where kept, need be propagated by bfill
     so that diff returns 0 or the length of runs.
     
-    I believe that it works for unevenly spaced `coord`,
+    I believe that it works for unevenly spaced `dim`,
     only we then don't know what the units of the result are.
     
     Examples
@@ -97,17 +97,18 @@ def longest_run_length(flagged_data, coord):
     Attributes: description:  Longest Run Length
     
     """
+    
     # Special case coord.size = 1
     lrl = flagged_data
-    if lrl[coord].size != 1:
+    if lrl[dim].size != 1:
         # Points to apply diff to
         unflagged_and_ends = (flagged_data == 0) * 1
-        unflagged_and_ends[{coord: [0, -1]}] = 1
+        unflagged_and_ends[{dim: [0, -1]}] = 1
     
-        lrl = lrl.cumsum(coord).where(unflagged_and_ends, other = np.nan).where(
+        lrl = lrl.cumsum(dim=dim).where(unflagged_and_ends, other = np.nan).where(
             # first cumul point must be set to 0
-            lambda x: x[coord] != lrl[coord][0], other=0
-        ).bfill(coord).diff(coord).max(coord)
+            lambda x: x[dim] != lrl[dim][0], other=0
+        ).bfill(dim).diff(dim).max(dim=dim)
     lrl.attrs = dict(description="Longest Run Length")
     return lrl
 
--- a/onset_maproom/tests/test_calc.py
+++ b/onset_maproom/tests/test_calc.py
@@ -74,6 +74,7 @@ def test_longest_run_length_coord_size_1():
     assert lds == 1
     assert lds0 == 0
 
+
 def test_water_balance_intializes_right():
 
     precip = precip_sample()


onset_maproom/calc.py
```diff
@@ -51,6 +51,29 @@ def water_balance(
...
+    Because diff needs at least 2 points,
+    we need to keep (where) the unflagged and first and last
+    with the cumulative value for last
+    and 0 for first
+    """
+    # Points to apply diff to
+    unflagged_and_ends = (flagged_data == 0) * 1
+    unflagged_and_ends[{coord: [0, -1]}] = 1
+    
+    ls = flagged_data.cumsum(coord).where(unflagged_and_ends, drop=True).where(
```
aaron-kaplan, line 69: I don't think this will work as you intended with multidimensional inputs. `where` with `drop=True` only drops a coordinate label if *all* the mask values with that label are False. That is, if the input has dimensions (T, Y, X), then T=i will be dropped only if mask[i, x, y] is False for all values of x and y.

Demonstration: if the DataArray has only one dimension, then there are no NaNs in the result of the `where`
```
data = [1, 0]
da = xr.DataArray(data=data)
print(da.where(da, drop=True))
# <xarray.DataArray (dim_0: 1)>
# array([1.])
```
but if it has two dimensions, there are NaNs in the result.
```
data = [[1, 0], [0, 1]]
da = xr.DataArray(data=data)
print(da.where(da, drop=True))
# <xarray.DataArray (dim_0: 2, dim_1: 2)>
# array([[ 1., nan],
#        [nan,  1.]])
```

onset_maproom/calc.py
```diff
@@ -51,6 +51,29 @@ def water_balance(
...
+        # first cumul point must be set to 0
+        lambda x: x[coord] != flagged_data[coord][0], other=0
```
aaron-kaplan, line 71: Possibly moot given my previous comment, but...

IIUC this is here because without it, if the 0th point is flagged then the 0th cumulative sum is 1, and consequently the length of the first run gets counted as n - 1 when it should be n.

This solution was hard for me to understand, and it's still not obvious to me that it's correct in all cases, even though I tried and failed to find a case where it's wrong. Wouldn't it be simpler to pad `coord` on the left with a slice of zeros, e.g. with `xr.concat`?

onset_maproom/calc.py
remicousin: good catch

onset_maproom/calc.py
remicousin: You understand very well. It covers the case where the 1st day is flagged and contributes to the longest run. I am not able to explain it very well either. I think the best way to see it is that diff needs pairs so its result drops a data point and there is an origin issue.
I am not super excited about concatenating. Adding size to an object... plus is it not going to be a headache depending on the nature of the coord (regular step, or not...). Is it not going to create trouble when feeding this to a rolling?

Also, after I pushed, I realized that: I claimed that it would work for whatever dimension/step, but that is less clear to me now. The "length" that is computed is in deltas of steps. So if 9, it's 9 "steps." And that is likely true only for regular coords. If irregular, it's not 9 but it should be a sum of 9 deltas that vary... That brings back both the matters of working with units, and with coordinates having width, or not...

Maybe I should not try and make this that general....?

aaron-kaplan: > I think the best way to see it is that diff needs pairs so its result drops a data point and there is an origin issue.

Exactly, that's why adding an extra data point seems like the cleanest solution. I don't care enough to insist on it, though.

onset_maproom/environment.yml
```diff
@@ -19,6 +19,7 @@ dependencies:
   - xarray
   - yaml
   - zarr
+  - bottleneck
```
aaron-kaplan, line 22: Please update environment_linux.yml as follows (on a linux machine):
- conda deactivate
- conda env remove -n enactsmaproom
- conda env create -f environment.yml -n enactsmaproom
- conda activate enactsmaproom
- conda env export > environment_linux.yml

Then test the application thoroughly, because in addition to adding bottleneck this will update everything to the latest version, and it's possible that one of our dependencies released a breaking change since the last time we rebuilt the environment.

(I'm not bothering to document this somewhere more permanent because I'm working on what I hope will be a better way of managing environments.)

onset_maproom/tests/test_calc.py
```diff
@@ -5,6 +5,55 @@
 import data_test_calc
 
 
+def test_longest_run_length():
```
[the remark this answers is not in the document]
aaron-kaplan: I retract this comment. There's a way to specify the exact test name rather than a pattern that might match multiple tests:

```
python -m pytest tests/test_calc.py::test_longest_run_length
```

onset_maproom/calc.py
```diff
@@ -51,6 +51,64 @@ def water_balance(
...
+    coord : str
+        coordinate of `flagged_data` along which to search for runs
+        
+    Returns
+    -------
+    DataArray
+        Array of length of longest run along `coord`
+        
+    See Also
+    --------
```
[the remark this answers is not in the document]
remicousin: I'd rather keep it as a reminder to document relations between the functions we build.

remicousin:
It seems because I rebased (and thus force-pushed) and/or because I worked on this branch on my Mac, I can't pull on the devi clone. That happened before with a branch of yours (Aaron) that you had rebased and you then said to fetch. But I'd rather double check with you before moving forward and because I am obviously not understanding things well in my manipulations across clones.

onset_maproom/calc.py
remicousin: you are right: I found cases that don't work. Concatenating with 0 is the right thing to do. However, after review, it turns out that doing nothing is equivalent to concatenating 0 (especially now that there is bfill).
If the first point is 0, it remains 0 through bfill, then either there were a run of n ones after and diff will return n-0 as we wish; and having concatenating 0 would have added the point 0-0 that doesn't interfere with max.
If the first point is 1, then whatever the rest of the series is, it's backfilled by bfill with the length of the run it belongs to, say n (n >= 1); then concatenating 0 preserves n = n-0 and we have what we want.

Thus morality: I just need to replace unflagged_and_ends with an unflagged_and_tail. That is still needed to cover the case where the last point is 1, so that the run it belongs to is taken into account. And at the same time it insures that there is at least 1 value, thus more than 1 after bfill, and thus diff can operate.

remicousin: I was wrong: i don't have a counter-example that doesn't work. And my proposed solution breaks the case where all are flagged, what shed some light on all this.

This:
"IIUC this is here because without it, if the 0th point is flagged then the 0th cumulative sum is 1, and consequently the length of the first run gets counted as n - 1 when it should be n."

is not correct. Or possibly is not correct anymore now that there is bfill. The consequence is that the length of the 1st spell would be counted as 0 (not n-1).

bfill propagates the value of the length of the first run, backwards, until there is a 0. If there is a 0, then that 0 remains 0 and diff gives the result we want. If there is no 0 (which also means that that first run starts on the first point), then the propagation goes all the way through and diff returns 0 for all those points. And thus, if it turned out that this was the longest run, max will miss it.

Consequently, concatenating an initial 0 covers that case, as our intuition tells us as we naturally understand pair-wise diff. Now, why is it equivalent to reset the first point to 0 before back filling? Because in that very case, we know the 1st point must be 1, thus the result of cumsum for the second point is at least 1 too (it's 1 or 2). Thus, the information in the 1st point must be contained in the 2nd and thus replacing the 1st point by 0 bears no consequences, it insures that the first run length is counted correctly, and it requires not use concat. That also means that the function is ambiguous in the case where the function is applied to a series of 1 point. If the series is 1 flagged point, then the where will try to replace that very point which is both ends at the same time by 0 and by 1. I don't know what it will do. Then diff will likely fail. Whereas the concat solution will return the right result in that very case.

Ok now. I am writing all this just to try to convince myself that I can explain what I code. I've been wrong before in my tentative explanations so I'm not 100% convinced by myself even now. Now I still have reservation with the concat solution (that we can however explain better), besides the fact that it increases the size of the object, is that won't I need to build a dim for that 0 with coords that somehow are anterior to the coord of the flagged_data? In other words I will have to make some assumption or get some info about the coord that so far I could ignore entirely? won't I?

remicousin: I pushed what I have so far. I guess either we are both convinced enough of what I wrote above, or not and I can be reasonable and implement concat solution

aaron-kaplan: [np.diff](https://numpy.org/doc/stable/reference/generated/numpy.diff.html) has an optional kwarg `prepend`, which looks like it exists for this exact reason.
```
>>> np.diff([1], prepend=0)
array([1])
```
But DataArray.diff doesn't support that option.

I still don't have an intuitive grasp of why your version works, but I worked through all the cases of length 2 and 3 and it gives the right answer every time, so I'm pretty convinced. You could add a special case for when the length of the sequence is 1.

Just for the sake of argument...

> won't I need to build a dim for that 0 with coords that somehow are anterior to the coord of the flagged_data? In other words I will have to make some assumption or get some info about the coord that so far I could ignore entirely? 

I think you could just drop the coord. Xarray tolerates dimensions without coordinates, and it doesn't seem like you need it (since you're only looking for the length of the longest run, not its location). Equivalently, you could extract the values as a numpy array and do the calculation in numpy (which has no notion of coordinates). But this is a moot point because your current implementation seems to work.

onset_maproom/tests/test_calc.py
```diff
@@ -5,6 +5,65 @@
...
+
+    precip = precip_sample()
```
aaron-kaplan, line 10: What do you gain by using `precip_sample` in these tests? `longest_run_length` works for any sequence of 0s and 1s, and the tests would be a lot easier to read if they looked like
```
assert calc.longest_run_length([0, 1, 1]) == 2
```
It's not quite that simple because the argument has to be a DataArray, but you could e.g.
```
def lrl(sequence):
    return calc.longest_run_length(xr.DataArray(sequence, dims='T'), 'T')

def test_longest_run_length():
    assert lrl([0, 1, 1]) == 2
```
The same helper function would work for all the tests except the 2d one.

aaron-kaplan: (Incidentally, if the DataArray has only one dimension, you might just look for sequences along that dimension instead of requiring that the caller name it. That seems to be how most numpy and xarray functions with an axis argument do it.)

aaron-kaplan (approved):
Consider adding a special case to handle sequences of length 1, and rewriting the tests as suggested to make them more readable.

aaron-kaplan (changes requested):
Whoops, I'm un-approving. When I run the app, I get deprecation warnings because queuepool is incompatible with pandas 1.4+. The installation of pandas 1.4 predates your branch, but it wasn't causing problems until you started making db queries.

Until we manage to fix or eliminate queuepool, please pin pandas to version 1.3 in environment.yml as we did [in fbfmaproom](https://github.com/iridl/python-maprooms/blob/master/fbfmaproom/environment.yml#L17-L21).

onset_maproom/tests/test_calc.py
remicousin: I think originally mostly because laze to create DataArrays.

If I trust xarray documentation, the behavior you describe on dim=None, axis=None doesn't apply to all functions. e.g. it applies to all axes if not specified for sum. It doesn't say what it would do for cumsum. It requires dim for bfill and diff. These last 2 make sense since you can't bfill/diff on all axes at once (maybe you could with diff which is a linear application).

So for now I've just renamed coord dim and stuck as much to what xarray's doc claims it does.

remicousin:
I am writing comments on my commits now. one second

remicousin:
> Whoops, I'm un-approving. When I run the app, I get deprecation warnings because queuepool is incompatible with pandas 1.4+. The installation of pandas 1.4 predates your branch, but it wasn't causing problems until you started making db queries.
> 
> Until we manage to fix or eliminate queuepool, please pin pandas to version 1.3 in environment.yml as we did [in fbfmaproom](https://github.com/iridl/python-maprooms/blob/master/fbfmaproom/environment.yml#L17-L21).

no queuepool here, only bottleneck, you must be using environment from monitpd branch.

aaron-kaplan:
> no queuepool here, only bottleneck, you must be using environment from monitpd branch.

Indeed.

onset_maproom/tests/test_calc.py
aaron-kaplan: Good research on xarray conventions/defaults. I still think in this case it would make sense to search along the lone dimension if there's only one and no dim name is specified, but it's not a blocker.

aaron-kaplan (approved):

PR #85, opened 2022-04-18
Remic/monitpd

remicousin:
The main goal of this PR was to have a default map for Senegal that find the latest found germinating rains (aka the wet spell of the onset definition). Currently, there is no near-real time data so it uses the data we have for this. That analysis will become relevant when the data updates, and/or we have this part of the analysis point to near-real time Monitoring ENACTS data.
Doing so, I re-enabled the control for choices for the map, namely climatological average and probability of exceeding, as they are also requested for Senegal project. ( I removed std. dev. as it turns out xarray std doesn't deal with NaT and std. dev. has not really be requested by orders).
I then added a series of contents/cosmetics to go along with those changes, applied all the changes to Ethiopia when relevant or through configuration to distinguish the needs of each country.
I then was carried away and set up the admin borders on top of the map, setting opacity to 1 and leaving Topo/Street in background, as done for FbF Maprooms.

@kgraaf , remember that I am still new to Dash & Co (and Widgets that I changed too) so please don't hesitate in criticizing as it is very likely that I did clumsy things or re-invented the wheel. @drewmresnick you can review as you please.

The commits are mostly clean and in a meaningful chronological order so it can be a good way to do the review.

You need not change any of the config files to run the app, apart  from picking your servers' choices.

@kgraaf , in spite of the review on-going, since things work (even if possibly not the right way), I'd like to show this to IRI colleagues. Is it difficult to set up a permanent URL somewhere?

N.B.:

- Changes to pingrid for colorscale business were approved in the pyCPT Maproom PR by @aaron-kaplan, even though that doesn't mean he hasn't in mind revision of what I did there (so this is just an FYI).
- I was not able to center the map title.
- admin layers are hard-coded for 3 per country. This probably needs to be tailorable. We actually have 4 admin levels for Senegal so there is a case study there (and there probably are places where we have only 2 levels). But probably fixing that doesn't belong to that PR, other than doing the hard-coded version right.
- admin layers callback use a dummy Input. That is an "import" from my work with pyCTP Maprooms. @aaron-kaplan said he would look into that process.
-  @kgraaf you will see what server I query to get the geometries, and in the config files, you will see the 6 tables involved

Todo (in addition to in-line comments):
- [x] Pin pandas=1.3
- [x] (Re-)adapt enactstozarr.py for Senegal
- [x] Do you want to use a coarser resolution and/or rechunk to speed it up?
- [x] More robust nan-handling for climatological map?


remicousin:
I forgot to mention that testing can be a bit tedious. I believe there is a coarser version of ENACTS Ethiopia that @aaron-kaplan set up somewhere but I don't know where...

kgraaf:
@remicousin Let me take some time to review this. In the mean time, yes I'm happy to set up a test environment for you and I'll send you an email when I've done that.

remicousin:
You mean I run it and let it run?

aaron-kaplan:
Yes. I usually run it in a tmux session so that it doesn't die when I disconnect (e.g. if my VPN session crashes).

remicousin:
I set this up:
devi:8061/when-prepare-planting/
It worked fine right after I set it up and detached screen.
Then Jim complained he got an error and now when I visit (after reconnecting VPN), it's just hanging (Firefox loading signal on and page remains fully blank). When I re-attach my screen, as far as I can tell (which is not very far), things looked normal in the Terminal...
Maybe I missed something? I don't think I changed the mode (debug etc.). Does that matter?

aaron-kaplan:
I don't know what's wrong. I see your process running, but it's not
responding to http requests. Try running it in prod mode, you might get
more informative messages in the terminal when something goes wrong.

remicousin:
About this last commit: We need the onset date to accept inputs so that it doesn't look for dry spells. In that case, onset = germinating rains and there is no question of false onset. This is needed for using the function in Monitoring mode, in particular. I had overlooked that case because in the Ingrid version, putting one of the 2 dry_spell_lenght/search to 0 was doing the trick. In Xarray, it doesn't work because rolling doesn't accept a window of 0.
I think this solution should do. I added a corresponding test and all the other tests passed.

aaron-kaplan:
> You need not change any of the config files to run the app, apart from picking your servers' choices.

I've removed the db password from the config files, so now you do need to modify them.

aaron-kaplan:
BTW `/data/aaron/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/rr_mrg_ALL` was produced by `enactstozarr.py`, which reminds me that we need an `enactstozarr.py` for Senegal.

remicousin:
> > admin layers are hard-coded for 3 per country. This probably needs to be tailorable. We actually have 4 admin levels for Senegal so there is a case study there (and there probably are places where we have only 2 levels). But probably fixing that doesn't belong to that PR, other than doing the hard-coded version right.
> 
> I'm fine with leaving it hard-coded for now if that's what you prefer, but you can look at fbfmaproom for an example of how to make it configurable per country. I see that you have a map layer per admin level; when the number of levels becomes configurable, you could handle that by having a loop that generates as many map layers as there are levels, rather than hard-coding three of them.

Did you have a clear idea on top of your head on how to do that? Right now (not pushed), I have functions to generate dlf.Overlay for each admin level from the info in the config file. This seems to work fine. Then I have a callback that loops to concatenate them in an array that I intend to return to dlf.LayersControls' children, adding to the existing elements. But I get an error that isn't understandable to me (e.props undefined) and when I print the result of concatenate, it doesn't seem right. Not sure.
I could use a clue as to direction to take as web search inspires me for different directions (I don't do the concatenation right; or I maybe should use something like addLayer; or I have the callback generates the whole dlf.LayersControls -- even though there is id=onset_layer in it that thus may require subtleties).

aaron-kaplan:
> I got it... Sometimes it just helps to write things down...

https://en.wikipedia.org/wiki/Rubber_duck_debugging

remicousin:
skouik skouik

aaron-kaplan:
I don't know what's wrong with the database queries. Can you just roll back and leave the number of layers fixed at 3 for now, and we'll deal with this separately? If not, push your latest version and I'll try to debug.

aaron-kaplan:
As I noted [here](https://github.com/iridl/python-maprooms/pull/82#pullrequestreview-958419202) (which was the wrong place for it), we need to pin pandas=1.3 in environment.yml (and then rebuild the environment and regenerate environment_linux.yml). The current version of queuepool is incompatible with pandas 1.4.

(I briefly wondered if this explained your current problem, but it doesn't seem to. The error message is completely different.)

aaron-kaplan (commented):
I started a review last week and never finished it. Here's what I've entered so far, will continue reviewing after I send it.

In addition to the in-line comments that are part of this review, I'll gather a summary of my other unaddressed comments here:

- [ ] Pin pandas=1.3
- [ ] Create an enactstozarr.py for Senegal, or adapt the existing one to handle both countries.
- [ ] Do you want to use a coarser resolution to speed it up?

onset_maproom/config-anacim-senegal.yaml
```diff
@@ -11,6 +11,19 @@ listen_port: 8063
 dev_processes: 4  # number of server processes in dev setup. Not used when running in apache.
 #Jeff set up host 0.0.0.0 port 8000 to run app on  http://shortfin01.iri.columbia.edu:8000/onset-maproom/ for demos
 
+dbpool:
+    name: iridb  # postgres database name
+    host: devi.iri.columbia.edu
```
aaron-kaplan, line 16: Did you add new tables to devi to support this project?

Could you try changing this to use dlcomputemon1 instead of devi? The db on devi was part of the plan for upgrading postgres, but the plan has changed, and I would like to avoid adding new clients of devi because eventually we need to turn it off.

onset_maproom/config-nma-ethiopia.yaml
```diff
@@ -11,6 +11,19 @@ listen_port: 8050
 dev_processes: 4  # number of server processes in dev setup. Not used when running in apache.
 #Jeff set up host 0.0.0.0 port 8000 to run app on  http://shortfin01.iri.columbia.edu:8000/onset-maproom/ for demos
 
+dbpool:
+    name: iridb  # postgres database name
+    host: devi.iri.columbia.edu
```
aaron-kaplan, line 16: devi -> dlcomputemon1

onset_maproom/maproom.py
```diff
@@ -119,10 +215,45 @@ def map_click(click_lat_lng):
...
+)
+
+def write_map_title(search_start_day, search_start_month, yearly_stats_input, probExcThresh1):
+
+    if yearly_stats_input == "monit":
+        search_start_month1 = calc.strftimeb2int(search_start_month)
+        first_day = rr_mrg.precip["T"][-366:-1].where(
+            lambda x: (x.dt.day == int(search_start_day)) & (x.dt.month == search_start_month1),
+            drop=True,
+        )[0].dt.strftime("%Y-%m-%d").values
```
aaron-kaplan, line 233: This `where` condition is repeated in lines 661-664. Is it worth factoring out the common code into a function?

onset_maproom/maproom.py
```diff
@@ -165,7 +296,8 @@ def onset_plots(
                 color="danger",
                 dismissable=True,
             )
-            return errorFig, errorFig, alert1
+            germ_sentence = ""
+            return errorFig, errorFig, alert1, germ_sentence
```
aaron-kaplan, line 300: Didn't we decide on another PR to get rid of the alerts because the errorFigs make them redundant? I thought we had already done that; maybe that was a different situation and they're still needed here for some reason.

onset_maproom/maproom.py
```diff
@@ -276,7 +412,17 @@ def onset_plots(
         yaxis_title="Probability of Exceeding",
         xaxis_title=f"Onset Date [days since {search_start_day} {search_start_month}]",
     )
-    return onsetDate_graph, probExceed_onset, None
+    if pd.isnull(onsetDate["onset"].iloc[-1]):
+        germ_sentence = (
+            "Germinating rains have not yet occured as of "
```
aaron-kaplan, line 417: Is this text correct given the new understanding of the difference between germinating rains and onset date? Seems like this is actually about onset date, not germinating rains.

onset_maproom/maproom.py
```diff
@@ -276,7 +412,17 @@ def onset_plots(
...
+            + precip["T"][-1].dt.strftime("%B %d, %Y").values
+        )
+    else:
+        germ_sentence = (
+            "Germinating rains have occured on "
```
aaron-kaplan, line 422: As above.

onset_maproom/maproom.py
```diff
@@ -435,6 +581,7 @@ def cess_plots(
 # TODO can we make this any more redundant?
 @APP.callback(
     Output("onset_layer", "url"),
+    Input("yearly_stats_input", "value"),
```
aaron-kaplan, line 584: This callback has 11 `Input`s, but this is the only one whose name ends in `_input`. I would remove that.

Also, as I commented elsewhere, "yearly" doesn't seem appropriate now that "monit" is an option.

onset_maproom/maproom.py
```diff
@@ -501,30 +656,67 @@ def onset_tile(tz, tx, ty):
...
+        if yearly_stats_input == "pe":
+            mymap = (
+                onset_dates.onset_delta.fillna(
+                    np.timedelta64(search_days+1, 'D')
+                ) > np.timedelta64(probExcThresh1, 'D')
+            ).mean("T") * 100
+            mymap_min = 0
+            mymap_max = 100
+            mycolormap = pingrid.CORRELATION_COLORMAP
+            mymap.attrs["colormapkey"] = np.array([0, 0.1, 0.2, 0.35, 0.45, 0.45+1/256., 0.55-1/256., 0.55, 0.7, 0.85, 1])
```
aaron-kaplan, line 714: What is the significance of this choice of thresholds?

onset_maproom/maproom.py
```diff
@@ -537,14 +729,24 @@ def onset_tile(tz, tx, ty):
...
-def set_colorbar(search_start_day, search_start_month, search_days):
-    return (
-        f"Mean onset date in days past {search_start_day} {search_start_month}",
-        pingrid.to_dash_colorscale(pingrid.RAINBOW_COLORMAP),
-        int(search_days),
-        [i for i in range(0, int(search_days) + 1) if i % 10 == 0],
-    )
+def set_colorbar(search_start_day, search_start_month, search_days, yearly_stats_input):
+    if yearly_stats_input == "pe":
+        thresholds = np.array([0, 0.1, 0.2, 0.35, 0.45, 0.45+1/256., 0.55-1/256., 0.55, 0.7, 0.85, 1])
```
aaron-kaplan, line 736: Same thresholds as in line 714. Factor it out and give it a name.

pingrid.py
```diff
@@ -448,13 +452,20 @@ def parse_colormap(s: str) -> np.ndarray:
...
-    rs = np.array([vs[int(i / 256.0 * len(vs))] for i in range(0, 256)], np.uint8)
+    if thresholds is None:
+        rs = np.array([vs[int(i / 256.0 * len(vs))] for i in range(0, 256)], np.uint8)
+    else:
+        rs = np.full([256, 4], np.nan)
+        for i in range(0, thresholds.size):
+            rs[int(255.*(thresholds[i]-thresholds[0])/(thresholds[-1]-thresholds[0])), :] = np.array([vs[i]], np.uint8)
+        dfrs = pd.DataFrame(rs)
+        rs = dfrs.interpolate().astype(int).values
     return rs
```
aaron-kaplan, line 463: Is this the "colorscale business" you're saying I approved in the pycpt maproom PR? It doesn't look familiar to me.

Seems ok, but this is a note to myself to verify that it doesn't break fbfmaproom before I approve.

pingrid.py
```diff
@@ -448,13 +452,20 @@ def parse_colormap(s: str) -> np.ndarray:
...
     #     [f"{v.red:02x}{v.green:02x}{v.blue:02x}{v.alpha:02x}" for v in vs],
     # )
-    rs = np.array([vs[int(i / 256.0 * len(vs))] for i in range(0, 256)], np.uint8)
+    if thresholds is None:
+        rs = np.array([vs[int(i / 256.0 * len(vs))] for i in range(0, 256)], np.uint8)
+    else:
+        rs = np.full([256, 4], np.nan)
+        for i in range(0, thresholds.size):
+            rs[int(255.*(thresholds[i]-thresholds[0])/(thresholds[-1]-thresholds[0])), :] = np.array([vs[i]], np.uint8)
+        dfrs = pd.DataFrame(rs)
```
aaron-kaplan, line 461: We've mostly been using the convention of suffixes `_df`, `_da`, and `_ds` for DataFrame, DataArray, and DataSet. Following that convention, this would be called `rs_df`.

onset_maproom/widgets.py
```diff
@@ -50,12 +50,12 @@ def Sentence(*elems):
         groups.extend(elems[0])
 
     for i in range(start, len(elems) - (1 if tail else 0), 2):
-        assert isinstance(elems[i], str)
+        assert (isinstance(elems[i], str) or isinstance(elems[i], html.Span))
```
aaron-kaplan, line 53: @kgraaf could you comment here? I'm not sure I understand the intent of the original assertion so I don't know if it's appropriate to relax it like this.

remicousin:
> I started a review last week and never finished it. Here's what I've entered so far, will continue reviewing after I send it.
> 
> In addition to the in-line comments that are part of this review, I'll gather a summary of my other unaddressed comments here:
> 
>     * [ ]  Pin pandas=1.3
> 
>     * [ ]  Create an enactstozarr.py for Senegal, or adapt the existing one to handle both countries.
> 
>     * [ ]  Do you want to use a coarser resolution to speed it up?

enactstozarr also created Senegal dataset. This is relatively old work:
https://github.com/iridl/python-maprooms/commits/master/onset_maproom/enactstozarr.py
I can't remember that I actually tested it against the NMA nc files... I could do that now on a subset of NMA nc files.

Performance varies a lot depending where I run the app from (office, home with VPN) and just the moment... Do you expect performance boost from wherever the public version will be installed? if yes, we're all set. If no or don't know, I'll ask the IRI users/reviewers to comment on performance.

onset_maproom/layout.py
```diff
@@ -217,27 +230,29 @@ def controls_layout():
...
+                dbc.Select(
+                    id="yearly_stats_input",
+                    value="monit",
+                    # bs_size="sm",
+                    options=[
+                        {"label": "Has germinating rain occured?", "value": "monit"},
+                        {"label": "When to prepare for planting?", "value": "mean"},
+                        #as of now, xr.std doesn't know how to deal with NaT
+                        #{"label": "Climatological Standard deviation", "value": "stddev"},
+                        {"label": "How risky to plant...", "value": "pe"},
```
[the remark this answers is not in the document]
remicousin: I commented on that in the users feedback document so that they can comment. I am not a fan of "mean: when to prepare for planting." 'mean' says too much or too little. Having Maprooms speak the language of users (ie Human rather than maths) is popular when there is a clear audience identified who we know what language they speak. Then documentation in the app describes the maths definitions.
In Senegal it seems the audience is identified, in Ethiopia, not so much, that's why the question whether it applies in Ethiopia set up is a valid one. I assumed, to start with, that yes it does, but we'll see what Tufa/Amanda say

onset_maproom/config-anacim-senegal.yaml
```diff
@@ -19,12 +32,23 @@ rr_mrg_zarr_path: /data/remic/mydatafiles/ANACIM/anacim_v3/ALL_block2/rr_mrg_dai
...
 
 # Onset_and_Cessation
-onset_and_cessation_title: When to prepare land and inputs for planting?
+app_title: Planting Maproom
+onset_and_cessation_title: Planting Date Decision Support Maproom
 default_search_month: May
 default_running_days: 3
 default_min_rainy_days: 1
 ison_cess_date_hist: False
+and_cess_text: ""
```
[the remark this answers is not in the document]
remicousin: agreed for ison_cess_date_hist and and_cess_text; but who knows what, say Rwanda, would ask for a titles...
Then I created and_cess_text because I didn't manage to have code with the """ text...

onset_maproom/layout.py
```diff
@@ -169,19 +169,32 @@ def controls_layout():
...
                 """
             ),
             html.P(
                 """
-                By enabling the exploration of the history of onset
-                dates, the Maproom allows to understand the spatial
-                and temporal variability of this phenomenon and
-                therefore characterize the risk for a successful
-                agricultural campaign associated with it.
+                The default map shows whether germinating rains have occured
```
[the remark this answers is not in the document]
remicousin: In my experience, there are choices to make to accommodate first time users of the page, and returning ones. There is a conflict because first-time users need all the info about the app, whereas returning users want to easily jump to the controls (the already know what all is about). In light of this, my practice has been:

1. first thing first: describe what is seen when landing (maybe should us "map of landing page" rather "default map")
2. controls
3. detailed description of the choices and contents

Putting more of 3. in 1. is annoying for returning users by pushing back the controls. Having less of 1. is not ideal because bad users could then jump to controls, never read anything and then ask annoying questions.

But I am totally open to new strategy, possibly enabled by new techno we didn't use in Ingrid Maprooms. Like collapsable text for instance?

aaron-kaplan:
> enactstozarr also created Senegal dataset. 

Ah, ok, but then I added resampling and rechunking with hard-coded values ([bounding box and chunk structure](https://github.com/iridl/python-maprooms/blob/master/onset_maproom/enactstozarr.py#L46-L48), and RESOLUTION) for Ethiopia, not realizing that the script was also used for Senegal. So we need to move those hard-coded values to the config file.

aaron-kaplan:
> Performance varies a lot depending where I run the app from (office, home with VPN) and just the moment.

Are you running on devi/shortfin01 via VPN, or have you copied the data to your laptop?

Depending on where the bottleneck lies, it might or might not be faster in production than in your test setup. I will look into that when I test the maproom, which I haven't gotten a chance to do yet.

onset_maproom/layout.py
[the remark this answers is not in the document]
aaron-kaplan: > first thing first: describe what is seen when landing (maybe should us "map of landing page" rather "default map")

What about "first thing first: describe what is currently displayed?" The explanatory text can change depending on the value of the control.

aaron-kaplan:
When I zoom way in, I see gaps between the tiles. Most likely a bug in pingrid.py, so I'll work on that in a separate branch. In the meantime, just don't zoom too far in when you demo :-)

image

aaron-kaplan:
The graphs are very slow to load because the dataset hasn't been rechunked as we did for Ethiopia. Rechunking will speed up the graphs at the expense of making the map slower; I tried to find a reasonable compromise.

We don't necesssarily need to address this immediately.

aaron-kaplan:
There are some gaps in coverage in the climatological map that aren't there in the monitoring map. I would guess that they're caused by missing data in one or more historical years. Do you want to put in more robust nan-handling?

image

pingrid.py
remicousin: That and more. But it may be inaccurate to claim you have approved. See commits
[pingrid can now make colormaps unequally spaced](https://github.com/kjhall-iri/pycpt-remic/pull/7/commits/6f6f49ce8fe1ab190a0b69dfabe41212591c1941)
to
[the right static colorscale](https://github.com/kjhall-iri/pycpt-remic/pull/7/commits/c01311bf2a60fb18b85a886b4a84bb7822efd74b)
from here:
https://github.com/kjhall-iri/pycpt-remic/pull/7/commits

remicousin:
> There are some gaps in coverage in the climatological map that aren't there in the monitoring map. I would guess that they're caused by missing data in one or more historical years. Do you want to put in more robust nan-handling?
> 
> image

no. It's actually quite possible. Remember that onset_date  = germinating rains date + check if no following dry spells. The monitoring map shows germinating date (since by definition while monitoring, one can not know if a dry spell is coming). The climatological analyses work on onset dates, i.e. including the dry spell check. Thus... There can be germinating rains every year and no year with onset dates. In other words, the drier north of Senegal is more prone to false onset dates.

onset_maproom/config-anacim-senegal.yaml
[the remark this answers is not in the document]
remicousin: I got errors when trying to squeeze an if in the """-string to check boolean and act accordingly. Maybe f-string will solve that?

aaron-kaplan:
Your explanation for the pixels with no mean onset date makes sense, but it makes me realize that mean onset date could be a misleading/dangerous statistic to use if not accompanied by information about how often there's no onset date at all.

onset_maproom/layout.py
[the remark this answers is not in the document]
remicousin: That sounds like a new strategy enabled by new techno as we were not (much) doing dynamic text in Ingrid Maprooms. Probably need signaling that text is updating, maybe some blinking or what-have-you for the user to realize that it is.
Will make an issue out of this as I don't think I have time to do that on this PR.

onset_maproom/maproom.py
remicousin: Not feeling completely confident I implemented this in the most generalizable way... If we wanted to apply it elsewhere...

onset_maproom/maproom.py
remicousin: Text correct but data used incorrect

onset_maproom/config-anacim-senegal.yaml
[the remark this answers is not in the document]
aaron-kaplan: I don't think so. I'd have to see the example to know what you were doing wrong.

In either case, whether you write
```
'abc' + foo + 'def'
```
or 
```
f'abc{foo}def'
```
`foo` has to be an expression, not a statement. Maybe the [ternary operator](https://docs.python.org/3/reference/expressions.html#conditional-expressions) is what you were missing?

remicousin:
> Your explanation for the pixels with no mean onset date makes sense, but it makes me realize that mean onset date could be a misleading/dangerous statistic to use if not accompanied by information about how often there's no onset date at all.

Xarray's [mean](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.mean.html) is missing a minfrac argument... Could implement something instead here and make the input a maproom's form and/or a masking layer.

That also makes me think that Jim asked if we could draw stripes over areas. That could be a way here to still plot the result (with color) and give a sense of cautious, with stripes on top... But that might not work very well if to-be-strapped pixels are isolated...

onset_maproom/maproom.py
aaron-kaplan: > Not feeling completely confident I implemented this in the most generalizable way.

Maybe not, but it looks good for now.

onset_maproom/config-anacim-senegal.yaml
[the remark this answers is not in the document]
remicousin: it's the syntax for if/else showed in the link you shared that I didn't know... I was doing something like 
if boolean: expression

remicousin:
A little recap on this PR, blocking approvals/merging:

- I've just pushed something that improves the config file using testing to print or not text in layout
- @aaron-kaplan questioned whether the menu of Questions was appropriate and whether so for both Sen and Eth. That questions have been asked to IRI colleagues so I suggest we consider this a problem outside that PR. If/when colleagues comment on that, will open a new PR
- xarray's mean could use a "minfrac" optional argument. It doesn't seem to have one. Climatological maps can return values even if there is only 1 year with an onset date. I am open but my gut feeling is to take this problem out of this PR; my asking IRI colleagues what behavior they want about this; and revisit accordingly (Should not xarray folks add this functionality?)
- @aaron-kaplan asked for more info about my colorscale work in pingrid. I provided that in the corresponding conversation. Please review
- @kgraaf was asked to review changes I made to widgets

At this point this is not really time-sensitive other than getting all those changes in a hanging PR.

onset_maproom/widgets.py
aaron-kaplan: @kgraaf please comment.

onset_maproom/maproom.py
```diff
@@ -91,6 +98,92 @@ def initialize(toto):
...
+            df = pd.read_sql(
+                s,
+                conn,
+            )
+    df["the_geom"] = df["the_geom"].apply(lambda x: wkb.loads(x.tobytes()))
+    df["the_geom"] = df["the_geom"].apply(
+        lambda x: x if isinstance(x, MultiPolygon) else MultiPolygon([x])
+    )
+    shapes = df["the_geom"].apply(shapely.geometry.mapping)
+    return {"features": shapes}
```
aaron-kaplan, line 129: This function retrieves shapes from postgis in wkb form, and then uses shapely to convert from wkb to geojson, which is what leaflet expects. Shapely seems unnecessary here, because postgis can generate geojson directly. Try replacing ST_AsBinary with ST_AsGeoJSON in the query (which is in the config file) and replacing lines 124-129 with
```
return {"features": df["the_geom"]}
```

aaron-kaplan:
This is quite a large PR so I'm sure I'll have something new to say about it every time I look, but we don't have to resolve all of them before it's merged. The one thing that seems potentially important is the comment about mymap_max being undefined in one case. Rechunking for better performance also seems important, but not necessarily before merging this PR.

onset_maproom/maproom.py
```diff
@@ -501,30 +620,65 @@ def onset_tile(tz, tx, ty):
...
+            wet_spell_thresh,
+            min_wet_days,
+            dry_spell_length,
+            dry_spell_search,
+        )
+        if map_choice == "mean":
+            mymap = onset_dates.onset_delta.mean("T")
+            mymap_max = np.timedelta64(search_days, 'D')
+        if map_choice == "stddev":
+            mymap = onset_dates.onset_delta.std("T", skipna=True)
```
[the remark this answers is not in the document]
remicousin: The option is in the original Ingrid Maproom. Maybe I just need to convert the xarray from timedeltas to days as numbers and thus NaT as NaN and then apply std?

onset_maproom/maproom.py
```diff
@@ -91,6 +98,92 @@ def initialize(toto):
...
+    shapes = df["the_geom"].apply(shapely.geometry.mapping)
+    return {"features": shapes}
+
+
+@APP.callback(
+    Output("borders_adm1", "data"),
+    Input("submitLatLng","n_clicks"),
+)
+def adm1_borders(toto):
+    return adm_borders(CONFIG["shapes_adm1"])
```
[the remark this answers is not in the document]
remicousin: I will apply the changes in subseas_... once you approve them, with changes if need be

onset_maproom/maproom.py
remicousin: I did what you said but the geometries just don't show, without error information. Well there are warnings but those have been around since day 1 when I picked this code from fbfmaproom, and they are not trivial for me to interpret. I also tried to apply ST_Transform(the_geom, 4326) but that didn't make a difference.
I printed what is returned in both set ups and the prints looked very similar, except with some syntax variations that seemed harmless, but also parentheses turned brackets...

remicousin:
I applied some of the changes you also asked for in subseas_.
Going to look into the standard deviation problem now.

remicousin:
I tried to put the reading of the shapes inside layout, but I most of the time would get:
psycopg2.InterfaceError: connection already closed
Or I would get the app to load without any borders drawn. So I reverted that back. I think there was more to do anyway on that front, in particular not having the number of admin layers hard-coded. So maybe this should be taken care of together in an other PR.

onset_maproom/maproom.py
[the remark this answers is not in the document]
aaron-kaplan: OK. In the meantime, let's use ``Input("location", "pathname")` instead of `submitLatLng` for the dummy input. The Location path never changes after the initial page load, so that will prevent this callback from being called unnecessarily.

onset_maproom/maproom.py
remicousin: This is now an Issue

remicousin:
I made issues out of the 2 outstanding problems: dummy callback for admin borders and useless use of shapely.

I think that might be ready enough to merge so that @drewmresnick can resume work on this project from an updated master. It's been out there for a while so if I can merge and resolve conflicts this morning, that will likely be better than let others do that.

onset_maproom/maproom.py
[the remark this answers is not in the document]
remicousin: ok... on it now

remicousin:
rebased, regenerated linux env as conflicts are too tedious to figure out. Passed all tests, both Senegal and Ethiopia apps behaved.

aaron-kaplan (approved):

onset_maproom/widgets.py
remicousin: make it an issue?

aaron-kaplan: We have bigger fish to fry.

PR #91, opened 2022-05-09
Eliminate confusing reference to issue months by position in the list

aaron-kaplan:
Internal representations of months were 0-based month numbers (0 = January) in some places, and indexes into the issue month list (0 = whatever month is listed first in the config file) in others. This makes it consistently 0-based month numbers everywhere.

1-based everywhere would probably be better (less mental gymnastics when debugging), but that's a bigger change that I'll leave for another day.

### DIFF
--- a/fbfmaproom/fbfmaproom.py
+++ b/fbfmaproom/fbfmaproom.py
@@ -294,14 +294,14 @@ def generate_tables(
     obs_dataset_key,
     season_config,
     table_columns,
-    issue_month_idx,
+    issue_month0,
     freq,
     mode,
     geom_key,
     severity,
 ):
     basic_ds = fundamental_table_data(country_key, obs_dataset_key,
-                                      season_config, issue_month_idx,
+                                      season_config, issue_month0,
                                       freq, mode, geom_key)
     main_df, summary_df, prob_thresh = augment_table_data(basic_ds.to_dataframe(), freq)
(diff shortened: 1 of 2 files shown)


kgraaf (approved):

fbfmaproom/fbfmaproom.py
```diff
@@ -1077,16 +1074,12 @@ def download_table():
 
     country_config = CONFIG["countries"][country_key]
     season_config = country_config["seasons"][season_id]
-    try:
-        month0 = abbrev_to_month0[issue_month]
-        issue_month_idx = season_config["issue_months"].index(month0)
-    except ValueError:
-        raise pingrid.NotFoundError("No forecasts issued in {issue_month}")
+    issue_month0 = abbrev_to_month0[issue_month]
```
kgraaf, line 1077: I know this was already merged, but I'm just curious why this exception raising was removed

aaron-kaplan: The line that could have raised the exception (line 1082) has been deleted, so there's no longer anything to catch here.

Empirically, it looks like we're still properly handling the case where the user selects a forecast that hasn't been issued yet (I see the alert). That's not the case that was being handled here. This was the case where somehow the UI requested the forecast for a month that the config file doesn't list as a valid month. Now that latter case will be handled by the same logic as the former.

PR #92, opened 2022-05-10
Generalize enactstozarr.py to handle both Ethiopia and Senegal

aaron-kaplan:


onset_maproom/enactstozarr.py
```diff
@@ -10,21 +10,22 @@
...
 #Concatenate them against added time dim made up from filenames
 
 RR_MRG_PATH = Path(RR_MRG_NC_PATH)
 RR_MRG_FILE = list(sorted(RR_MRG_PATH.glob("*.nc")))
 
-RESOLUTION = .25
-
 
 def set_up_dims(xda):
+    datestr = Path(xda.encoding["source"]).name.split("_")[2]
```
remicousin, line 24: hopefully this will work for a good deal of ENACTS datasets (I mean 3rd of split by _)

onset_maproom/enactstozarr.py
```diff
@@ -43,9 +46,12 @@ def set_up_dims(xda):
     # and look into xESMF.
```
remicousin, line 46: could this be the culprit for the bands of empty data when zooming the Senegal map? I guess not because you are not doing regridding for Senegal, and you wrote the rechunked Senegal data after seeing the bands... correct?

remicousin (approved):

onset_maproom/enactstozarr.py
aaron-kaplan: Correct on both counts.

PR #93, opened 2022-05-13
Simplify DataTable formatting, eliminate worst_* columns

aaron-kaplan:
This didn't work out as well as I hoped. Apparently `null < 1` in the DataTable conditional expression language, so cells with no value get colored in the obs column. Also DataTable's number formatting didn't work, so I had to keep the forecast column as a string that we format in python before sending it. I'm sure we could find workarounds for these issues, but it doesn't seem worth it if we plan to drop DataTable soon anyway.

Sending this in case there's anything useful to port over to the supertable branch.

### DIFF
--- a/fbfmaproom/tests/test_fbfmaproom.py
+++ b/fbfmaproom/tests/test_fbfmaproom.py
@@ -22,7 +22,7 @@ def test_from_month_since_360Day():
     assert fbfmaproom.from_month_since_360Day(735.5) == DT360(2021, 4, 16)
 
 def test_generate_tables():
-    main_df, summary_df, prob_thresh = fbfmaproom.generate_tables(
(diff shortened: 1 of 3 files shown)


aaron-kaplan:
Pushed another proof-of-concept commit demonstrating that we can solve the formatting problems easily if we generate the HTML ourselves.

image

fbfmaproom/tests/test_fbfmaproom.py
```diff
@@ -22,7 +22,7 @@ def test_from_month_since_360Day():
     assert fbfmaproom.from_month_since_360Day(735.5) == DT360(2021, 4, 16)
 
 def test_generate_tables():
```
aaron-kaplan, line 24: This test probably isn't working right now.

PR #96, opened 2022-05-16
Niger new obs

aaron-kaplan:
Added several new observational datasets for Niger. In the process, I made some changes to the download script to make it a little more convenient and to support datsets that Ingrid can't provide in a single download. I also had to reorder some mathematical operations to avoid an integer overflow.

fbf-update-data.py now uses open_mfdataset, which requires dash. For the moment I'm using the enactsmaproom environment to run it, whichi is obviously not a good long-term solution. We need to have a formal description of separate dev and prod environments.

### DIFF
--- a/fbfmaproom/fbf-update-data.py
+++ b/fbfmaproom/fbf-update-data.py
@@ -11,89 +11,87 @@ base = "http://iridl.ldeo.columbia.edu"
 datasets = [
     (
         "enso",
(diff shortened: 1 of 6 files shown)


kgraaf:
So why did you add the domain name prefix to each URL instead of how it worked before?

aaron-kaplan:
> So why did you add the domain name prefix to each URL instead of how it worked before?

I added the protocol and host at the beginning, and removed the data.nc at the end, so that I can easily copy-paste between the URL bar of a browser and the editor. I found myself making those same edits again and again.

kgraaf:
Makes sense

kgraaf (approved):

aaron-kaplan:
I'll merge now to avoid future conflicts, but the new datasets aren't going into production until they've been ok'd by FIST.

PR #104, opened 2022-05-19
Multiple obs columns

aaron-kaplan:
Using a multi-select just above the table instead of buttons/menus within the table. This might not be the ideal UI, but it was quick to implement, which gives us a chance to try implementing triggers on obs before the deadline.

### DIFF
--- a/fbfmaproom/fbfmaproom.py
+++ b/fbfmaproom/fbfmaproom.py
@@ -629,7 +629,7 @@ def _(pathname):
         )
         for k, v in obs_datasets_cfg.items()
     ]
-    obs_datasets_value = next(iter(obs_datasets_cfg.keys()))
+    obs_datasets_value = [next(iter(obs_datasets_cfg.keys()))]
     return (
         f"{PFX}/custom/{c['logo']}",
         [cy, cx],
@@ -796,10 +796,9 @@ def display_prob_thresh(val):
     Input("obs_datasets", "value"),
     State("season", "value"),
 )
-def _(issue_month0, freq, mode, geom_key, pathname, severity, obs_dataset_key, season):
+def _(issue_month0, freq, mode, geom_key, pathname, severity, obs_dataset_keys, season):
     country_key = country(pathname)
     config = CONFIG["countries"][country_key]
-    obs_dataset_keys = [obs_dataset_key]
     tcs = table_columns(config["datasets"]["observations"], obs_dataset_keys)
     try:
         dft, dfs, prob_thresh = generate_tables(
--- a/fbfmaproom/fbfmaproom.py
+++ b/fbfmaproom/fbfmaproom.py
@@ -77,14 +77,14 @@ def table_columns(obs_config, obs_dataset_keys):
 
     obs_dataset_names = {k: v["label"] for k, v in obs_config.items()}
     tcs = OrderedDict()
-    tcs["year_label"] = dict(name="Year", dynamic=None, style=None,
+    tcs["year_label"] = dict(name="Year", style=None,
                              tooltip="The year whose forecast is displayed on the map")
-    tcs["enso_state"] = dict(name="ENSO State", dynamic=None,
+    tcs["enso_state"] = dict(name="ENSO State",
                              tooltip="Displays whether an El Niño, Neutral, or La Niña state occurred during the year",
                              style=lambda row: {'El Niño': 'cell-el-nino',
                                                 'La Niña': 'cell-la-nina',
                                                 'Neutral': 'cell-neutral'}.get(row['enso_state'], ""))
-    tcs["forecast"] = dict(name="Forecast, %", dynamic=None,
+    tcs["forecast"] = dict(name="Forecast, %",
                            tooltip="Displays all the historical flexible forecast for the selected issue month and location",
                            style=lambda row: "cell-severity-" + str(row['severity']) if row['worst_pnep'] == 1 else "")
 
@@ -93,7 +93,6 @@ def table_columns(obs_config, obs_dataset_keys):
             name=f"{obs_dataset_names[obs_key]} Rank",
             style=lambda row: "cell-severity-" + str(row['severity']) if row[f'worst_{obs_key}'] == 1 else "",
             tooltip=None,
-            dynamic=None
         )
 
     for obs_key in obs_dataset_keys:
@@ -108,7 +107,7 @@ def table_columns(obs_config, obs_dataset_keys):
                 return "cell-good-year"
         else:
             return ""
-    tcs["bad_year"] = dict(name="Reported Bad Years", dynamic=None,
+    tcs["bad_year"] = dict(name="Reported Bad Years",
                            style=bad_year_css,
                            tooltip="Historical drought years based on farmers' recollection")
     return tcs
--- a/fbfmaproom/fbfmaproom.py
+++ b/fbfmaproom/fbfmaproom.py
@@ -72,16 +72,8 @@ APP.title = "FBF--Maproom"
 
 APP.layout = fbflayout.app_layout()
 
-def table_columns(obs_config, obs_dataset_key):
-    def bad_year_css(row):
-        r = row['bad_year']
-        if pd.notna(r):
-            if r == "Bad":
-                return "cell-bad-year"
-            else:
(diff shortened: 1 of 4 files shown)


aaron-kaplan:
Oops, I had uncommitted changes when I pushed. Hold on...

aaron-kaplan:
> Oops, I had uncommitted changes when I pushed. Hold on...

Fixed.

fbfmaproom/fbfmaproom.py
```diff
@@ -95,9 +87,27 @@ def bad_year_css(row):
...
+    def make_obs_column(obs_key):
+        return dict(
+            name=f"{obs_dataset_names[obs_key]} Rank",
+            style=lambda row: "cell-severity-" + str(row['severity']) if row[f'worst_{obs_key}'] == 1 else "",
+            tooltip=None,
+            dynamic=None
+        )
+
+    for obs_key in obs_dataset_keys:
+        tcs[f"{obs_key}_rank"] = make_obs_column(obs_key)
```
aaron-kaplan, line 100: Interesting tidbit: in an earlier version I wrote
```
    for obs_key in obs_dataset_keys:
        tcs[f"{obs_key}_rank"] = dict(name=f"{obs_dataset_names[obs_key]} Rank",
                                      style=lambda row: "cell-severity-" + str(row['severity']) if row[f'worst_{obs_key}'] == 1 else "",
```
which resulted in each row being the same color across all the obs columns. The problem is that the lambda is a closure over the variable `obs_key`, and in that earlier version, the lambdas for all the columns end up being closures over one and the same variable (the loop variable). The version in the PR avoids that by making a new scope for `obs_key` on each iteration.

aaron-kaplan:
Just noticed a bug: Ethiopia NDVI is undefined for 2000 and earlier; on this branch, the undefined years are ranked numbers 1-18 instead of being unranked.

aaron-kaplan:
> Just noticed a bug: Ethiopia NDVI is undefined for 2000 and earlier; on this branch, the undefined years are ranked numbers 1-18 instead of being unranked.

Fixed. Also removed the unused kwarg "dynamic". @kgraaf please review again.

fbfmaproom/tests/test_fbfmaproom.py
```diff
@@ -76,21 +77,33 @@ def test_generate_tables():
...
+        worst_rain=[
             0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 1, 0, 1, 1, 0, 0, 0, 1, 0, 1, 1, 1,
             1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 1, 0
         ],
+        worst_ndvi=[
+            0.,  0.,  1.,  0.,  1.,  0.,  0.,  0.,  0.,  0.,  1.,  0.,  1.,  1.,
+            0.,  0.,  0.,  0., 0.,  0.,  1., np.nan, np.nan, np.nan, np.nan,
+            np.nan, np.nan, np.nan, np.nan, np.nan, np.nan, np.nan, np.nan,
+            np.nan, np.nan, np.nan, np.nan, np.nan, np.nan
+        ],
```
aaron-kaplan, line 106: This is kind of ugly--if the column has NaNs in it, it gets cast to float instead of int, because int doesn't have a missing value marker. I don't really want to enshrine this behavior in a test, but I decided to do the expedient thing for now because I plan to eliminate the worst_* columns soon enough.

fbfmaproom/fbfmaproom.py
kgraaf: Yet another way in which Python is a fundamentally broken language... Can't even get proper lexical scope right

aaron-kaplan: Having the loop variable persist after the loop ends is sometimes handy, particularly when you use `break`. But yeah, it's unorthodox and thus surprising.

fbfmaproom/fbfmaproom.py
```diff
@@ -72,33 +72,42 @@
...
-    tcs["bad_year"] = dict(name="Reported Bad Years", dynamic=None,
+
+    def make_obs_column(obs_key):
+        return dict(
+            name=f"{obs_dataset_names[obs_key]} Rank",
+            style=lambda row: "cell-severity-" + str(row['severity']) if row[f'worst_{obs_key}'] == 1 else "",
+            tooltip=None,
+        )
+
+    for obs_key in obs_dataset_keys:
```
kgraaf, line 98: So is the idea here that we know whether a particular column is an obs column through in-band messaging via the `_rank` suffix in the key? That probably works well enough

fbfmaproom/fbfmaproom.py
```diff
@@ -1137,6 +1166,9 @@ def download_table():
     df["time"] = time
     df["bad_year"] = df["bad_year"].astype("float") # to acommodate NaN as missing value indicator
 
+    # rename the obs column "obs" for backwards compatibility
+    df = df.rename(columns={obs_dataset_key: "obs"})
```
kgraaf, line 1170: backwards compatibility with what?

fbfmaproom/fbfmaproom.py
```diff
@@ -72,33 +72,42 @@
...
                            style=lambda row: "cell-severity-" + str(row['severity']) if row['worst_pnep'] == 1 else "")
-    tcs["obs_rank"] = dict(name=f"{obs_dataset_names[obs_dataset_key]} Rank",
-                           style=lambda row: "cell-severity-" + str(row['severity']) if row['worst_obs'] == 1 else "",
-                           tooltip=None, dynamic=None)
-    tcs["bad_year"] = dict(name="Reported Bad Years", dynamic=None,
+
+    def make_obs_column(obs_key):
+        return dict(
+            name=f"{obs_dataset_names[obs_key]} Rank",
+            style=lambda row: "cell-severity-" + str(row['severity']) if row[f'worst_{obs_key}'] == 1 else "",
```
kgraaf, line 94: Side note: I am pleased with how the fbftable component has worked given that, since we're using these lambdas to control the style logic, we didn't actually have to modify the component itself. I guess that's a sign the abstraction actually worked.

pingrid.py
```diff
@@ -540,9 +540,13 @@ def average_over(ds, s, lon_name="lon", lat_name="lat", all_touched=False):
     r = r0 * np.cos(np.deg2rad(ds[lat_name]))
 
     norm = r / r.sum([lat_name, lon_name])
-    res = (ds * norm).sum([lat_name, lon_name], skipna=True)
+    res = (ds * norm).sum([lat_name, lon_name], skipna=True, min_count=1)
```
kgraaf, line 543: what is the significance of this `min_count`?

fbfmaproom/fbfmaproom.py
aaron-kaplan: No, the obs columns are identified by `obs_dataset_keys`. We never scan through the columns and guess which ones are obs.

pingrid.py
aaron-kaplan: This fixed the bug where years with no NDVI score were being counted as having an NDVI of zero instead of missing. If da is all NaNs, `da.sum(skipna=True)` returns zero whereas `da.sum(skipna=True, min_count=1)` returns NaN unless there's at least one non-missing value.

fbfmaproom/fbfmaproom.py
kgraaf: Oh ok I see. That's probably fine. Do you think it would make more sense to somehow attach that to the table_columns data structure, or not worth it?

aaron-kaplan: To what end?

fbfmaproom/fbfmaproom.py
aaron-kaplan: Existing clients of this HTML API, namely Max's R Shiny app.

fbfmaproom/fbfmaproom.py
kgraaf: My rationale was so the extra parameter would not need to be passed, but unnecessary as this likely temporary ultimately (once we eliminate the forecast-obs distinction)

kgraaf (approved):

PR #105, opened 2022-05-23
Display underlying obs values instead of ranks

aaron-kaplan:
@kgraaf Dan would like to share this with partners by tommorrow, so please review promptly.

In a separate branch I'm working on per-column formatting rules, but that will take more time. For now, we have a single formatting function that handles both integer and timedelta columns. It doesn't cover all possible cases, just the ones we need right now.

I made an [inventory](https://docs.google.com/document/d/1UOcAsQkA9y99eIa_um12K8gRD7BKIVtBSgnXa63W75g/edit#) of the obs datasets for all the countries and confirmed that they all have correct units metadata. After sending this, I'm going to work on exposing the units in the table heading, and will push another commit if that's successful. We should deploy this by tomorrow with or without that addition.

Please review commit-by-commit and see individual commit comments.

### DIFF
--- a/fbfmaproom/fbf-update-data.py
+++ b/fbfmaproom/fbf-update-data.py
@@ -67,19 +67,19 @@ datasets = [
     ),
     (
         'niger/chirps-precip-jun',
-        'https://iridl.ldeo.columbia.edu/SOURCES/.UCSB/.CHIRPS/.v2p0/.monthly/.global/.precipitation/X/(0)/(16)/RANGE/Y/(11)/(24)/RANGE/T/(Jun)/seasonalAverage/T//pointwidth/0/def/2/shiftGRID/c%3A/1//units//months/def/%3Ac/mul/'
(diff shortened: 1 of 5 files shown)


kgraaf:
All looks good to me. Should the column header state the unit though?

kgraaf:
(Deleted erroneous comment about units in header, missed where you said this would be a follow up commit)

All looks good to me! I studied this pretty carefully and don't really see anything I would change. I will look it over once more and then approve

kgraaf (approved):

aaron-kaplan:
Well, I got the units in, but there are some problematic cases. Onset date and dry spells have proper unit metadata ("days") in the zarr file. When we open that with xarray, xarray converts the numbers to np.timedelta64 and removes the `units` attribute, so it's not there to be displayed. Furthermore, I think the units for onset date should be displayed as "days since May 1", but dry spell length should be just "days." The way I've accommodated these cases is very hacky for now. In my subsequent branch where I'm doing per-column formatting rules, I'll see if this can be done more cleanly. If not, I might disable xarray's interpretation of CF metadata for timedeltas.

@kgraaf please look again.

PR #107, opened 2022-05-25
Ranked bad years, selectable bad years column

aaron-kaplan:


aaron-kaplan:
@kgraaf this doesn't introduce any new functionality yet, but it's already quite large, and it's currently in a working state, so I think we should try to get it merged before it gets so big it's unreviewable. When we talk in the morning we can discuss relative priority.

The commit history is clean and in small, digestible steps, so I highly recommend reading it one commit at a time.

PR #123, opened 2022-06-18
Ethiopia dataset updates

aaron-kaplan:


aaron-kaplan:
Before merging, check out Rémi's changes in /data/aaron and change the update script to read from there.

PR #133, opened 2022-07-14
Make table fill available vertical space

aaron-kaplan:
Makes the table fill the available space, rather than using a fixed height of 700px. The impetus for working on this was a [complaint](https://trello.com/c/BnvCHqFf/29-remove-not-official-govt-maproom) about the "Not an official government maproom" disclaimer appearing on top of the table under certain circumstances, but having the table stretch vertically is something I've wished for independently for a long time.

I recommend reading one commit at a time.

### DIFF
--- a/fbfmaproom/assets/style.css
+++ b/fbfmaproom/assets/style.css
@@ -50,8 +50,6 @@ table.supertable {
     border-spacing: 0;
     width: 100%;
     table-layout: fixed;
-    overflow-y: auto;
-    overflow-x: hidden;
 
     border-style: solid;
     border-width: 1px;
@@ -63,7 +61,7 @@ table.supertable {
     font-family: monospace;
 
     overflow: auto;
-    height: 700px;
+    height: 100%;
     display: block;
 }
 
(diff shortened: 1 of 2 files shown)


aaron-kaplan:
image
image

aaron-kaplan:
There were some changes in my first version that were leftovers from experimentation and were no longer necessary. I've rewritten the branch to change only the things that need to be changed.

fbfmaproom/assets/style.css
```diff
@@ -50,8 +50,6 @@ table.supertable {
     border-spacing: 0;
     width: 100%;
     table-layout: fixed;
-    overflow-y: auto;
-    overflow-x: hidden;
```
aaron-kaplan, line 54: I don't think these ever had any effect, as they're overwritten by line 63.

kgraaf: Yes. These were left over from some of my experiments, as far as I know they shouldn't do anything

kgraaf (approved):

PR #134, opened 2022-07-15
Support triggers on any column, and potentially multiple forecasts

aaron-kaplan:
Implements https://trello.com/c/2ieEQPol/32-ability-to-trigger-on-an-obs-column-multiple-forecasts . Also including a fix for https://trello.com/c/iCRBXlvW/43-default-year-for-trigger-year because it would have conflicted with this if I did it on a separate branch.

This is built on top of PR #133, which is useful and deployable on its own, so please review that one first.

I suggest reviewing commit by commit.

### DIFF
--- a/fbfmaproom/assets/style.css
+++ b/fbfmaproom/assets/style.css
@@ -92,6 +92,15 @@ table.supertable tr th:first-child {
     color: rgb(67, 104, 176);
 }
 
+table.supertable thead tr:first-child th {
+    background-color: rgb(241, 241, 241);
+    border-right-style: none;
+}
+
+table.supertable thead tr:first-child th:last-child {
+    border-right-style: solid;
+}
+
 table.supertable thead tr:last-child th {
     background-color: rgb(241, 241, 241);
     color: black;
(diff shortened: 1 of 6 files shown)


aaron-kaplan:
@jturmelle I removed the `found` field from the `trigger_check` response, and added tests to cement the current behavior of returning status 404 if the requested data aren't available.

fbfmaproom/assets/style.css
```diff
@@ -92,6 +92,15 @@ table.supertable tr th:first-child {
     color: rgb(67, 104, 176);
 }
 
+table.supertable thead tr:first-child th {
+    background-color: rgb(241, 241, 241);
+    border-right-style: none;
+}
+
```
kgraaf, line 99: did you make sure a check these changes don't produce any visual glitches? they can be subtle and surprising

aaron-kaplan: I did find and fix some glitches, but it would be good to get more eyes on it. It's running on http://shortfin01.iri.columbia.edu:9999/fbfmaproom/niger

kgraaf (approved):

aaron-kaplan:
Not merging until the Gantt tool is ready to make the switch.

PR #136, opened 2022-07-19
place onset_maproom under enacts_maprooms

kgraaf:
mirror structure of old data library


aaron-kaplan:
This doesn't work, does it? The pingrid.py symlink is broken.

kgraaf:
> This doesn't work, does it? The pingrid.py symlink is broken.

You're right, oversight on my part. This should be fixed, let's merge this as soon as we can so I can the rest of my stuff if that's ok

PR #137, opened 2022-07-20
New Madagascar forecast

aaron-kaplan:
The new forecast covers all of Madagascar, for for now FBF is still using only the four southernmost regions.


aaron-kaplan:
I made a mistake in merging this--the new forecast hasn't been approved for release to the public server yet. /data/aaron/fbf-candidate contains the new forecast, but /data/aaron/fbf still has the old one for now. I will leave it like that for a bit; if we don't get approval by the end of the day today I'll revert this change to get master back in sync with the state of the production server.

PR #138, opened 2022-07-22
updated ethiopia ndvi datasets to correct months

kgraaf:


fbfmaproom/fbf-update-data.py
```diff
@@ -104,8 +104,16 @@
         "http://iridl.ldeo.columbia.edu/SOURCES/.USGS/.LandDAAC/.MODIS/.version_006/.EAF/.NDVI/X/39.875/.25/47.875/GRID/Y/3.625/.25/10.875/GRID/T/(Mar-May)/seasonalAverage/",
     ),
     (
-        "ethiopia/ndvi-prev-seas-ond",
-        "http://iridl.ldeo.columbia.edu/SOURCES/.USGS/.LandDAAC/.MODIS/.version_006/.EAF/.NDVI/X/39.875/.25/47.875/GRID/Y/3.625/.25/10.875/GRID/T/(Mar-May)/seasonalAverage/T/7/shiftGRID/",
+        "ethiopia/ndvi-jun-ond",
+        "http://iridl.ldeo.columbia.edu/SOURCES/.USGS/.LandDAAC/.MODIS/.version_006/.EAF/.NDVI/X/39.875/.25/47.875/GRID/Y/3.625/.25/10.875/GRID/T/(Jun)/seasonalAverage/T/5/shiftGRID/",
```
aaron-kaplan, line 108: Please update the spatial domain to match ethiopia/rain-mam. We need to do this eventually for all the obs datasets, but for now it's sufficient to do it for the ones you're adding. This is necessary because one of the woredas extends beyond the border of the Somali region, plus we intend eventually to extend the application to more regions.

You will need to rm -r the existing .nc file before running the script again, otherwise it just compares the last-modified date of the existing file to that of the datset, and skips the download. (Maybe we should add a checksum of the URL to the filename or something like that.)

aaron-kaplan:
You didn't mean to commit local.yaml, did you? It contains sensitive information. Please edit the branch history to remove it, and force push.

kgraaf:
> You didn't mean to commit local.yaml, did you? It contains sensitive information. Please edit the branch history to remove it, and force push.

Yes. This was a mistake. I'll update the spaital domain

fbfmaproom/fbf-update-data.py
aaron-kaplan: I believe this is the request that failed with the old spatial domain. It's currently passing; hopefully that means you've already downloaded the new dataset with the expanded spatial domain. If not, let me know so I can dig deeper. I don't want to have a regression on this bug I just fixed a couple days ago.

http://devi.iri.columbia.edu:8079/fbfmaproom/ethiopia-ond/export?&mode=2&region=(ET05,ET0511,ET051103)&season=season2&issue_month0=6&freq=30&predictor=pnep&predictand=rain

aaron-kaplan: Whoops, that's for rain, not ndvi. Constructing a new test URL...

kgraaf: I am doing this right now. This needs updated MAM monthly ndvi datasets too, right?

aaron-kaplan: I think this would be it. Should pass once you've downloaded the new datasets, or else I still have something wrong.

http://devi.iri.columbia.edu:8079/fbfmaproom/ethiopia-ond/export?&mode=2&region=(ET05,ET0511,ET051103)&season=season2&issue_month0=6&freq=30&predictor=pnep&predictand=ndvi-jul

aaron-kaplan: > This needs updated MAM monthly ndvi datasets too, right?

Not sure what you meant by "this." The test URL is for ethiopia-ond, so there's no dependence on the MAM datasets. We should add monthly NDVI variables for MAM, but that's less urgent than what's currently in this PR. Max is doing a workshop with the partner Monday morning, and we need to get this into the OND page by then.

kgraaf: Updated the spatial domain, waiting for files to download

aaron-kaplan (approved):

PR #141, opened 2022-07-22
Draw border under last table cell

aaron-kaplan:
This rule was here to avoid a double-wide bottom border, since there's also a border around the whole table. But now that the table stretches to fill the available space, sometimes the last row doesn't reach the bottom of the box, and then it looks weird:

image

With this fix and a tall window:

image

With this fix and a short window (you can see the double-wide bottom border):

image

### DIFF
--- a/fbfmaproom/assets/style.css
+++ b/fbfmaproom/assets/style.css
@@ -101,10 +101,6 @@ table.supertable tbody {
     background-color: rgb(248, 248, 248);
 }
 
-table.supertable tr:last-child td {
-    border-bottom: 0px;
-}
-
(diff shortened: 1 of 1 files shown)


aaron-kaplan:
A better solution might be to have the div that contains the table stretch to at most the full height, but not more than needed to fit the content. I tried that and didn't get it to work, so I fell back to this solution, which is not quite right but better than the status quo.

kgraaf:
We can probably revisit this in the future. I have some ideas on how this issue could be fixed but it's not worth devoting the time to

kgraaf (approved):

PR #142, opened 2022-07-22
Calculate spatial resolution before spatial subsetting

aaron-kaplan:
in case the subset is only 1 pixel wide/tall.

This was the root cause of https://trello.com/c/1wuQcqvi/46-error-during-ethiopia-admin-3-download . If I'd realized it was going to be this easy to fix, I would have done this instead of extending the spatial domain of the obs variables. :-(

### DIFF
--- a/pingrid.py
+++ b/pingrid.py
@@ -519,11 +519,8 @@ def trim_to_bbox(ds, s, lon_name="lon", lat_name="lat"):
     )
 
 
-def average_over(ds, s, lon_name="lon", lat_name="lat", all_touched=False):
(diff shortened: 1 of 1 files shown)


kgraaf:
Why does calculating the resolution outside of the function prevent the underflow (I guess you could call it) issue? Isn't it the exact same code?

aaron-kaplan:
What's important is that the resolution calculation happens prior to the call to `trim_to_bbox`, which replaces a large `ds` with a smaller spatial subset.

aaron-kaplan:
@kgraaf we discussed this in the call yesterday. Please approve.

kgraaf (approved):

PR #147, opened 2022-07-26
Rechunk for faster tiles

aaron-kaplan:
This significantly reduces the memory used by the Niger maproom, and makes Niger tiles load in about 500 ms compared to 1600 with the original chunking. It will probably help with the other countries too, but Niger was the biggest offender, since it has the biggest dataset (area x resolution).

Also including a couple small, unrelated changes to things I noticed while investingating the problem.

BTW after this change, the majority of the time used by the tile endpoint is spent querying the db to get the clipping shape. I have ideas about how to optimize that, which I will try in another branch.

### DIFF
--- a/fbfmaproom/fbf-update-data.py
+++ b/fbfmaproom/fbf-update-data.py
@@ -262,6 +262,8 @@ for dataset in url_datasets:
         # TODO do this in Ingrid
         if 'Y' in ds and ds['Y'][0] > ds['Y'][1]:
             ds = ds.reindex(Y=ds['Y'][::-1])
+        if 'P' in ds:
+            ds = ds.chunk({'P': 1})
         if os.path.exists(zarrpath):
             shutil.rmtree(zarrpath)
         ds.to_zarr(zarrpath)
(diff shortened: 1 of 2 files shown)


fbfmaproom/fbf-update-data.py
```diff
@@ -262,6 +262,8 @@
         # TODO do this in Ingrid
         if 'Y' in ds and ds['Y'][0] > ds['Y'][1]:
             ds = ds.reindex(Y=ds['Y'][::-1])
+        if 'P' in ds:
+            ds = ds.chunk({'P': 1})
```
aaron-kaplan, line 266: Original chunking is a single chunk containing the entire variable. New chunking is one chunk per value of P (percentile).

aaron-kaplan:
I regenerated all the zarrs in /data/aaron/fbf-candidate, so you can try it out there. Compare the load time of

http://shortfin01.iri.columbia.edu:9999/fbfmaproom/niger

to

http://iridl.ldeo.columbia.edu/fbfmaproom2/niger

aaron-kaplan:
@kgraaf please review. It's a two-line change.

kgraaf (approved):

kgraaf:
@aaron-kaplan sorry about that, I thought I had, think I was on a bad connection that didn't go through

PR #148, opened 2022-07-27
Revert Ethiopia spatial domain change

aaron-kaplan:
Now that PR #142 is merged, we no longer need the hack of extending the spatial domain, and the hack caused reproducibility issues because it filled in some pixels that should have been included in the original but weren't.

Already applied in production.


kgraaf (approved):

kgraaf:
Do we/I need to undo the change to the spatial review that's live on the prod now?

aaron-kaplan:
> Do we/I need to undo the change to the spatial domain change that's live on prod now?

I already did, after discussion with Max.

kgraaf:
Great

PR #149, opened 2022-07-27
Short circuit tile generation for tiles outside of dataset 

aaron-kaplan:
We've been spending 400ms on every tile, including tiles where there's no data, because fetching the clipping shape from the db is strangely expensive. (I wonder if that's because the shapes are in a "foreign table.") In this PR we defer loading the shape until we know we need it, getting the time for an empty tile down to 60ms.


aaron-kaplan:
@kgraaf this is waiting for review.

aaron-kaplan:
@kgraaf this is still waiting for review.

kgraaf (approved):

PR #152, opened 2022-07-28
No flash of western Europe on initial page load

aaron-kaplan:
See comment in the code. This annoying behavior is [specific to dash-leaflet](https://github.com/thedirtyfew/dash-leaflet/blob/master/src/lib/components/Map.react.js#L61), it's not leaflet's fault.


kgraaf:
I suppose we have to know retire the ever-useful acrony FOWE

kgraaf (approved):

PR #156, opened 2022-08-05
Drewr/subs data

drewmresnick:
The current state of things is this:

I wrote a new function, `combine_cptdataset()` which opens all of the datasets matching a specific file pattern (see config file for the file pattern variables). The function then strips information from the file names and uses the start date and lead time info to combine all of the datasets into a single dataframe which can then be indexed using xarray indexing. 

I added S,L inputs to be used by the callbacks, which looks like it should work okay.

Where I have left off is that for the tile generation `fcst_tiles()` on line 396, I will need to figure out how to update this dynamically using S,L. Since to me it does not look like a callback, I do not know how to give it an input like the regular callbacks, since it would need to be updated according to what the user chooses for the S,L in the navbar.


aaron-kaplan:
> Where I have left off is that for the tile generation fcst_tiles() on line 396, I will need to figure out how to update this dynamically using S,L. Since to me it does not look like a callback, I do not know how to give it an input like the regular callbacks

There are two steps. First `fcst_tile_url_callback`, which is a dash callback, tells the leaflet component what URL to use for the tiles; then leaflet makes tile requests, which are handled by `fcst_tiles`, which as you observed is not a dash callback. To add S and L to the tile parameters, first you need to add them as `Input`s of `fcst_tile_url_callback`, and have it add them to the tile URL. Then modify `fcst_tiles` to make it use those (add them to the path in the `@SERVER.route` decoration, and to the function's argument list).

If you need more help, let's talk.

remicousin:
Please let know when we should go ahead and review.

drewmresnick:
> Please let know when we should go ahead and review.

@remicousin Yes I will let you know when it is ready.

drewmresnick:
> > Where I have left off is that for the tile generation fcst_tiles() on line 396, I will need to figure out how to update this dynamically using S,L. Since to me it does not look like a callback, I do not know how to give it an input like the regular callbacks
> 
> There are two steps. First `fcst_tile_url_callback`, which is a dash callback, tells the leaflet component what URL to use for the tiles; then leaflet makes tile requests, which are handled by `fcst_tiles`, which as you observed is not a dash callback. To add S and L to the tile parameters, first you need to add them as `Input`s of `fcst_tile_url_callback`, and have it add them to the tile URL. Then modify `fcst_tiles` to make it use those (add them to the path in the `@SERVER.route` decoration, and to the function's argument list).
> 
> If you need more help, let's talk.

Is there some specific order in which the lead time and start date should be added to the url? And do they need to be any specific data type?

drewmresnick:
@aaron-kaplan I have not been able to get the tiles to update correctly, and my main issue at the moment is that I don't get any warning / error messages printed for me to debug. Do you know how to get them printed if the error is somewhere in the server routing code (I have narrowed an error to somewhere in lines 442-456)

aaron-kaplan:
Set the mode to prod instead of debug in the config file. In debug mode it tries to display error messages in the browser instead of on the console, but sometimes it fails to do that and the messages don't get displayed at all.

subseas_flex_fcst/maproom.py
```diff
@@ -453,8 +453,7 @@ def fcst_tiles(tz, tx, ty, proba, variable, percentile, threshold, startDate,lea
         coords = fcst_mu.rename({"X": "lon", "Y": "lat"}).coords,
         dims = fcst_mu.rename({"X": "lon", "Y": "lat"}).dims
     # pingrid.tile wants 2D data
-    ).squeeze("T")
-    print("test2")
+    )#.squeeze("T")
```
drewmresnick, line 456: Perhaps because of how the data is combined, the need to remove the T dimension is not required? Not sure exactly but it seems to work okay without it....

drewmresnick:
Okay @aaron-kaplan @remicousin I have a working version of this branch now, the tiles seem to be updating correctly though a bit slowly.

subseas_flex_fcst/maproom.py
```diff
@@ -314,42 +375,49 @@ def draw_colorbar(proba, variable, percentile):
...
-
+    fcst_mu, fcst_var, dofVar, obs, hcst = read_cptdataset(leadTime, startDate, y_transform=CONFIG["y_transform"])
     # Get Issue date and Target season
-    # Hard coded for now as I am not sure how we are going to deal with time
-    issue_date = pd.to_datetime(["2022-04-01"])
-    target_start = pd.to_datetime(["2022-04-02"])
-    target_end = pd.to_datetime(["2022-04-08"])
+    issue_date_td = pd.to_datetime(startDate)
+    if leadTime == "Week 1":
+        lead_time = 1
```
aaron-kaplan, line 407: Are you sure this should be 1 (the day after the issue date) and not 0?

aaron-kaplan: I'm asking a question about the meaning of the original data, not about how we want to represent it internally (we're having that latter discussion on an email thread).

subseas_flex_fcst/maproom.py
```diff
@@ -45,26 +48,75 @@
...
+
+        leadTimeWk = [x for x in fileCompList if x.startswith('wk')]
+
+        dateFormat = re.compile(".*-.*-.*")
+        startDateStr = list(filter(dateFormat.match,fileCompList))
+        startDate = startDateStr[0]
+        #startDate = datetime.strptime(startDateStr[0],
+        #'%b-%d-%Y').date()
+        if leadTimeWk[0] == 'wk1':
+            leadTimeDate = 'Week 1'#'Apr 2' #should it be made into dates or kept as wk number?
```
aaron-kaplan, line 67: Why don't you just use the numeric lead times as the values of L, rather than using strings here and then converting them to numbers later?

subseas_flex_fcst/maproom.py
```diff
@@ -45,26 +48,75 @@
...
+            ds = ds.assign(dof=dofVar)
+            #dofDict = {'dof':[dof]}
+            #ds =ds.assign_coords(dof=("dof",[dof]))
+            #ds = ds.expand_dims(dofDict)
+        if len(ds['T']) == 1:
+            startDate
+            ds['T'] = [startDate] #for obs data there are many dates
+        ds = ds.expand_dims(leadTimeDict)
+        dataDic[idx] = ds
+    dataCombine = xr.combine_by_coords([dataDic[x] for x in dataDic],combine_attrs="drop_conflicts")
```
aaron-kaplan, line 88: Is drop_conflicts necessary? It seems like it could cover up mistakes that we would rather know about so we can fix them.

subseas_flex_fcst/maproom.py
```diff
@@ -45,26 +48,75 @@
...
+
+        ds = cptio.open_cptdataset(filesNameList[idx])
+        if dof == True:
+            dofVar = float(ds["tp_pred_err_var"].attrs["dof"])
+            ds = ds.assign(dof=dofVar)
+            #dofDict = {'dof':[dof]}
+            #ds =ds.assign_coords(dof=("dof",[dof]))
+            #ds = ds.expand_dims(dofDict)
+        if len(ds['T']) == 1:
+            startDate
```
aaron-kaplan, line 84: ??

subseas_flex_fcst/maproom.py
```diff
@@ -45,26 +48,75 @@
...
         hcst_name = list(hcst.data_vars)[0]
         hcst = hcst[hcst_name]
     else:
         hcst=None
-    return fcst_mu, fcst_var, obs, hcst
+    return fcst_mu, fcst_var, dofVar, obs, hcst
 
+#Not sure if i should call the full data once outside of callbacks like this
+#So that it only opens the datasets once instead of every time the callback is fired?
+#fullDS = combine_cptdataset(DATA_PATH,CONFIG["forecast_mu_filePattern"],y_transform=False)
```
aaron-kaplan, line 119: Remove comment. As we discussed on zoom, you shouldn't do that.

subseas_flex_fcst/maproom.py
```diff
@@ -45,26 +48,75 @@
 
 APP.layout = layout.app_layout
 
-
-def read_cptdataset(y_transform=False):
-
-    fcst_mu = cptio.open_cptdataset(Path(DATA_PATH, Path(CONFIG["forecast_mu_file"])))
+#function to open all datasets and then combine them
+def combine_cptdataset(dataDir,filePattern,dof=False,y_transform=False):
```
aaron-kaplan, line 52: `y_transform` argument is never used?

subseas_flex_fcst/maproom.py
```diff
@@ -45,26 +48,75 @@
...
     obs = obs[obs_name]
     if y_transform:
-        hcst = cptio.open_cptdataset(Path(DATA_PATH, Path(CONFIG["hcst_file"])))
+        #hcst = cptio.open_cptdataset(Path(DATA_PATH, Path(CONFIG["hcst_file"])))
+        hcst = combine_cptdataset(DATA_PATH,CONFIG["hcst_filePattern"],y_transform=False)
+        hcst = hcst.sel(L=leadTime)
         hcst_name = list(hcst.data_vars)[0]
         hcst = hcst[hcst_name]
     else:
         hcst=None
```
aaron-kaplan, line 114: I don't understand how `y_transform` is being used here. Probably a question for @remicousin since it seems to have been this way in the previous version already. If `y_transform` is false then we don't read the hindcast data? That was probably only meant to be temporary.

subseas_flex_fcst/maproom.py
drewmresnick: The numbers I use here are based off the email Remi originally sent regarding S,L time and based off his original code from when he was only using a single data file where he specified:

issue_date = 2022-04-01 (start date)
target_start = 2022-04-02
target_end = 2022-04-08

with a lead time of "week 1" which I assumed would be adding a single day since target_start = issue_date +1

subseas_flex_fcst/maproom.py
drewmresnick: I did this based off of how the data was given to me; I figured we would want to keep it as close to the way the original data was expressed. But I can change this since, based off the way to email thread is going, it seems like we might want to express S,L in a more mathematically easeful way within the code over trying to stick to user convenience / similarity to the metadata?

subseas_flex_fcst/maproom.py
drewmresnick: Basically the issue was that, the dof (degree of freedom) was only expressed as a variable attribute within each data file, and each file might have a different dof associated with the data. So when I was combining the data sets, I was running into a problem because an attribute could not have more than one value. So this was the way I got around that (I do add the dof as a variable in the ds before they are removed as an attribute in line 77-79 to be used later)

subseas_flex_fcst/maproom.py
drewmresnick: Yes, this I cannot speak on. Will have to wait for @remicousin response on that.

subseas_flex_fcst/config-sample.yaml
```diff
@@ -12,10 +12,15 @@ loglevel: info
 # Forecast options
 y_transform: true
 forecast_mu_file: CFSv2_SubXPRCP_CCAFCST_mu_Apr_Apr-1-2022_wk1.txt
```
aaron-kaplan, line 14: Remove these old config options that are no longer used.

subseas_flex_fcst/maproom.py
```diff
@@ -45,26 +48,75 @@
...
+    dataCombine = xr.combine_by_coords([dataDic[x] for x in dataDic],combine_attrs="drop_conflicts")
+    return dataCombine
+
+def read_cptdataset(leadTime, startDate, y_transform=False): #add leadTime and startDate as inputs
+    fcst_mu = combine_cptdataset(DATA_PATH,CONFIG["forecast_mu_filePattern"],y_transform=False)
+    fcst_mu = fcst_mu.sel(L=leadTime, T=startDate)
     fcst_mu_name = list(fcst_mu.data_vars)[0]
     fcst_mu = fcst_mu[fcst_mu_name]
-    fcst_var = cptio.open_cptdataset(Path(DATA_PATH, Path(CONFIG["forecast_var_file"])))
+    #fcst_var = cptio.open_cptdataset(Path(DATA_PATH, Path(CONFIG["forecast_var_file"])))
```
aaron-kaplan, line 96: Remove commented old versions here and below.

subseas_flex_fcst/maproom.py
drewmresnick: @remicousin

subseas_flex_fcst/maproom.py
```diff
@@ -45,26 +48,75 @@
...
+        leadTimeDict = {'L':[leadTimeDate]}
+
+        ds = cptio.open_cptdataset(filesNameList[idx])
+        if dof == True:
+            dofVar = float(ds["tp_pred_err_var"].attrs["dof"])
+            ds = ds.assign(dof=dofVar)
+            #dofDict = {'dof':[dof]}
+            #ds =ds.assign_coords(dof=("dof",[dof]))
+            #ds = ds.expand_dims(dofDict)
+        if len(ds['T']) == 1:
```
remicousin, line 78: More needs to happen here: obs and hcst must be dealt with differently from mu and var. mu and var have one time dimension, whereas obs and hcst have two. mu and var time dimension is S, the date when the forecast is issued and that is organized by file and retrievable in the filename pattern. hcst and obs have two dimensions. The first one is the same S; the second one is a 78-point long time dimension which are the years of training. They are stored within each S-dependent file as a dimension of the xarray.

So we need to discriminate those 2 time dimensions. Right now they are mixed up. Good news is that the hcst/obs time dimension is already one provided when reading with open_cpt_dataset, so we need not create it of whave have you. Bad news that even though it's there, its values are crap. Good news is that it's crap the same way for hcst and obs, and hcst and obs only interact together before this time dimension is reduced, so it's fine to keep it as crap as it is.

So I would suggest (unless we want to clean up all the crappiness):

- for mu and var, after having read one file, remove the dummy 1-point T dimension that is there and then combine against S
- for obs and hcst: Don't touch their T dimension and combine also against S.

remicousin: And also, back to my more overarching comment, if we choose not to combine the data by S and L, but just use S/L to scrawl data files and make menus, then that problem would vanished on its own.

subseas_flex_fcst/config-sample.yaml
remicousin: That very one is used by layout for its initialization. But we can probably replace what is in the layout with something that would just pick the first mu file of the list. Just need to read one.

drewmresnick:
A few thoughts / observations I wanted to put here:

1. I think based off of the email conversation so far, it sounds like I could remove the lead time definition from the distributions plots titles, but then I will want to describe in the maproom description what the lead time is so the user knows what is going on when they select from the lead time dropdown (tbd how we handle the lead time in the actual code but that is a separate thing).

2. I noticed a bug with the layers selection where, if you deselect the tiles layer and then try to add it back on in the layers dropdown, the tile does not reappear. Something to look in to (and I wonder if this occurs in other maprooms as well?).

3. The maproom has slowed down quite a bit since applying these changes- not sure if it is now slow enough to warrant exploring options to make it faster or if we think this speed is somewhat normal for our maprooms, given the tiles generation.

subseas_flex_fcst/config-sample.yaml
```diff
@@ -12,10 +12,15 @@ loglevel: info
...
+forecast_mu_filePattern: CFSv2_SubXPRCP_CCAFCST_mu_*.txt
 forecast_var_file: CFSv2_SubXPRCP_CCAFCST_var_Apr_Apr-1-2022_wk1.txt
+forecast_var_filePattern: CFSv2_SubXPRCP_CCAFCST_var_*.txt
 obs_file: CFSv2_SubXPRCP_CCAFCST_obs_Apr_Apr-1-2022_wk1.txt
+obs_filePattern: CFSv2_SubXPRCP_CCAFCST_obs_*.txt
 hcst_file: CFSv2_SubXPRCP_CCAFCST_xvPr_Apr_Apr-1-2022_wk1.txt
+hcst_filePattern: CFSv2_SubXPRCP_CCAFCST_xvPr_*.txt
 variable: Precipitation Anomaly
+tp_length: 6
```
remicousin, line 19: This is currently not used. I feel it will be at some point, but then it should be 7 rather than 6. More intuitive for a forecast predicting a week.

subseas_flex_fcst/maproom.py
```diff
@@ -166,7 +227,7 @@ def local_plots(n_clicks, click_lat_lng, latitude, longitude):
 
     # Forecast CDF
     fcst_q, fcst_mu = xr.broadcast(quantiles, fcst_mu)
-    fcst_dof = int(fcst_var.attrs["dof"])
+    fcst_dof = int(dofVar) #int(fcst_var.attrs["dof"])
```
remicousin, line 213: if you make dofVar int from the get-go, then you just need fcst_dof = dofVar. And you could use "fcst_dof" isntead of "dofVar" when calling read_cptdataset and you'd have nothing to do here.

subseas_flex_fcst/maproom.py
```diff
@@ -45,26 +48,75 @@
...
+            leadTimeDate = 'Week 2'#'Apr 9'
+        elif leadTimeWk[0] == 'wk3':
+            leadTimeDate = 'Week 3'#'Apr 16'
+        elif leadTimeWk[0] == 'wk4':
+            leadTimeDate = 'Week 4'#'Apr 23'
+        leadTimeDict = {'L':[leadTimeDate]}
+
+        ds = cptio.open_cptdataset(filesNameList[idx])
+        if dof == True:
+            dofVar = float(ds["tp_pred_err_var"].attrs["dof"])
```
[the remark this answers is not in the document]
remicousin: And can be set up as int from here

subseas_flex_fcst/maproom.py
```diff
@@ -360,9 +428,8 @@ def fcst_tiles(tz, tx, ty, proba, variable, percentile, threshold):
         )
     else:
         obs_ppf = threshold
-
     # Forecast CDF
-    fcst_dof = int(fcst_var.attrs["dof"])
+    fcst_dof = int(dofVar)#int(fcst_var.attrs["dof"])
```
remicousin: as above, could be set as int when first set up, and data reading could call it fcst_dof right away in which case the whole lien can be removed.

remicousin (changes requested):

subseas_flex_fcst/config-sample.yaml
```diff
@@ -11,13 +11,9 @@ loglevel: info
 
 # Forecast options
 y_transform: true
-forecast_mu_file: CFSv2_SubXPRCP_CCAFCST_mu_Apr_Apr-1-2022_wk1.txt
```
remicousin, line 14: Note that you need one file to read for the layout initialization

subseas_flex_fcst/maproom.py
```diff
@@ -396,21 +396,6 @@ def fcst_tile_url_callback(proba, variable, percentile, threshold, startDate, le
 def fcst_tiles(tz, tx, ty, proba, variable, percentile, threshold, startDate,leadTime):
     # Reading
     fcst_mu, fcst_var, dofVar, obs, hcst = read_cptdataset(leadTime, startDate, y_transform=CONFIG["y_transform"])
-    # Get Issue date and Target season
```
remicousin, line 345: I probably meant to print a title for the map and somehow failed to do so... we can take care of that later.

remicousin:
> A few thoughts / observations I wanted to put here:
> 
>     1. I think based off of the email conversation so far, it sounds like I could remove the lead time definition from the distributions plots titles, but then I will want to describe in the maproom description what the lead time is so the user knows what is going on when they select from the lead time dropdown (tbd how we handle the lead time in the actual code but that is a separate thing).
> 
>     2. I noticed a bug with the layers selection where, if you deselect the tiles layer and then try to add it back on in the layers dropdown, the tile does not reappear. Something to look in to (and I wonder if this occurs in other maprooms as well?).
> 
>     3. The maproom has slowed down quite a bit since applying these changes- not sure if it is now slow enough to warrant exploring options to make it faster or if we think this speed is somewhat normal for our maprooms, given the tiles generation.

I've just finished my review so there is a bit of cross-posting.

For 1., yes except the dropdown should have labels as target date, values as numerical "lead" and in that case lead disappears entirely to the users' eyes. Could still plot "Human" lead (e.g. week 1) in titles if feel the need. All that is in my review.

For 3. I think we need to decide whether we want to combine the data against S/L or just use S/L to read the write data. Also in my review. Speed factor may influence the decision.

subseas_flex_fcst/config-sample.yaml
aaron-kaplan: If we keep it, I would prefer to call it target_period_length.

subseas_flex_fcst/maproom.py
aaron-kaplan: But the variable is never used in this function. It could be eliminated without changing the behavior.

remicousin: that is true. read_cptdataset deals with it.

subseas_flex_fcst/maproom.py
```diff
@@ -45,26 +48,75 @@
 
 APP.layout = layout.app_layout
 
-
-def read_cptdataset(y_transform=False):
-
-    fcst_mu = cptio.open_cptdataset(Path(DATA_PATH, Path(CONFIG["forecast_mu_file"])))
+#function to open all datasets and then combine them
```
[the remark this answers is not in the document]
aaron-kaplan: You're right that we needn't open all of the files in order to get the data for a single (S, L) pair, and doing that is probably what's making it slow.

I encouraged Drew to write something like this `combine_cptdataset` function because, as you say, I have in mind a future in which all datasets will be available in a format like zarr that lets you quickly open a multi-file dataset as as a single multidimensional array. I wanted to build this maproom in a way that would make it easy to migrate to that future data environment. But in retrospect, that was bad advice, because it's so slow.

Keep a copy of `combine_cptdataset` somewhere, because we'll probably eventually want to write a zarrification script, and that will be the core of it. But for now rewrite `read_cptdataset` to open just one file instead of all of them.

subseas_flex_fcst/maproom.py
aaron-kaplan: That sounds appropriate. I forgot this was about metadata attributes rather than data.

subseas_flex_fcst/maproom.py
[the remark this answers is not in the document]
aaron-kaplan: ... and when it comes time to read from zarr instead of netcdf, it will be easy to reimplement `read_cptdataset` to do that.

subseas_flex_fcst/maproom.py
```diff
@@ -314,42 +375,49 @@ def draw_colorbar(proba, variable, percentile):
     Input("proba", "value"),
     Input("variable", "value"),
     Input("percentile", "value"),
-    Input("threshold", "value")
+    Input("threshold", "value"),
+    Input("startDate","value"),
+    Input("leadTime","value")
```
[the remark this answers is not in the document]
aaron-kaplan: @remicousin You're commenting on the inputs to `fcst_tile_url_callback`, not `draw_colorbar`.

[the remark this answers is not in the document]
remicousin: Oops

aaron-kaplan:
Unrelated to Drew's changes, I notice that when you click on the map, the location of the marker doesn't get updated until the graphs finish rendering. Could we update the marker in a separate callback instead of piggybacking it on `local_plots`?

remicousin:
> Unrelated to Drew's changes, I notice that when you click on the map, the location of the marker doesn't get updated until the graphs finish rendering. Could we update the marker in a separate callback instead of piggybacking it on `local_plots`?

Probably. That would probably call for function(s) to select nearest center, and checking outside of domain clicks, etc. that will be used by both callbacks.

subseas_flex_fcst/maproom.py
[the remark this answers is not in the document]
remicousin: That might make obsolete a bunch of other problems if S and L are no more dimensions of the xarrays. Unless we'd want to try and keep clean 1-point such dims?

subseas_flex_fcst/layout.py
```diff
@@ -228,7 +231,80 @@ def navbar_layout(phys_units):
...
+                        value="Apr-1-2022", 
+                    ),
+                ],style={"width":"6%"},
+            ),
+            html.Div(
+                [
+                    help_layout(
+                        "Lead time",
+                        "lead_time",
+                        "Lead time range choices, in weeks from the start date",
```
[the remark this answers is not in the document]
drewmresnick: And would the target date be the target_start value?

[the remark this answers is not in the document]
remicousin: It would ideally be target_start - target_end appended, e.g. 2 Apr - 8 Apr 2022. This means this control needs to be a callback dependent on the choice of S, at least for its labels, as its values can remain 1, 2, 3, 4.

subseas_flex_fcst/maproom.py
drewmresnick: So I will remove `y_transform` from the `combine_cptdataset` function.

Then if I am understanding correctly @remicousin , there are cases in this app where `y_transform` both needs to be false and true? if that is the case then we would not bother to have it in the config file?

subseas_flex_fcst/maproom.py
drewmresnick: is this repeat comment then saying that the config file should have y_transform set to true for all cases it is used? I am a bit confused now..

subseas_flex_fcst/config-sample.yaml
drewmresnick: Ah yes I did not notice that.

subseas_flex_fcst/layout.py
```diff
@@ -298,6 +374,7 @@ def controls_layout(lat_min, lat_max, lon_min, lon_max, lat_label, lon_label):
                     ],
                 ),
             ),
+            html.P([],id="dataFrameSelected"),
```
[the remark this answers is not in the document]
drewmresnick: Something I forgot to delete from when I was testing something :)

subseas_flex_fcst/maproom.py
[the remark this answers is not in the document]
drewmresnick: Alright so I will work on redoing the code.. anywhere smart I might save the code I have already written for later?

subseas_flex_fcst/maproom.py
remicousin: No, for a given app/set-of-files, y_transform is either true or false throughout. To manipulate distributions, we need mean and standard deviation. The standard deviation of the forecast comes from the var (variance) file. However, that variable has sometimes been through a so-called Y-transform. If no Y-transform (false), we can use the var file. If Y-transform (true), the var file is not the variance we expect and we don't know how to use it or transform it back to what we need. But we can then estimate that variance (or standard deviation) using the hindcast and obs. So the y_transform boolean triggers using that or the other.

subseas_flex_fcst/maproom.py
remicousin: no sorry for the confusion. Throughout the code, y_transform should point to the value in the config file. In this case at hand, that value should be true. (except at the function definition where y_transform should be set up to a default. I believed I had picked False as default, maybe that's where this False repeated everywhere came from).

subseas_flex_fcst/maproom.py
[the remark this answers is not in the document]
remicousin: OR... is there time to implement the zarrification process to this project? Maybe not. I'd like this to work for seasonal forecasts by the trainings starting in mid-Ocotber and we're not quite there yet. And there are other things to attend (partners' DLs updates, onset python maprooms, prepare for training). That would also need to work with an updating framework since there will be new seasonal forecasts at partners' sites monthly. And so we'd need to train them on that as well. So probably a bit tight, especially given Aaron didn't suggest that in the first place.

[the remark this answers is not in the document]
remicousin: As for where to put the code. Maybe it makes sense to start another .py file, maybe predictions.py as there are some functions that are surfacing as needed and that would potentially be needed by other tools. There is combine and read cpt datasets. We need to make a function out of what computes the target start/end from S/L and we'll need more like this for other type of forecasts.

drewmresnick:
> > A few thoughts / observations I wanted to put here:
> > ```
> > 1. I think based off of the email conversation so far, it sounds like I could remove the lead time definition from the distributions plots titles, but then I will want to describe in the maproom description what the lead time is so the user knows what is going on when they select from the lead time dropdown (tbd how we handle the lead time in the actual code but that is a separate thing).
> > 
> > 2. I noticed a bug with the layers selection where, if you deselect the tiles layer and then try to add it back on in the layers dropdown, the tile does not reappear. Something to look in to (and I wonder if this occurs in other maprooms as well?).
> > 
> > 3. The maproom has slowed down quite a bit since applying these changes- not sure if it is now slow enough to warrant exploring options to make it faster or if we think this speed is somewhat normal for our maprooms, given the tiles generation.
> > ```
> 
> I've just finished my review so there is a bit of cross-posting.
> 
> For 1., yes except the dropdown should have labels as target date, values as numerical "lead" and in that case lead disappears entirely to the users' eyes. Could still plot "Human" lead (e.g. week 1) in titles if feel the need. All that is in my review.
> 
> For 3. I think we need to decide whether we want to combine the data against S/L or just use S/L to read the write data. Also in my review. Speed factor may influence the decision.

As of now the conversation on this PR lead me to start working on using the S/L to read the data instead of combining it, so I am working on that now. Then I have made a checklist of improvements based off your reviews I will start getting to.

remicousin:
ok. Consider merging when you are in an ok state and save improvements for other PRs (done by you or others). It's easy to "convert" a review comment into an issue. There are buttons for that.

drewmresnick:
> ok. Consider merging when you are in an ok state and save improvements for other PRs (done by you or others). It's easy to "convert" a review comment into an issue. There are buttons for that.

Sure. I am still trying to sift through everything to decipher which are just improvements versus which comments are the more important changes to make before merging. 

I would think that changing the maproom to select a single file instead of opening and combining all files is important. And, changing the representation of the leads. Then maybe the rest can be made into "issues" to address in other PR's?

subseas_flex_fcst/layout.py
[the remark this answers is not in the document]
drewmresnick: will be addressed in separate issue.

subseas_flex_fcst/maproom.py
```diff
@@ -135,19 +185,30 @@ def local_plots(n_clicks, click_lat_lng, latitude, longitude):
...
+        lead_time = 1
+        #lead_time = (issue_date_td + timedelta(days=(targetStart_add + 3))).strftime("%-d %b %Y") #from when I made lead time the midpoint date
+    elif fcst_var["L"].values == "Week 2":
+        lead_time = 8
+        #lead_time = (issue_date_td + timedelta(days=(targetStart_add +3))).strftime("%-d %b %Y")
+    elif fcst_var["L"].values == "Week 3":
+        lead_time = 15
+        #lead_time = (issue_date_td + timedelta(days=(targetStart_add+3))).strftime("%-d %b %Y")
+    elif fcst_var["L"].values == "Week 4":
+        lead_time = 22
```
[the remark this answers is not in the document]
drewmresnick: For this, is there a mathematical expression used to calculate the dates? Since I am failing to see how changing L to be integers like that would be any simpler in the way I have coded things.

subseas_flex_fcst/maproom.py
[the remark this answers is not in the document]
drewmresnick: Okay, I changed the code to open only a single ds at a time, pushing those changes soon to this PR for review. I have the old function saved in a temporary file - I can make a separate predictions.py file and move all the functions into that file if that is direction we are moving in with all of the functions for this app.

subseas_flex_fcst/maproom.py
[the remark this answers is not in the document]
drewmresnick: Currently not relevant due to updated code of opening a single file at a time.

subseas_flex_fcst/maproom.py
```diff
@@ -45,26 +48,75 @@
...
+        fileCompList = re.split('[_.]',filesNameList[idx])
+        fileCompList = [x for x in fileCompList if "/" not in x]
+
+        leadTimeWk = [x for x in fileCompList if x.startswith('wk')]
+
+        dateFormat = re.compile(".*-.*-.*")
+        startDateStr = list(filter(dateFormat.match,fileCompList))
+        startDate = startDateStr[0]
+        #startDate = datetime.strptime(startDateStr[0],
+        #'%b-%d-%Y').date()
```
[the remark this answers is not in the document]
drewmresnick: Also no longer relevant in updated code (would need to be fixed if we ever end up using the combine_cptdataset() function.

subseas_flex_fcst/maproom.py
drewmresnick: For now, as you said, the problem solved itself. So this is an issue for a later date if we end up wanting to use the old function that combines all the data.

subseas_flex_fcst/maproom.py
drewmresnick: Made into an issue for later.

subseas_flex_fcst/config-sample.yaml
drewmresnick: I do currently use this in the code when calculating target start / end. But if there is a better way to do this calculation then I will change it.

For now I kept it in the config file but changed the name to be more descriptive.

drewmresnick:
I have pushed commits to solve a few of bigger changes. Then I have made new issues out of the remaining items as suggested by Remi. Those remaining items seem like things that I could do in this PR before it is merged I would think, unless it is better practice to merge this before this PR gets any clunkier and deal with those issues on their own?

The issues that have not been completed are:

1. Make map marker update in its own callback (although now that we are not opening all of the data files, it updates more quickly. So not sure if this is necessary, or if we should just do it anyways..)
2. Make a map title callback 
3. Update the lead time and target dates: representation, calculation, etc

subseas_flex_fcst/config-sample.yaml
```diff
@@ -17,7 +17,7 @@ forecast_var_filePattern: CFSv2_SubXPRCP_CCAFCST_var_*.txt
 obs_filePattern: CFSv2_SubXPRCP_CCAFCST_obs_*.txt
 hcst_filePattern: CFSv2_SubXPRCP_CCAFCST_xvPr_*.txt
 variable: Precipitation Anomaly
-tp_length: 6
+target_period_length: 6
```
aaron-kaplan, line 20: I assume the target period is a week, i.e. 7 days. Then target_period_length should be 7, and the end of the period should be start + length - 1.

subseas_flex_fcst/maproom.py
```diff
@@ -244,7 +244,7 @@ def local_plots(n_clicks, click_lat_lng, startDate, leadTime, latitude, longitud
         xaxis_title=f'{CONFIG["variable"]} ({fcst_mu.attrs["units"]})',
         yaxis_title="Probability of exceeding",
         title={
-            "text": f"{target_start} - {target_end} forecast issued {issue_date} (lead time {lead_time} days) <br> at ({fcst_mu.Y.values}N,{fcst_mu.X.values}E)",
+            "text": f"{target_start} - {target_end} forecast issued {issue_date} <br> at ({fcst_mu.Y.values}N,{fcst_mu.X.values}E)",
```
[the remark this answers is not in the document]
drewmresnick: Yes, and I have currently created an issue where a part of the PR will be to edit the input dropdown to change it from displaying leads, to displaying target dates.

subseas_flex_fcst/maproom.py
```diff
@@ -48,62 +48,37 @@
...
-        if dof == True:
-            dofVar = float(ds["tp_pred_err_var"].attrs["dof"])
-            ds = ds.assign(dof=dofVar)
-        if len(ds['T']) == 1:
-            ds['T'] = [startDate] #for obs data there are many dates
-        ds = ds.expand_dims(leadTimeDict)
-        dataDic[idx] = ds
-    dataCombine = xr.combine_by_coords([dataDic[x] for x in dataDic],combine_attrs="drop_conflicts")
-    return dataCombine
+        x = re.search(f"{pattern}",filesNameList[idx])
```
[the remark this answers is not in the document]
drewmresnick: yes, I seem to have made things much harder than they needed to be on that one :)

subseas_flex_fcst/maproom.py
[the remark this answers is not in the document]
drewmresnick: Okay. I am still deciding how to process on this part. If everything else is wrapped up, I think I will merge this PR and then start a new one to deal with all the lead time stuff, unless there are objections on that.

aaron-kaplan (approved):

subseas_flex_fcst/layout.py
```diff
@@ -18,11 +23,23 @@
...
+#Initialization for start date dropdown to get a list of start dates according to files available
+filesNameList = glob.glob(f'{DATA_PATH}/{CONFIG["forecast_mu_filePattern"]}')
+startDates = []
+for idx, i in enumerate(filesNameList):
+    startDate = re.search("\w{3}-\w{1,2}-\w{4}",filesNameList[idx])
+    startDatedt = datetime.strptime(startDate.group(),"%b-%d-%Y")
+    startDateStr = startDatedt.strftime("%b-%-d-%Y")
+    startDates.append(startDatedt)
+startDates = sorted(set(startDates))
+startDates = [i.strftime("%b-%-d-%Y") for i in startDates] #needs to have date with no zero padding to match the file path namesme
```
remicousin, line 37: This will probably help to convert to a function as well, so we can modulate when wanting to use this for other kind of forecasts. Can make that an issue for now.

subseas_flex_fcst/layout.py
```diff
@@ -228,7 +248,66 @@ def navbar_layout(phys_units):
...
+                    "vertical-align": "top",
+                }
+            ),
+            html.Div(
+                [
+                    dcc.Dropdown(
+                        id="startDate",
+                        clearable=False,
+                        options = startDates,
+                        value=startDates[-1],
```
remicousin, line 274: In Ingrid's version, I reverted the order of the list so that he most recent Start date are at the top of the menu. Supposedly users will be interested in the most recent stuff, so it makes sense to do that. It's possibly a little less true with the type-in functionality to find possible options, as it makes their ordering less relevant.
Either way... no strong opinion on reverting order or not.

subseas_flex_fcst/maproom.py
```diff
@@ -45,27 +49,28 @@
 
 APP.layout = layout.app_layout
 
-
-def read_cptdataset(y_transform=False):
-
-    fcst_mu = cptio.open_cptdataset(Path(DATA_PATH, Path(CONFIG["forecast_mu_file"])))
+#Should I move this function into the predictions.py file where I put the other funcs?
+#if we do so maybe I should redo the func to be more flexible since it is hard coded to read each file separately..
+def read_cptdataset(leadTime, startDate, y_transform=CONFIG["y_transform"]):
```
remicousin, line 54: Possibly... or make it an issue. The key here is that is we are going to use the same app for other cpt forecasts (biweekly, seasonal...) so all the functions associated with reading will either have to be more fancy to accommodate different cases, or have counterparts for the different cases, and the config file will tell which case we're in.
I am just not sure what shape that will all take as I review this now. So it may make sense to leave to other PR.

remicousin:
I put a few comments that can all become issues.
I am about to approve but can't find the button and need to run to bus now...

subseas_flex_fcst/layout.py
drewmresnick: converted to issue.

subseas_flex_fcst/layout.py
drewmresnick: I will make note of this, and can ask for feedback when I show the new changes.

subseas_flex_fcst/maproom.py
drewmresnick: I think it is best to make a separate PR for now; I have made a new issue for this.

PR #159, opened 2022-08-15
Go back to old Ethiopia forecast configuration that covers only Somali region

aaron-kaplan:
Azhar came out with a new forecast configuration, and Rémi set up a new catalog entry for it, but then we reverted to the old one because triggers have already been set so we're not allowed to change the forecast configuration until the season is over. I seem to have restored the old data files without rolling back the change in the update script.


aaron-kaplan:
Seems like my memory is faulty. Apparently we decided to go with the new configuration after all.

PR #161, opened 2022-08-17
Toggle to include/exclude upcoming/current season in threshold calculation.

aaron-kaplan:
The goal of this PR was to give the user control over whether the forecast for the upcoming season was included in the threshold calculation or not. While working on that I also did some refactoring, and changed to a different Bootstrap theme because the color choices of the one we were using seemed unconventional and disturbing. Please read commit-by-commit.

### DIFF
--- a/fbfmaproom/__about__.py
+++ b/fbfmaproom/__about__.py
@@ -4,4 +4,4 @@
 name = "fbfmaproom"
 author = "IRI, Columbia University"
 email = "help@iri.columbia.edu"
-version = "2.35.0"
+version = "2.36.0"
--- a/fbfmaproom/assets/style.css
+++ b/fbfmaproom/assets/style.css
@@ -114,12 +114,24 @@ table.supertable tbody {
     background-color: #fdfd96;
 }
 
+.cell-excluded-severity-0 {
+    background-color: #98985a;
+}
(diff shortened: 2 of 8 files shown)


aaron-kaplan:
Currently running on http://shortfin01.iri.columbia.edu:9999/fbfmaproom/niger

aaron-kaplan:
- [ ] Update export endpoint
- [ ] Fix tests
- [ ] Tooltip for checkbox
- [ ] Fix column alignment and cell padding

aaron-kaplan:
I got the checkbox in the table working, but the code was messy. I worked on cleaning it up for a while, but it became apparent that it would take more work than it was worth, given that we didn't intend to keep that UI forever. So I moved the checkbox to the table controls, which in retrospect is probably where Dan would have wanted it in the first place.

I've retained some of the refactoring I did while trying to get the messy version cleaned up. Best read commit-by-commit.

@kgraaf please review.

aaron-kaplan:
Dan has approved. @kgraaf please review.

aaron-kaplan:
@kgraaf this has been waiting for your review for over two weeks. Please finish reviewing.

kgraaf (approved):

Issue #146, opened 2022-07-26
Confusion about Ingrid colormap representation


remicousin:
> Turns out that out of the 4 cases I ran into: fbf's Ingrid colormap's representation; correlationcolorscale and the 2 PoE colorscales (once static); only the fbf's one has Ingrid's colormap returns something that corresponds to what start/endcolormap states. In the 3 other cases, colors order were switched, and/or replaced, and/or additional incrementations/sequencing were added...

I didn't understand this. If you want me to help look into it, please provide more details. I think we can merge this in the meantime.

_Originally posted by @aaron-kaplan in https://github.com/iridl/python-maprooms/issues/113#issuecomment-1194425554_

remicousin:
Ingrid's colormap is not doing what we expect. Worst case scenario I found:
correlationcolorscale, defined in [Ingrid](https://bitbucket.org/iridl/ingrid/src/master/src/ingridinit.tex#lines-17290) returns a [colormap](http://iridl.ldeo.columbia.edu/SOURCES/.NOAA/.NCEP-NCAR/.CDAS-1/.MONTHLY/.Intrinsic/.PressureLevel/.phi/correlationcolorscale/colormap/#expert) where:

- colors' order are reverted (reds first then blues, e.g. first color 8388608 is [Dark Red](https://www.mathsisfun.com/hexadecimal-decimal-colors.html)
- Neutral color 11920639 is not Ingrid's mocassin and 13959039 is not Ingrid's aquamarine

Interestingly enough, Ingrid's aquamarine is
/aquamarine     127 255 212     RGBdef
but  13959039 [returns](https://www.mathsisfun.com/hexadecimal-decimal-colors.html) RGB 212 255 127 

As far as I can tell, pingrid.py returns what colormap is, which is not the Ingrid colorscale. So can't blame pingrid.py... but can't rely on Ingrid's colormap, thus can't reallt use pingrid.py features that rely on it.

I ran into other similar cases. The Ingrid colormap in fbfmaproom config file is reproducing what it's supposed to.

aaron-kaplan:
[...]

PR #165, opened 2022-09-06
dummy ad hoc changes to illustrate colorscale flipping

remicousin:
This is not a PR meant to be merged. Rather to more easily describe changes to support discussion for this [issue](https://github.com/iridl/python-maprooms/issues/146). The change Aaron pointed to is one thing but there is also a change eliminating DataArrayEntry which makes changed necessary in both maproom.py and pingrid.py to try to repeat this past behavior.

The minimalist changes I've done intend to reproduce what was happening before using the correlationcolorscale case study. So when running the app, change to Precipitation Anomaly to fall into that case. Those changes revert the colorscale for both the colorscale and its application to the data on the map. However, even though one can see that that correlation colorscale is now in the right order (from blue to red), there seems to be 2 wrong individual colors: the central neutral color which is
/moccasin       255 228 181     RGBdef
in Ingrid (some sort of beige color). And the following color starting the "reds" should be yellow and it looks like some pale green. [Reminder](http://iridl.ldeo.columbia.edu/ds%3A/SOURCES/.NOAA/.NCEP-NCAR/.CDAS-1/.MONTHLY/.Intrinsic/.PressureLevel/.phi/P/(250)/VALUE/dup/T/(Jan%201991)/(Dec%202020)/RANGE/yearly-climatology/sub/std_anomaly_colors2//long_name/(Geopotential%20Height%20Anomaly)/def//name/(phi1)/def/SOURCES/.NOAA/.NCEP-NCAR/.CDAS-1/.MONTHLY/.Intrinsic/.PressureLevel/.phi/P/(250)/VALUE/T/(Jan%201991)/(Dec%202020)/RANGE/yearly-climatology/DATA/60/STEP//name/(phiclim)/def/%3Ads/a-/.phi1/correlationcolorscale/DATA/AUTO/AUTO/RANGE/-a-/.phiclim/-a/X/Y/fig%3A/colors/grey/unlabelledcontours/black/thin/solid/coasts/countries/%3Afig/(phi1)/-200/200/plotrange//T/last/plotvalue/X/-20/340/plotrange/Y/-65/75/plotrange//plotaxislength/650/psdef//plotborder/72/psdef//XOVY/null/psdef//color_smoothing/null/psdef//antialias/true/psdef/#expert) of what Ingrid's colorscale looks like.

So I'd think something is off in the color reverting as I applied it. Either what I did did reproduce what was happening before and there has always been a problem with that very case (that was either ignored or undiscovered).

Given then many things have changed since then in pingrid.py. It may not be worth to try and understand the changes that matter to this problem since then, and rather either reinstate a new reverting option or abandon trying to convert ingrid colorscales to python.

### DIFF
--- a/pingrid.py
+++ b/pingrid.py
@@ -56,6 +56,19 @@ class BGRA(NamedTuple):
     alpha: int
 
 
+class RGB(NamedTuple):
+    red: int
+    green: int
+    blue: int
+
+
+class RGBA(NamedTuple):
+    red: int
+    green: int
+    blue: int
+    alpha: int
+
+
 class DrawAttrs(NamedTuple):
     line_color: Union[int, BGR, BGRA]
     background_color: Union[int, BGR, BGRA]
@@ -171,7 +184,7 @@ def tile(da, tx, ty, tz, clipping=None, test_tile=False):
         return empty_tile()
 
     im = (z - da.attrs["scale_min"]) * 255 / (da.attrs["scale_max"] - da.attrs["scale_min"])
-    im = apply_colormap(im, parse_colormap(da.attrs["colormap"])[::-1])
+    im = apply_colormap(im, parse_colormap(da.attrs["colormap"]))
     if clipping is not None:
         if callable(clipping):
             clipping = clipping()
@@ -408,9 +421,9 @@ def produce_test_tile(
     return im
 
 
-def parse_color(s: str) -> BGRA:
+def parse_color(s: str) -> RGBA:
     v = int(s, 0)  # 0 tells int() to guess radix
-    return BGRA(v >> 0 & 0xFF, v >> 8 & 0xFF, v >> 16 & 0xFF, 255)
+    return RGBA(v >> 0 & 0xFF, v >> 8 & 0xFF, v >> 16 & 0xFF, 255)
 
 
 def parse_color_item(vs: List[BGRA], s: str) -> List[BGRA]:
--- a/pingrid.py
+++ b/pingrid.py
@@ -26,6 +26,7 @@ RAINBOW_COLORMAP = "[0x0000ff [0x00ffff 51] [0x00ff00 51] [0xffff00 51] [0xff000
 RAIN_POE_COLORMAP = "[0x000000 [0xa52a2a 35] [0xffa500 36] [0xffff00 36] 0xffe465 [0xffe465 35] 0x32cd32 [0x40e0d0 35] [0x0000ff 36] [0xa020f0 36]]"
 RAIN_PNE_COLORMAP = "[0xa020f0 [0x0000ff 35] [0x40e0d0 36] [0x32cd32 36] 0xffe465 [0xffe465 35] 0xffff00 [0xffa500 35] [0xa52a2a 36] [0x000000 36]]"
 CORRELATION_COLORMAP = "[0x000000 0x000080 [0x0000ff 25] [0x00bfff 26] [0x7fffd4 39] [0x98fb98 26] 0xffe465 [0xffe465 25] 0xffff00 [0xff8c00 38] [0xff0000 39] [0x800000 39] 0xa52a2a]"
+CORRELATION_COLORMAP = "[null 0 8388608 [16711680 25] [16760576 26] [13959039 39] [10025880 26] 11920639 [11920639 26] 65535 [36095 38] [255 38] [128 39] 2763429]"
 
 def error_fig(error_msg="error"):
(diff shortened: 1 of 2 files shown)


aaron-kaplan:
> This other dummy change prints the right colorscale from Ingrid

Cool, sounds like we've gotten to the bottom of it? Did you verify all the colormaps we use, or just one?

remicousin:
I can test the other ingrid colorscales in forecast.

Then is pingrid.RAINBOWCOLORMAP an ingrid one or other? I am not sure what is used in fbf. Then parse_color, that I changed, is used by cascading functions, so I am not sure where the proper changes should be. I guess sometimes we are feeding a RGB series and sometimes a BGR and Igor had a check on that somewhere to apply the switch or not, with his reverseclormap boolean. But that's in a function that is gone.

I don't know if it's any clearer to you. Just checking before trying and going to the bottom of things.

aaron-kaplan:
I don't remember, but I think RAINBOW_COLORMAP probably came from Ingrid. That's circumstantially supported by the fact that there's one called rainbowColorMap in ingridinit.tex.

The colormaps used by FBF are in the [config file](https://github.com/iridl/python-maprooms/blob/master/fbfmaproom/fbfmaproom-sample.yaml) -- search for &raincm, &pnepcm, and &vulncm.

Looking at the tile generation code, I think rather than introducing the RGB/RGBA classes as you've done here, you should stick with BGR/BGRA, and instead flipt the order of the bytes in `parse_color`. i.e.

```
    return BGRA(v >> 16 & 0xFF, v >> 8 & 0xFF, v >> 0 & 0xFF, 255)

```

remicousin:
hmmm... I also need to know what CONFIG to give to run the tests.

remicousin:
hmmm... I am not quite there yet. It's running it seems but this:
http://devi:8063/fbfmaproom/niger/
returns a JSON (I added the country as without, it returns KeyError: 'path'

I am also not on the post pyaconf era, should I rebase to get going?

aaron-kaplan:
That's weird. I don't know what would cause that. Please do rebase to get us to a good starting place for debugging, but a priori I can't see how that would fix this.

remicousin:
Rebasing and remaking environment from lock allowed me to build and serve. Now it's apparent to me that the colorscales in fbfmaprooms (at least some), need not be reverted (RGB/BGR) like for Ingrid ones. How do we want to go about this?

remicousin:
@aaron-kaplan let me know if now is a good time to resolve this. As a reminder, the issue is that some colorscales are defined as RGB and others as BRG but our code ignores that and assume always one of the 2. So we need either to pick a case and change the colorscales defined according to the other case, either we need code to be told what to do and when.

aaron-kaplan:
> Now it's apparent to me that the colorscales in fbfmaprooms (at least some), need not be reverted (RGB/BGR) like for Ingrid ones.

But the colorscales in fbfmaproom came from ingrid. At least I thought they did. Have you compared pingrid.RAINBOW_COLORMAP with ingrid's rainbowColorMap?

In any case, if you have a way of interpreting colormap representations that works for all ingrid colormaps (not just for one example), then we will simply define that to be the right interpretation, and rewrite the colormaps in pingrid such that they get interpreted as intended by that definition. Just be sure you've got the ingrid end right first.

remicousin:
With the changes in this PR:

- master's pingrid.RAINBOW_COLORMAP is different from Ingrid's. Using Ingrid's does the right thing.
- the 3 colorscales for flex fcst (PO/N/E and correlation) come from Ingrid and are rendered correctly.
- fbfmaproom's prob colorscale goes now from blue to red (instead of red to blue)
- fbfmaproom's vuln colorscale is now shared of blue (instead of shades of red)
- fbfmaproom's prcp colorscale: I don't know how to reveal it.

So in short, all the colorscales I added as direct copies of Ingrid's colormaps are rendered as expected. All pre-existing colorscales are flipped (RGB/GRB)

remicousin:
Shall I rewrite the FbFmaproom colorscales through the same order (BGR?) than the other ones? so that i) we can readily use something from Ingrid and ii) we don't have to bother with cases where we flip or don't?

aaron-kaplan:
Yes, please.

remicousin:
ok. Can you tell me how to see:
prcp colorscale
?

aaron-kaplan:
I don't know that it's being used in fbf. grep for it, and if you don't
find it, just delete it.

remicousin:
Was resolved through PR #244

PR #166, opened 2022-09-07
Remic/zarring

remicousin:
This is the 1st step to setting up onset maproom for these 3 new countries. I've actually run the zarr-ing for ZMD ENACTS, but not for the others. Next step will be to adapt the config for the Maproom for ZMD, serving as well to test if zarr-ing worked well. Once that future PR is approved, will run zarr-ing to other 2 countries and adapt their config for Maproom.

I am putting a lot of reviewers for both FYIs and actual review/approval, depending who you are.


remicousin:
May I get a review/approval from maybe @aaron-kaplan on this? It's really just the "zarr conversion" section of the new config files that have been set up. There is not much. So I can keep going.

aaron-kaplan (approved):

PR #171, opened 2022-09-14
Onset maproom minor improvements 

drewmresnick:
A few minor things that has been requested by Tufa.

The one thing I could not figure out a good solution for was the min/max zoom option. I know it had been said that setting min/max zoom in the way I did was not a good permanent solution because it is dependent on screen size (I think was the reason?). I couldn't find a workaround yet. so pushed my changes with what I have for now.

The other thing I noticed that might be nice would be to have an option to view boundary names, perhaps when you mouseover? Since you can't really tell what you are looking at very well now. Just a thought...

### DIFF
--- a/onset_maproom/config-nma-ethiopia.yaml
+++ b/onset_maproom/config-nma-ethiopia.yaml
@@ -52,7 +52,7 @@ name_adm3: Woredas
 
 # Onset_and_Cessation
 app_title: Growing Season Maproom
-onset_and_cessation_title: Planting and Harvest Decision Support Maproom
+onset_and_cessation_title: Planting Date Decision Support Maproom
 default_search_month: Jun
 default_running_days: 5
 default_min_rainy_days: 3
--- a/onset_maproom/config-nma-ethiopia.yaml
+++ b/onset_maproom/config-nma-ethiopia.yaml
@@ -37,7 +37,8 @@ rr_mrg_zarr_path: /data/aaron/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/rr
 # Ethiopia
 logo: Ethiopia_IRI_98x48.png
 institution: NMA
-zoom: 6
+zoom: 5
+maxBounds: [[-5.790897,25.839844],[17.978733,61.523438]] #user map pan limit as [[lat],[lng]]
 shapes_adm1: select gid as key, name_1 as label, ST_AsBinary(the_geom) as the_geom
                   from eth_regions_dd
 name_adm1: Regions
@@ -52,7 +53,7 @@ name_adm3: Woredas
 
 # Onset_and_Cessation
 app_title: Growing Season Maproom
-onset_and_cessation_title: Planting Date Decision Support Maproom
+onset_and_cessation_title: Planting and Harvest Decision Support Maproom
 default_search_month: Jun
 default_running_days: 5
 default_min_rainy_days: 3
--- a/onset_maproom/layout.py
+++ b/onset_maproom/layout.py
@@ -65,6 +65,7 @@ def app_layout():
                                         },
                                     ),
                                 ],
+                                style={"overflow":"scroll","height":"45vh"},
                                 className="g-0",
                             ),
                             dbc.Row(
@@ -81,7 +82,7 @@ def app_layout():
                                         },
                                     ),
                                 ],
-                                style={"overflow":"scroll","height":"55vh"},
+                                style={"overflow":"scroll","height":"50vh"},
                                 className="g-0",
                             ),
                         ],
@@ -239,9 +240,9 @@ def controls_layout(lat_min, lat_max, lon_min, lon_max, lat_label, lon_label):
                         """
                     ),
                 ],
-                style={"position":"relative","height":"30vh", "overflow":"scroll"}, #,"height":"20%"
+                style={"position":"relative","height":"28vh", "overflow":"scroll"}, #,"height":"20%"
             ),
-            html.H3("Controls Panel",style={"padding":"1rem 1rem"}),
+            html.H3("Controls Panel",style={"padding":".5rem"}),
             html.Div(
                 [
                     Block("Pick a point",
@@ -352,12 +353,12 @@ def controls_layout(lat_min, lat_max, lon_min, lon_max, lat_label, lon_label):
                         ison=CONFIG["ison_cess_date_hist"]
                     ),
                 ],
-                style={"position":"relative","height":"70vh", "overflow":"scroll"},
+                style={"position":"relative","height":"53vh", "overflow":"scroll"},
             ),
         ], #end container
         fluid=True,
         className="scrollable-panel p-3",
-        style={"overflow":"scroll","height":"100vh","padding-bottom": "1rem", "padding-top": "1rem"},
+        style={"overflow":"scroll","height":"95vh","padding-bottom": "1rem", "padding-top": "1rem"},
     )
 
 def map_layout(center_of_the_map):
@@ -451,7 +452,7 @@ def map_layout(center_of_the_map):
                 minZoom = CONFIG["zoom"] - 1,
                 style={
                     "width": "100%",
-                    "height": "45vh",
+                    "height": "39vh",
                 },
             ),
             html.H6(
@@ -459,7 +460,7 @@ def map_layout(center_of_the_map):
             )
         ],
         fluid=True,
-        style={"padding": "1rem"},
+        style={"padding": "0rem", "height":"8vh"},
     )
 
 
--- a/onset_maproom/layout.py
+++ b/onset_maproom/layout.py
@@ -81,20 +81,19 @@ def app_layout():
                                         },
                                     ),
                                 ],
-                                style={"overflow":"scroll","height":"50vh"},
+                                style={"overflow":"scroll","height":"55vh"},
                                 className="g-0",
                             ),
                         ],
                         sm=12,
                         md=8,
-                        style={"background-color": "white"},
                     ),
                 ],
                 className="g-0",
             ),
         ],
         fluid=True,
-        style={"padding-left": "0px", "padding-right": "0px"},
+        style={"padding-left": "1px", "padding-right": "1px"},
     )
 
 
@@ -449,6 +448,7 @@ def map_layout(center_of_the_map):
                 id="map",
                 center=center_of_the_map,
                 zoom=CONFIG["zoom"],
+                minZoom = CONFIG["zoom"] - 1,
                 style={
                     "width": "100%",
                     "height": "45vh",
@@ -499,5 +499,4 @@ def results_layout():
             #     ),
             # )
         ], #end div
-        style={"height":"60vh"}
     )
--- a/onset_maproom/layout.py
+++ b/onset_maproom/layout.py
@@ -414,8 +414,7 @@ def map_layout(center_of_the_map):
                             dlf.Overlay(
                                 dlf.GeoJSON(
                                     id="borders_adm1",
-                                    data={"features": []},
-                                    children=dlf.Tooltip("label",direction="center"),#adds tooltip on hover but only when hover over edges unless fill set to True
+                                    data= {},
                                     options={
                                         "fill": True,
                                         "color": "black",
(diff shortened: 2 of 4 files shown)


aaron-kaplan:
Does this help with restricting the zoom? https://leafletjs.com/reference.html#map-setmaxbounds

drewmresnick:
> 

This does not help with the zoom. If there is no minZoom defined it will let the user zoom out as much as they want- what it does is just that if the user moves out of the bounds, it will bounce back to having the bounding box region be in the center of the map, which is useful but not necessarily what we are looking for. It seems like a lot of the zoom and bounding keyword arguments for dl.Map are in some way dependent on the defined zoom levels.

Could you remind me why it is not recommended to use zoom levels? I just read this [leaflet tutorial ](https://leafletjs.com/examples/zoom-levels/)to try and jog my memory on the subject but I can't really recall the reasons you had given for why we want to stay away from using the zoom levels.

aaron-kaplan:
I see. Can we use https://leafletjs.com/reference.html#map-fitbounds , check what zoom level it picked, and then set minZoom to that?

There's nothing inherently wrong with using zoom levels, but I don't think it's easy to pick one that will work for everyone, given the range of screen resolutions that are available and the fact that the browser window isn't necessarily maximized. Also it would be nice to have a method that would just automatically display the dataset in question at an appropriate zoom level, as opposed to making the programmer try different zoom levels and decide which one to use as the min.

It's not do-or-die. If you can't get the code to figure out minZoom automatically, hard-coding it to something that works well for your screen is ok for now.

drewmresnick:
@aaron-kaplan Latest commit is the best solution I could manage for adding the feature to print admin boundary names when hovering over a feature.

A little laggy, but I was not able to figure out a more clear solution than this after searching for a while.

Also haven't yet worked out how to style the tooltip, but I can look into this if we continue working on this feature.

aaron-kaplan:
> A little laggy, but I was not able to figure out a more clear solution than this after searching for a while.

We might be able to make it faster by using client-side javascript code. What you've done is probably the best we can do in python. Good enough for now.

The tooltip placement isn't always great. Sometimes it appears underneath the mouse pointer so I can't read it, and for the Oromia region the tooltip overlaps the border. Minor things, not worth spending a lot of time on at this point.

aaron-kaplan:
Are we going to finish reproducing the original Ingrid version in python? Features that I notice are missing (probably not an exhaustive list):
 
- Display borders for lower admin levels
- Display the name of the region that the selected point is in
- Standard deviation and probability of exceedance maps
- I could have sworn there was a PDF export option on the original, though I can't find it now. @remicousin ?

drewmresnick:
> Are we going to finish reproducing the original Ingrid version in python? Features that I notice are missing (probably not an exhaustive list):
> 
>     * Display borders for lower admin levels

_You can select which admin level you want to display by using the modes dropdown on the top right of the app, like you would for the fbf maproom (though admittedly it is not the best placed, I could not decide where it would best be located..) This was the only solution for having one admin level selected at once, unless we would rather have two of the layer buttons on the map (one for the base map selection and one for the admin levels)._ 

>     * Display the name of the region that the selected point is in

_This I can add as well._
 
>     * Standard deviation and probability of exceedance maps
> 
>     * I could have sworn there was a PDF export option on the original, though I can't find it now. @remicousin ?

_The last two I cannot speak on so I will let Remi respond._

onset_maproom/maproom.py
```diff
@@ -22,6 +22,7 @@
 from shapely import wkb
 from shapely.geometry.multipolygon import MultiPolygon
 from psycopg2 import sql
+import geopandas as gpd
```
remicousin, line 25: this has been added but not environment files haven't been changed accordingly.
Which way should be go about this? Add manually to the environment, test that things work, then update environment in version-control? or synchronously?

onset_maproom/config-nma-ethiopia.yaml
```diff
@@ -52,7 +52,7 @@ name_adm3: Woredas
 
 # Onset_and_Cessation
 app_title: Growing Season Maproom
-onset_and_cessation_title: Planting and Harvest Decision Support Maproom
+onset_and_cessation_title: Planting Date Decision Support Maproom
```
remicousin, line 55: Tufa wanted that change? It's not very indicative that there is cessation date shown in the App too...

onset_maproom/maproom.py
aaron-kaplan: What do you mean by synchronousy?

In this case, geopandas isn't actually being used, so the solution is just to remove this line.

drewmresnick: Yes in this case I do just need to delete the line since I did not end up using it. However this is a broader discussion I would like to have at some point since it is still confusing to me exactly how we go about updating environments properly. But this conversation seems best to have elsewhere.

onset_maproom/config-nma-ethiopia.yaml
drewmresnick: That is what he asked it to be named last I spoke with him.

remicousin:
A few comments quickly to make sure I can drop them before catching the bus.

I am not really convinced that it is the way to go for the following items:

1. Plot one admin boundary at a time. I don't think that's a good thing. Some places/users would want more than one at a time and thus checkboxes are needed for that rather than drop-down menu as a result. And checkboxes allow to have one at a time so covers that case.
2. Why make a dedicated menu for it when there is a layer control built in the mapping functionality? It makes more sense to me that layers controls are together, and if so, with the control that comes with the map.

The other things look good but I've not had the time to have a closer look yet. I just wanted to post that as I see that as problematic.

remicousin:
standard deviaton and prob. exc. maps are here: they are the last 2 choices of the Ask the map control.

remicousin:
Alright... I have to go now... more thorough review tomorrow (hopefully).

drewmresnick:
> A few comments quickly to make sure I can drop them before catching the bus.
> 
> I am not really convinced that it is the way to go for the following items:
> 
>     1. Plot one admin boundary at a time. I don't think that's a good thing. Some places/users would want more than one at a time and thus checkboxes are needed for that rather than drop-down menu as a result. And checkboxes allow to have one at a time so covers that case.
> 
>     2. Why make a dedicated menu for it when there is a layer control built in the mapping functionality? It makes more sense to me that layers controls are together, and if so, with the control that comes with the map.
> 
> 
> The other things look good but I've not had the time to have a closer look yet. I just wanted to post that as I see that as problematic.

This was a first pass at figuring out a way to print admin boundary names; the way things stand, doing so would only make sense if a single layer was displayed at a time, and I figured the user would not need to have more than one boundary layer at a time (but perhaps I assumed wrong about this). 

If we wanted to keep the functionality to print admin boundary names with mouseover, then we will need to keep the layers selected only one at a time. However if having the dropdown is not preferred, we could make a second layer control button. I was basing this off of what is being done in the fbf maproom Aaron had sent to me.

So I guess the real question is, which is more important? Having multiple of the boundary layers showing at the same time on the map, or being able to see the boundary names when mousing over the boundary?

aaron-kaplan:
We could have a single popup that gives the names of all regions that the mouse is currently inside.

(I would actually prefer a text box somewhere outside the map to a popup/tooltip. I find that tooltips often obscure the thing I'm trying to look at.)

remicousin:
I'd think we should be able to accommodate both independently. Checkboxes to control the different layers of admin on one hand. On the other hand, infer name of the of the smallest level admin the mouse is hovering, even if that level is not a currently checked layer. In almost all admin databases, they contain the name of the geometries and the names of the higher levels they belong, so one can get the full information by querying the smallest admin level and print e.g. name3, name2, name1. There will be cases when name2, name1 are not there... We'll have to deal with these as exceptions somehow.

(The piece of code generating that will likely be reusable for the other item Aaron was mentioning: print that same information along with the graphs when clicking a  gridbox, as Ingrid/uicore does.)

As to where to print the information when mousing over... I don't know. I agree that it's not great to obstruct the map with a pop-up. On the other hand, since it's meant to help figuring out where to click, it's convenient that it would be by the mouse... Maybe the bottom right corner of the map (by the "Leaflet" mark)? or right above or right below the map?

Sorry I feel like we are often make you redo things a good deal. Maybe we should have brief discussions before starting projects... On the other hand, doing something and then change to something else means more things done, thus more learning...

drewmresnick:
This is turning into a larger change than I had anticipated, I am thinking to create an issue of it so I can work on this in a separate PR after we have discussed a bit more about how to move forward with it.

Sure, we can try to have some discussion before I work on a new component or feature, I also feel like it might be a little bit of trial and error as we build things up.. I don't mind having to change things. But next time I start conceptualizing a new feature I will bring it up in the team meeting or send an email to run it by you in the future to save some time.

remicousin:
We should probably all do that. As we all have different experience and perspective. The weekly meeting is a good place to do that, as long as weekly frequency is manageable. Let know which way you decide to go. I have also on my to-do list to not have the number of admin levels pre-determined at 3. Because Zambia has only 2. Some have 4... So we need a way to list the admin levels in the config file and have the app generate the corresponding amount of layers.
I believe this is already (partly) done in fbfmaproom (At least Aaron pointed me to something when I asked about it).
So anyway... let's inform each other of what we're up to, as those 2 things might overlap a bit.

drewmresnick:
Okay I reverted the code to have the original functionality for the boundary layers, and made a separate issue to deal with this. So this PR is just back to a few small layout / aesthetic changes that should be good to merge.

onset_maproom/layout.py
```diff
@@ -438,50 +434,57 @@ def map_layout(center_of_the_map):
...
+            # return html.Img(
+            #     style={"width": "600px"},
+            #     src=(
+            #         "https://iridl.ldeo.columbia.edu/dlcharts/render/"
+            #         "905cdac6e87a58586967e115a18e615d01530ddd?_wd=1200px&_ht=600px"
+            #         "&_langs=en&_mimetype=image%2Fpng"
+            #         "&region=bb%3A39.375%3A7.125%3A39.625%3A7.375%3Abb"
+            #         "&waterBalanceCess=3&drySpellCess=10&plotrange2=15"
+            #     ),
+            # )
```
aaron-kaplan, line 488: Are you planning on using this commented code? I would prefer to see it removed.

onset_maproom/layout.py
```diff
@@ -438,50 +434,57 @@ def map_layout(center_of_the_map):
                 id="map",
                 center=center_of_the_map,
                 zoom=CONFIG["zoom"],
+                maxBounds = [[-5.790897,25.839844],[17.978733,61.523438]],
+                minZoom = CONFIG["zoom"] - 1,
+                #maxZoom = CONFIG["zoom"] + 10, #this was completely arbitrary
```
aaron-kaplan, line 439: Remove comment.

I like this approach to constraining the zoom.

onset_maproom/maproom.py
```diff
@@ -102,33 +102,33 @@ def adm_borders(shapes):
     df["the_geom"] = df["the_geom"].apply(
         lambda x: x if isinstance(x, MultiPolygon) else MultiPolygon([x])
     )
+    index = df.index.tolist()
     shapes = df["the_geom"].apply(shapely.geometry.mapping)
+    for i in index: #this adds the district layer as a label in the dict
```
aaron-kaplan, line 107: Minor stylistic point: I'm pretty sure you could write this as
```
for i in df.index:
```
and eliminate line 105.

aaron-kaplan (approved):

onset_maproom/layout.py
remicousin: I don't know what this is. Looks like someone was hoping to print dlchart graph in the app to compare Python with Ingrid/DLchart? It can very likely be removed

onset_maproom/layout.py
```diff
@@ -438,50 +434,57 @@ def map_layout(center_of_the_map):
                 id="map",
                 center=center_of_the_map,
                 zoom=CONFIG["zoom"],
+                maxBounds = [[-5.790897,25.839844],[17.978733,61.523438]],
```
remicousin, line 437: Put these in the config file so that layout remains country agnostic.

remicousin:
I approve in spirit after the max_bounds are moved to the config file. (I don't understand how Aaron manages to approve while having unresolved conversations still on...?). I am going to be off the rest of the day so please feel free to merge after that.

PR #172, opened 2022-09-15
No vuln data in popup

aaron-kaplan:
https://trello.com/c/ytq0V45x/87-only-admin-name-on-picker-selection

While I was in there, I also fixed a couple things that have always bugged me:
- Now you can move the marker by clicking on a spot, in addition to the existing mechanism of dragging the marker.
- Added a loading indicator to the popup so it no longer displays the name of the previously-selected region while it's waiting for the callback to finish.

### DIFF
--- a/fbfmaproom/fbflayout.py
+++ b/fbfmaproom/fbflayout.py
@@ -106,7 +106,9 @@ def map_layout():
                     ),
                     dlf.Marker(
                         [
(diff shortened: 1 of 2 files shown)


kgraaf (approved):

aaron-kaplan:
Waiting for Dan's approval on trello.

Issue #145, opened 2022-07-25
Configurable number of admin levels


aaron-kaplan:
FBF config file can define an arbitrary number of admin levels. Apply the same pattern to onset maproom.

aaron-kaplan:
Unfortunately if you suppress that exception you also suppress other such exceptions that might otherwise have alerted us to real bugs. I would rather not go down that path if we can avoid it.

One callback that generates the entire LayersControl, or its whole list of children, sounds like the way to go, if that's not too hard.

remicousin:
Resolved (in our current state of affairs by [this PR](https://github.com/iridl/python-maprooms/pull/177)

PR #177, opened 2022-09-22
Remic/onset zgm

remicousin:
Moving the set up of admin layers from dummy callbacks to initialization in the layout. Still hard-coded for 2 admin. Set up for Zambia only so far. However throws out errors about connection to database.

Next step after fixing this: dynamically set an arbitrary number of layers listed in the config file.


aaron-kaplan:
I've identified the problem. While I work on fixing it, you can work around it by adding the parens back after `layout.app_layout()` in line 58.

There are two separate bugs involved. One is that when APP.layout is a function, it gets called for every single request, not only the initial page load. I'm not sure yet if that's a bug in dash or if we're using it wrong. The second bug is that when there are too many concurrent requests to the database, something isn't handling it gracefully.

We're actually lucky that we hit the first bug, because it allowed us to discover the second. Ordinarily we don't see the second bug during development, because a single user doesn't usually generate that many db queries at once, but it could happen in production when several users are visiting the site simultaneously.

remicousin:
I am making progress but I don't quite see how can create or unpack a list of unknown number of dlf.Overlay geojson objects into the dlf.LayersControl. But maybe that's because it's past 5pm...

remicousin:
Adding Drew and Kevin while Aaron is away. The initial goal of this PR was to have working config files / maprooms for eacu country. Now they are.

I was however hoping to be able to add as many administrative boundaries as there are in the config file. But I could not figure that out. So now it's hard-coded to 2 levels (Ethiopia as 3 and Senegal 4...). If someone has an idea how to do that, I'll take it.

Then rebasing on Drew's recent work made me understand better what maxBounds did to the point that it is not needed in the config file and can be inferred from the data (lat_/lon_min/max were already there).

Note that there are zarr files for all the countries with a config file now. so You can test them all.

drewmresnick:
I don't quite understand what you need help with. I see the `make_adm_overlay()` function. Are you trying to figure out a way that uses the number of admin layers and returns that many layers using the `make_adm_overlay()` function - instead of having to hard code the function for each layer?

remicousin:
Yes. The config files lists the admin layers. There can be 2, 3, or 4 in the current countries. I could not figure out how to automatically populate the LayersControl with the right number of layers. So I've set it up with 2 right now as there are always at least 2 of them with the current countries. A callback could take care of that but I think we don't want that because it's a one time set up.

I tried to use addLayer or addTo but could not make those work. Not really sure they apply here...

It is also fine to merge the PR as is and figure that out in another PR, unless someone knows better and I can implement with some guidance.

remicousin:
Let me review this now, then.

onset_maproom/layout.py
```diff
@@ -21,20 +27,73 @@
...
+    lat_min = np.around((rr_mrg.Y[0]-lat_res/2).values, decimals=10)
+    lat_max = np.around((rr_mrg.Y[-1]+lat_res/2).values, decimals=10)
     lon_res = np.around((rr_mrg.X[1]-rr_mrg.X[0]).values, decimals=10)
-    lon_min = str(np.around((rr_mrg.X[0]-lon_res/2).values, decimals=10))
-    lon_max = str(np.around((rr_mrg.X[-1]+lon_res/2).values, decimals=10))
-    lat_label = lat_min+" to "+lat_max+" by "+str(lat_res)+"˚"
-    lon_label = lon_min+" to "+lon_max+" by "+str(lon_res)+"˚"
+    lon_min = np.around((rr_mrg.X[0]-lon_res/2).values, decimals=10)
+    lon_max = np.around((rr_mrg.X[-1]+lon_res/2).values, decimals=10)
+    lat_label = str(lat_min)+" to "+str(lat_max)+" by "+str(lat_res)+"˚"
```
drewmresnick, line 95: These labels become too long to properly fit inside the control for selecting the lat/lng - is it possible to round them to a shorter number so that all ten decimal points are not displayed for the control description?

onset_maproom/config-mali-meteo.yaml
```diff
@@ -33,19 +33,19 @@ rr_mrg_chunks:
...
-shapes_adm2: select gid as key, name_2 as label, ST_AsBinary(the_geom) as the_geom
-                  from eth_zones_dd
-name_adm2: Zones
-shapes_adm3: select gid as key, name_3 as label, ST_AsBinary(the_geom) as the_geom
-                  from eth_woredas_dd
-name_adm3: Woredas
+# Mali
+logo: mali.gif
+institution: MaliMétéo
+zoom: 5
```
drewmresnick, line 39: This zoom level feels almost too zoomed in when the user opens the page; I usually would think it is better to have a more zoomed out initial view than a view that does not show all of the data, but maybe that is just a personal opinion. I like it more when the zoom is set to 4

drewmresnick:
in the middle of reviewing, but the senegal maproom does not work for me (I get a problem loading page message in my browser). Would this be a server thing or could it be an issue with some of the code?

remicousin:
The Senegal case works fine with me. But indeed, the URL is different for Senegal. Senegal to the Onset Maproom to other horizons so it's a bit different than the other countries, even though still same maproom/layout pair.

onset_maproom/layout.py
remicousin: yeah... I am not sure what to do about that... The resolution is 0.0375, so getting the edges means to add half the resolution (0.01875) to a the first and last centers that themselves can be already fairly long... (the Ghana case is an extreme case where the nc files were made on not rounded X/Y grid values...).

Maybe should get away from those labels and convey the info about valid lat/lon entries otherwise... Or just rely on the Alerts that are in place.

drewmresnick: If we need to display the values with accuracy and thus all of the decimal places I would say moving them to the description of the maproom or maybe displaying them when hovering over the input box title?

remicousin: I made this an issue as this is not related to this PR.

drewmresnick (approved):

remicousin:
Thanks. I am going to rebase this and see what happens as Aaron's queuepool change likely has an impact here. And make sure he approves then.

remicousin:
@aaron-kaplan , I rebased and tested this and hopefully all ok. There are several things going in this PR. One major and a few minor.

The major was to get rid of the "dummy" callbacks and have the tools deal with unknown number of admin layers. Result: I don't know anymore if the dummy callback are not welcome, so that change might have turned completely irrelevant; and I could not figure out how to automatize the number of layers so now it's the lowest common denominator hard-coded for all countries: 2 layers.
Corollary change are in the config files with cleaner layers set up.

Minor changes are:
-- other cosmetics (logo, etc) so that all countries config files work
-- automation of Drew's maxBounds entry (no more in config but inferred by layout)

So at this point, please let know if it can go as is, or need to go back some.

onset_maproom/config-mali-meteo.yaml
aaron-kaplan: Created https://github.com/iridl/python-maprooms/issues/184 "Automatically choose initial zoom level that fits region of interest in window."

onset_maproom/layout.py
```diff
@@ -22,19 +29,67 @@
...
+    for i in df.index: #this adds the district layer as a label in the dict
+        shapes[i]['label'] = df['label'][i]
+    return {"features": shapes}
+
+
+def make_adm_overlay(adm_list, adm_lev, adm_weight):
+    border_id = f'borders_adm{adm_lev}'
+    adm_check = False
+    if adm_lev == 1:
+        adm_check = True
```
aaron-kaplan, line 64: I suggest making `is_checked` a parameter of this function so it can be used with different default layers in different contexts.

aaron-kaplan (changes requested):
Sorry about this, but I would prefer to go back to using the dummy callbacks until we resolve the issue with the layout function being called too many times. In the current state, shapes get loaded from the db at import time, which I really want to avoid.

I commented on https://github.com/iridl/python-maprooms/issues/145 about how to make the number of admin levels configurable. I'm fine with putting that off until later.

onset_maproom/layout.py
```diff
@@ -22,19 +29,67 @@
...
+
 def app_layout():
 
     # Initialization
     rr_mrg = calc.read_zarr_data(RR_MRG_ZARR)
     center_of_the_map = [((rr_mrg.Y[0]+rr_mrg.Y[-1])/2).values, ((rr_mrg.X[0]+rr_mrg.X[-1])/2).values]
     lat_res = np.around((rr_mrg.Y[1]-rr_mrg.Y[0]).values, decimals=10)
-    lat_min = str(np.around((rr_mrg.Y[0]-lat_res/2).values, decimals=10))
-    lat_max = str(np.around((rr_mrg.Y[-1]+lat_res/2).values, decimals=10))
+    lat_min = np.around((rr_mrg.Y[0]-lat_res/2).values, decimals=10)
```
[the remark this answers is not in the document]
remicousin: It seems to be needed to avoid printing machine error values. See screenshot of an example. 10 is probably arbitrary. Might be needed for very high resolution stuff... That's probably why I went so far. Maybe there is a better way to go around the machine error problem?

Screen Shot 2022-09-29 at 11 54 06 AM

remicousin:
ok. That's what I got.

aaron-kaplan (approved):

PR #178, opened 2022-09-22
Change lead time representation in subs maproom

drewmresnick:
Based off of the email thread regarding lead times, I gathered that it did not matter directly how we represented the leads at least within the data itself, so my approach was:

Use the lead time representation and map those to some arbitrary lead time values that the user can configure both of in the configuration file. I tried to make it as flexible as possible to apply to other datasets.

Then in the maproom itself it allows the user to select from target date ranges.

### DIFF
--- a/subseas_flex_fcst/config-sample.yaml
+++ b/subseas_flex_fcst/config-sample.yaml
@@ -18,6 +18,7 @@ hcst_filePattern: CFSv2_SubXPRCP_CCAFCST_xvPr_Apr_*.txt
 variable: Precipitation Anomaly
 
 #S,L, target dates options
+target_period_length: 7
 leadInit: wk1
 leads: #provider_ID:leadTime_value
     wk1: 1
--- a/subseas_flex_fcst/config-sample.yaml
+++ b/subseas_flex_fcst/config-sample.yaml
@@ -16,6 +16,17 @@ forecast_var_filePattern: CFSv2_SubXPRCP_CCAFCST_var_Apr_*.txt
 obs_filePattern: CFSv2_SubXPRCP_CCAFCST_obs_Apr_*.txt
 hcst_filePattern: CFSv2_SubXPRCP_CCAFCST_xvPr_Apr_*.txt
 variable: Precipitation Anomaly
+
+#S,L, target dates options
+numberLeads: 4
+dataLead1: wk1
+lead1: 1
+dataLead2: wk2
+lead2: 8
+dataLead3: wk3
+lead3: 15
+dataLead4: wk4
+lead4: 22
 target_period_length: 7
 
 # Viz' set up
--- a/subseas_flex_fcst/config-sample.yaml
+++ b/subseas_flex_fcst/config-sample.yaml
@@ -18,16 +18,12 @@ hcst_filePattern: CFSv2_SubXPRCP_CCAFCST_xvPr_Apr_*.txt
 variable: Precipitation Anomaly
 
 #S,L, target dates options
-numberLeads: 4
-dataLead1: wk1
-lead1: 1
-dataLead2: wk2
-lead2: 8
-dataLead3: wk3
-lead3: 15
-dataLead4: wk4
-lead4: 22
-target_period_length: 7
+leadInit: wk1
+leads: #provider_ID:leadTime_value
+    wk1: 1
+    wk2: 8
+    wk3: 15
+    wk4: 22
 
 # Viz' set up
 zoom: 7
--- a/subseas_flex_fcst/config-sample.yaml
+++ b/subseas_flex_fcst/config-sample.yaml
@@ -19,7 +19,6 @@ variable: Precipitation Anomaly
 
 #S,L, target dates options
 target_period_length: 7
-leadInit: wk1
 leads: #provider_ID:leadTime_value
     wk1: 1
     wk2: 8
--- a/subseas_flex_fcst/layout.py
+++ b/subseas_flex_fcst/layout.py
@@ -273,7 +273,7 @@ def navbar_layout(phys_units):
                         options = startDates,
                         value=startDates[-1],
                     ),
-                ],style={"width":"8%"},
+                ],style={"width":"9%","font-size":".9vw","text-align":"center"},
             ),
             html.Div(
                 [
@@ -298,9 +298,9 @@ def navbar_layout(phys_units):
                         id="leadTime",
                         clearable=False,
                         options=[],
-                        value="wk1",
+                        value=CONFIG["leadInit"],
                     ),
-                ],style={"width":"9%"},
+                ],style={"width":"12%","font-size":".9vw","text-align":"center"},
             ),
             dbc.Alert(
                 "Please type-in a threshold for probability of non-/exceeding",
--- a/subseas_flex_fcst/layout.py
+++ b/subseas_flex_fcst/layout.py
@@ -253,7 +253,7 @@ def navbar_layout(phys_units):
                     help_layout(
                         "Start Date",
                         "start_date",
-                        "Start date choices",
+                        "Model start dates",
                     ),
                 ],
                 style={
@@ -278,15 +278,15 @@ def navbar_layout(phys_units):
             html.Div(
                 [
                     help_layout(
-                        "Lead time",
+                        "Target Date",
                         "lead_time",
-                        "Lead time range choices, in weeks from the start date",
+                        "Target date ranges calculated using lead times and model starts.",
                     ),
                 ],
                 style={
                     "color": "white",
                     "position": "relative",
-                    "width": "105px",
+                    "width": "115px",
                     "display": "inline-block",
                     "padding": "10px",
                     "vertical-align": "top",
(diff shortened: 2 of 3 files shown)


subseas_flex_fcst/config-sample.yaml
```diff
@@ -16,6 +16,17 @@ forecast_var_filePattern: CFSv2_SubXPRCP_CCAFCST_var_Apr_*.txt
 obs_filePattern: CFSv2_SubXPRCP_CCAFCST_obs_Apr_*.txt
 hcst_filePattern: CFSv2_SubXPRCP_CCAFCST_xvPr_Apr_*.txt
 variable: Precipitation Anomaly
+
```
drewmresnick, line 19: Would appreciate some help on better naming conventions here; my idea was that `dataLead` represents the data provider's lead representation, while `lead` represents how we use lead time to calculate the target dates

subseas_flex_fcst/layout.py
```diff
@@ -278,15 +278,15 @@ def navbar_layout(phys_units):
             html.Div(
                 [
                     help_layout(
-                        "Lead time",
+                        "Target Date",
                         "lead_time",
-                        "Lead time range choices, in weeks from the start date",
+                        "Target date ranges calculated using lead times and model starts.",
```
drewmresnick, line 283: Also this might have a better description than I wrote..

subseas_flex_fcst/maproom.py
```diff
@@ -93,6 +93,28 @@ def display_relevant_control(variable):
...
+@APP.callback(
+    Output("leadTime","options"),
+    Output("leadTime","value"),
+    Input("startDate","value"),
+)
+def targetStartOptions(startDate):
+    leadsDic = {}
+    for x in range(1,CONFIG["numberLeads"]+1):
+        dict = {}
+        dict["provider"] = CONFIG[f"dataLead{x}"]
```
drewmresnick, line 105: Here also the dictionary index names could be improved; currently `provider` is the provider's lead time rep, while `numVal` is the numerical value we use for calculations

subseas_flex_fcst/layout.py
remicousin: Maybe: "Time period being forecasted"
Maproom text can elaborate on the relationship between S/L/target if need be.

subseas_flex_fcst/layout.py
```diff
@@ -253,7 +253,7 @@ def navbar_layout(phys_units):
                     help_layout(
                         "Start Date",
                         "start_date",
-                        "Start date choices",
+                        "Model start dates",
```
remicousin, line 256: increase the Start Date box width so the date can fit in full.

subseas_flex_fcst/layout.py
```diff
@@ -278,15 +278,15 @@ def navbar_layout(phys_units):
...
                         "lead_time",
-                        "Lead time range choices, in weeks from the start date",
+                        "Target date ranges calculated using lead times and model starts.",
                     ),
                 ],
                 style={
                     "color": "white",
                     "position": "relative",
-                    "width": "105px",
+                    "width": "115px",
```
remicousin, line 289: Increase width so that target date fits in full. Also the target date needs to contain the year information.

subseas_flex_fcst/config-sample.yaml
```diff
@@ -16,6 +16,17 @@ forecast_var_filePattern: CFSv2_SubXPRCP_CCAFCST_var_Apr_*.txt
...
+#S,L, target dates options
+numberLeads: 4
+dataLead1: wk1
+lead1: 1
+dataLead2: wk2
+lead2: 8
+dataLead3: wk3
+lead3: 15
+dataLead4: wk4
+lead4: 22
```
remicousin, line 29: could this be a dictionary in the first place? or whatever nested structured are called in config files? And that might save you from needing a numberLeads that will be the size of that list.

remicousin (commented):
A superficial review for now. I'll have to look at it in more details later.

subseas_flex_fcst/config-sample.yaml
aaron-kaplan:
```
leads:
    wk1: 1
    wk2: 8
    wk3: 15
    wk4: 22
```

subseas_flex_fcst/layout.py
drewmresnick: For these issues I was trying to figure out a way to make the size of the text or the size of the div change to accommodate changing the window size, so that they would always show the  full dates. Not so much luck on that though.

remicousin: That would be nice but maybe not worth sweating too much over it. What could be nice is to have some clean code to print such dates range in a more concise manner when can be. e.g.
8 Apr 2022 - 14 Apr 2022
could be
8-14 Apr 2022.
Whereas 31 Dec 2022 - 6 Jan 2023 can only keep that format...
That would mean checking if the years are equal, then if the month are equals, and append the date string accordingly.

As for saving space, I can think of 2 ways:
1/ One could be to apply slightly smaller font size.
2/ Another could be to create some menu widget like uicore does, where the title of the control is put on top of it
http://iridl.ldeo.columbia.edu/maproom/Global/ForecastsS2S/precip_median_subx.html?S=0000%2023%20Sep%202022
Or a combination of both.

I would not make those a priority unless it appears to be trivial.

subseas_flex_fcst/maproom.py
remicousin: Aaron provided an example on how to write the config file. This should bear on this code so I'll review after you had a chance to revisit it.

subseas_flex_fcst/config-sample.yaml
```diff
@@ -18,16 +18,12 @@ hcst_filePattern: CFSv2_SubXPRCP_CCAFCST_xvPr_Apr_*.txt
...
-dataLead1: wk1
-lead1: 1
-dataLead2: wk2
-lead2: 8
-dataLead3: wk3
-lead3: 15
-dataLead4: wk4
-lead4: 22
-target_period_length: 7
+leadInit: wk1
```
remicousin, line 22: probably safe to assume that taking the first lead by default will be ok in all circumstances. In which case can get rid of that config entry.

subseas_flex_fcst/layout.py
drewmresnick: Okay I just pushed a commit that prints the date ranges more concisely. I am not seeing how putting the title of the control over the top will impact the display of the dates within the dropdown; if this is an aesthetic change you might like to have done still I can make it into an issue to address later.

drewmresnick: Okay should I make an issue for this then or is it not much worth thinking about in general?

remicousin: I think we can let go. I am sure there will be plenty of opportunities to remind us we can do much better aesthetically.... :)

drewmresnick: Okay. Then I think this PR is ready for approval unless any other things are found by you are Aaron to be changed.

drewmresnick:
@aaron-kaplan @remicousin Can I get one of your approvals to merge this if it looks good?

remicousin:
Will look at it now.

subseas_flex_fcst/maproom.py
```diff
@@ -94,6 +94,33 @@ def display_relevant_control(variable):
...
+    Output("leadTime","value"),
+    Input("startDate","value"),
+)
+def targetStartOptions(startDate):
+    leadsValues = list(CONFIG["leads"].values())
+    leadsKeys = list(CONFIG["leads"])
+    startDate = pd.to_datetime(startDate)
+    optionsDict = {}
+    for idx, x in enumerate(leadsValues):
+        endDate = x+CONFIG["target_period_length"]-1 #used to calculate the target end date (targetEnd = lead + period length - 1)
```
remicousin, line 108: I would find simpler to read this if the date for the target start and the date for the target end were computed first once and for all (for each loop iteration). Then you run the tests to decide the format of the target start date. Note that the format of the target end date is always the same so need to include it in the test.
That would all save some lines.

remicousin (approved):
I approve in spite of my last comment, so that you can merge now or after acting upon it as you wish.
But you might want to wait for Aaron's review: he's the Python expert.

subseas_flex_fcst/maproom.py
aaron-kaplan: I agree. It took me several minutes to figure out what was intended here; I think it would have been easier if the calculation were separate from the formatting.

Incidentally, I also find the term "issue date" clearer than "start date" because it eliminates the ambiguity/overloading of "start" to mean either S (issue date) or S + L (start of target period).

remicousin: if we are it... I didn't want to load the boat (and I might be contradicting myself from the past here) but let's use the terms already in use (they probably what they are now for a good reason)

http://iridl.ldeo.columbia.edu/maproom/Global/ForecastsS2S/precip_flex_subx_weekly.html?S=0000%2023%20Sep%202022

I mean for the controls:
"Forecast Issued"
"Target Period"

drewmresnick: Okay. Should I be changing the variable names throughout the code to reflect this?

remicousin:
nice

subseas_flex_fcst/maproom.py
drewmresnick: @remicousin Let me know if you would like me to make a new issue for changing variable names to reflect this difference, or if you just wanted it applied to the controls names.

remicousin: Hopefully soon, we will have to generalize a good deal of the things here to work with seasonal forecast. That will be an occasion to rationalize names, and possibly easier to do or think of as we will have a new case to run this all through.

drewmresnick: Okay then I will leave that for now.

PR #180, opened 2022-09-23
Add the southern oromia project region

kgraaf:


fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -686,6 +686,142 @@ countries:
...
+        marker:  [40.0, 5.7]
+        origin: [.125, .125]
+        onload_warning: *warn
+        shapes:
+            - name: Project Region
+              sql: select id_1 as key, 'Southern Oromia' as label, ST_AsBinary(ST_Union(the_geom)) as the_geom
+                  from iridb.eth_zones_dd
+                  where name_2 in ('East Shewa', 'Arsi', 'West Arsi', 'West Haraghe', 'East Haraghe', 'Bale', 'East Bale', 'Guji', 'Borena')
+                  group by 1, 2
+              vuln_sql: select cast(null as int) as key, 0 as year, 0 as vuln where 1 = 2
```
aaron-kaplan, line 704: What does `cast(null as int)` do?

fbfmaproom/fbfmaproom.py
```diff
@@ -383,6 +383,8 @@ def get_mpoly(mode, country_key, geom_key):
         mpolygon = MultiPolygon([Polygon([(x0, y0), (x0, y1), (x1, y1), (x1, y0)])])
     else:
         label, mpolygon = retrieve_geometry2(country_key, int(mode), geom_key)
+        if isinstance(mpolygon, Polygon):
```
kgraaf, line 386: Something that has been vexing me is why this is necessary at all, and why Madagascar works without it. What's weird is that there's two different pathways that shapes go through, and the first already did something like this, but the second (this one) did not

kgraaf: Well let me clarify: "mechanically" I know why it's necessary, it's because Polygons aren't iterable (of course). However, I just don't understand what's different about this circumstance such that we get a MultiPolygon from the query in all other circumstances, including another "computed" top-level boundary (in the case of Madagascar)

fbfmaproom/fbfmaproom.py
```diff
@@ -383,6 +383,8 @@ def get_mpoly(mode, country_key, geom_key):
...
+            mpolygon = MultiPolygon([mpolygon])
```
aaron-kaplan, line 387: Do this in retrieve_geometry2, so we don't have to remember to do it every time we call that function.

aaron-kaplan (commented):
Looks good. I believe we need `southern-oromia-ond` as well, but we can wait for comment from FIST on this version before doing that copy-paste job.

fbfmaproom/fbfmaproom.py
aaron-kaplan: Yeah, I don't know either. But we had this check elsewhere already (see retrieve_vulneraiblity), so it's not just Oromia.

fbfmaproom/fbfmaproom-sample.yaml
kgraaf: in SQL, nulls are typed, but they're typed as strings by default, so this gives you (as you might expect) an int-typed null. If you don't put this cast in it throws an error. 

That being said, I'm kinda thinking we should modify retrieve_vuln such that if there's no vuln_sql field it just doesn't do the join. Having to copy past this same code everywhere is annoying and confusing

fbfmaproom/fbfmaproom.py
kgraaf: Yeah I noticed that too. I'm very curious why this is only showing up now but I suppose it doesn't matter

PR #181, opened 2022-09-23
Remove queuepool

aaron-kaplan:
This is for @kgraaf to review; tagging the others as an FYI.

Igor's queuepool library now has three strikes against it:
- When the database server goes down for a while, the pool never recovers, even after the database comes back on line. We have to manually restart all of the python services to get them back.
- It breaks under heavy concurrent load. We get error messages like `psycopg2.OperationalError: lost synchronization with server: got message type "3", length 1630681443`. (This showed up as a side-effect of a bug that caused multiple db queries to be issued on each tile request.)
- It's not packaged for conda, so we have to install it with pip.

All of these problems could surely be fixed, but rather than spending time on that, I would prefer to switch to SQLAlchemy, which is the de facto standard in the python community and presumably very mature and well tested.

Igor's [justification](https://pypi.org/project/queuepool/) for writing his own connection pooling library is based on deficiencies of psycopg2's connection pool that SQLAlchemy's implementation doesn't share. SQLAlchemy has an implementation (not coincidentally named [`QueuePool`](https://docs-sqlalchemy.readthedocs.io/ko/latest/core/pooling.html#sqlalchemy.pool.QueuePool)) that blocks incoming requests instead of rejecting them when the pool is exhausted, and that does connection recycling.

Note that this PR doesn't replace the connection pool implementation, it just eliminates it. There's no discernable difference in performance with a single user. It might make a difference under load, but we don't have much load at the moment. I do intend to replace it at some point, but I don't think it's urgent.

### DIFF
--- a/fbfmaproom/environment.yml
+++ b/fbfmaproom/environment.yml
@@ -8,16 +8,18 @@ dependencies:
     # https://github.com/rasterio/rasterio/issues/2333
   - python=3.9
   - cftime
-  - dash
+
+  # some later version removes the __wrapped__ attribute on callbacks,
+  # which is used by some tests
+  - dash=2.3.1
+
   - dash-leaflet
   - dash-bootstrap-components
   - numpy
   - opencv
 
-    # Newer versions of Pandas do not appear to work correctly with queuepool.
-    # In particular, how that package forms SQL query objects needs to be updated. 
-    # Hopefully the solution here is to either send Igor a issue or to migrate to
-    # a different connection pooling library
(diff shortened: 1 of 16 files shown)


aaron-kaplan:
Looks like I forgot to regenerate some of the environment_linux.yml. I'll do that now.

kgraaf (approved):

aaron-kaplan:
Rebuilt the conda environments. I don't feel like it would be a good idea to merge this and then go away for four days, so I'll leave it unmerged until I get back.

PR #183, opened 2022-09-28
subs fcst map title

drewmresnick:
Making a map title. I put it on top of the map which felt intuitive to me but I know the onset maproom has it on the bottom..

### DIFF
--- a/subseas_flex_fcst/maproom.py
+++ b/subseas_flex_fcst/maproom.py
@@ -121,15 +121,14 @@ def targetStartOptions(startDate):
     return optionsDict, leadsKeys[0]
 
 @APP.callback(
-   Output("map_title","children"),
-   Input("variable","value"),
+   Output("mapTitle","children"),
    Input("startDate","value"),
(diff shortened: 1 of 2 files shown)


subseas_flex_fcst/maproom.py
```diff
@@ -120,6 +120,16 @@ def targetStartOptions(startDate):
...
+@APP.callback(
+   Output("map_title","children"),
+   Input("variable","value"),
+   Input("startDate","value"),
+   Input("leadTime","value"),
+   Input("leadTime","options"),
+)
+def write_map_title(variable, startDate, leadTime, leadTimeOptions):
+    targetPeriod = leadTimeOptions.get(leadTime)
+    return f"{targetPeriod} {variable} Forecast issued {startDate}"
```
remicousin, line 132: For "variable," I meant to use what's in the config file which is "Precipitation Anomaly" in that case.

remicousin (approved):
I approve so that you can merge when variable is switched.

PR #185, opened 2022-09-29
callback pathway for location and their graphs

remicousin:
This is an attempts, applied to forecast maproom for now, to cleanly layout callback that take care of location selection, then callback that take location to make a local graph (or return an error if outside domain or no data).

Points of concern:

1. I am really not sure I am doing the right thing to get the Marker position back
2. This seems to force LayerGroup to be for the Marker and only the Marker
3. Once cleaned up, it might make sense to factor things out for a generalized approach to apply to all maprooms

### DIFF
--- a/subseas_flex_fcst/layout.py
+++ b/subseas_flex_fcst/layout.py
@@ -28,7 +28,7 @@ def app_layout():
 
     # Initialization
     fcst_mu = predictions.sel_cpt_file(DATA_PATH,CONFIG["forecast_mu_file_pattern"],list(CONFIG["leads"])[0],startDates[-1])
-    center_of_the_map = [((fcst_mu.Y[0]+fcst_mu.Y[-1])/2).values, ((fcst_mu.X[0]+fcst_mu.X[-1])/2).values]
+    center_of_the_map = [((fcst_mu.Y[int(fcst_mu.Y.size/2)].values)), ((fcst_mu.X[int(fcst_mu.X.size/2)].values))]
     lat_res = (fcst_mu.Y[0]-fcst_mu.Y[1]).values
     lat_min = str((fcst_mu.Y[-1]-lat_res/2).values)
     lat_max = str((fcst_mu.Y[0]+lat_res/2).values)
@@ -412,7 +412,10 @@ def map_layout(center_of_the_map):
                         position="topleft",
                         id="layers_control",
                     ),
-                    dlf.LayerGroup(id="layers_group"),
+                    dlf.LayerGroup(
+                        [dlf.Marker(id="loc_marker", position=center_of_the_map)],
+                        id="layers_group"
+                    ),
                     dlf.ScaleControl(imperial=False, position="bottomleft"),
                     dlf.Colorbar(
                         id="fcst_colorbar",
--- a/subseas_flex_fcst/layout.py
+++ b/subseas_flex_fcst/layout.py
@@ -28,13 +28,13 @@ def app_layout():
 
     # Initialization
     fcst_mu = predictions.sel_cpt_file(DATA_PATH,CONFIG["forecast_mu_file_pattern"],list(CONFIG["leads"])[0],startDates[-1])
-    center_of_the_map = [((fcst_mu.Y[int(fcst_mu.Y.size/2)].values)), ((fcst_mu.X[int(fcst_mu.X.size/2)].values))]
-    lat_res = (fcst_mu.Y[0]-fcst_mu.Y[1]).values
-    lat_min = str((fcst_mu.Y[-1]-lat_res/2).values)
-    lat_max = str((fcst_mu.Y[0]+lat_res/2).values)
-    lon_res = (fcst_mu.X[1]-fcst_mu.X[0]).values
-    lon_min = str((fcst_mu.X[0]-lon_res/2).values)
-    lon_max = str((fcst_mu.X[-1]+lon_res/2).values)
+    center_of_the_map = [((fcst_mu["Y"][int(fcst_mu["Y"].size/2)].values)), ((fcst_mu["X"][int(fcst_mu["X"].size/2)].values))]
+    lat_res = (fcst_mu["Y"][0]-fcst_mu["Y"][1]).values
+    lat_min = str((fcst_mu["Y"][-1]-lat_res/2).values)
+    lat_max = str((fcst_mu["Y"][0]+lat_res/2).values)
+    lon_res = (fcst_mu["X"][1]-fcst_mu["X"][0]).values
+    lon_min = str((fcst_mu["X"][0]-lon_res/2).values)
+    lon_max = str((fcst_mu["X"][-1]+lon_res/2).values)
     lat_label = lat_min+" to "+lat_max+" by "+str(lat_res)+"˚"
     lon_label = lon_min+" to "+lon_max+" by "+str(lon_res)+"˚"
     fcst_mu_name = list(fcst_mu.data_vars)[0]
--- a/subseas_flex_fcst/maproom.py
+++ b/subseas_flex_fcst/maproom.py
@@ -117,6 +117,13 @@ def write_map_title(start_date, lead_time, lead_time_options):
     target_period = lead_time_options.get(lead_time)
     return f'{target_period} {CONFIG["variable"]} Forecast issued {start_date}'
 
+
+def sel_center(spatial_array, lat, lng, dim_y="Y", dim_x="X"):
+    half_res = (spatial_array[dim_y][1] - spatial_array[dim_x][0]) / 2
+    tol = np.sqrt(2 * np.square(half_res)).values
+    return spatial_array.sel(method="nearest", tolerance=tol, **{dim_x:lng, dim_y:lat})
+    
+
 @APP.callback(
     Output("loc_marker", "position"),
     Output("lat_input", "value"),
@@ -147,10 +154,8 @@ def pick_location(n_clicks, click_lat_lng, latitude, longitude):
         else:
             lat = latitude
             lng = longitude
-        half_res = (fcst_mu["X"][1] - fcst_mu["X"][0]) / 2
-        tol = np.sqrt(2 * np.square(half_res)).values
         try:
-            nearest_grid = fcst_mu.sel(X=lng, Y=lat, method="nearest", tolerance=tol)
+            nearest_grid = sel_center(fcst_mu, lat, lng)
             lat = nearest_grid.Y.values
             lng = nearest_grid.X.values
         except KeyError:
@@ -173,14 +178,12 @@ def local_plots(marker_pos, startDate, leadTime):
     fcst_mu, fcst_var, obs, hcst = read_cptdataset(leadTime, startDate, y_transform=CONFIG["y_transform"])
     # Errors handling
     try:
-        half_res = (fcst_mu["X"][1] - fcst_mu["X"][0]) / 2
-        tol = np.sqrt(2 * np.square(half_res)).values
-        fcst_mu = fcst_mu.sel(X=lng, Y=lat, method="nearest", tolerance=tol)
-        fcst_var = fcst_var.sel(X=lng, Y=lat, method="nearest", tolerance=tol)
-        obs = obs.sel(X=lng, Y=lat, method="nearest", tolerance=tol)
+        fcst_mu = sel_center(fcst_mu, lat, lng)
+        fcst_var = sel_center(fcst_var, lat, lng)
+        obs = sel_center(obs, lat, lng)
         isnan = np.isnan(fcst_mu).sum() + np.isnan(obs).sum()
         if CONFIG["y_transform"]:
-            hcst = hcst.sel(X=lng, Y=lat, method="nearest", tolerance=tol)
+            hcst = sel_center(hcst, lat, lng)
             isnan_yt = np.isnan(hcst).sum()
             isnan = isnan + isnan_yt
         if isnan > 0:
--- a/subseas_flex_fcst/maproom.py
+++ b/subseas_flex_fcst/maproom.py
@@ -118,7 +118,7 @@ def write_map_title(start_date, lead_time, lead_time_options):
     return f'{target_period} {CONFIG["variable"]} Forecast issued {start_date}'
 
 @APP.callback(
-    Output("layers_group", "children"),
+    Output("loc_marker", "position"),
     Output("lat_input", "value"),
     Output("lng_input", "value"),
     Input("submit_lat_lng","n_clicks"),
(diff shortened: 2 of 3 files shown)


subseas_flex_fcst/maproom.py
```diff
@@ -131,54 +131,75 @@ def write_mapTitle(startDate, leadTime, leadTimeOptions):
...
+    except KeyError:
+        lat = lat
+        lng = lng
+    return dlf.Marker(position=[lat, lng]), lat, lng, None
+
+
+@APP.callback(
+    Output("cdf_graph", "figure"),
+    Output("pdf_graph", "figure"),
+    Input("layers_group", "children"),
```
aaron-kaplan, line 178: Can we put the `Marker` in the layout, and have `pick_location` just set the Marker's position instead of generating the Marker itself? Then this input, which seems strange and brittle to me (couldn't `layers_group` have other children in the future?), could be replaced with that position.

subseas_flex_fcst/maproom.py
```diff
@@ -131,54 +131,75 @@ def write_mapTitle(startDate, leadTime, leadTimeOptions):
     return f'{targetPeriod} {CONFIG["variable"]} Forecast issued {startDate}'
 
 @APP.callback(
-    Output("cdf_graph", "figure"),
-    Output("pdf_graph", "figure"),
-    Output("map", "click_lat_lng"),
     Output("layers_group", "children"),
     Output("latInput", "value"),
     Output("lngInput", "value"),
+    Output("map", "click_lat_lng"),
```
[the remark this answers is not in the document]
remicousin: I did say I would clean up, but it turns out that it seems convenient to return click_lat_lng to None, so that the function can easily know if it's getting new info from a click, or from a submitted button.

subseas_flex_fcst/maproom.py
remicousin: yes, agreed. I was concerned that layers_group was sort of blocked for this one purpose.

but when I try to have a dlf.Marker in the layout and pass it "position" I get:

TypeError: Required argument `position` was not specified.

when I start to think I understood something... Maybe I'll try again freshly tomorrow morning

aaron-kaplan: OK, we can come back to this after (hopefully) fixing the problem with the deferred call to the layout function.

subseas_flex_fcst/maproom.py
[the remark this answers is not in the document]
aaron-kaplan: It was my understanding that click_lat_lng would always be None if the callback wasn't triggered by a map click. Was I wrong about that?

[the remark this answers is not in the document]
aaron-kaplan: Thank you for correcting my misunderstanding.

It seems strange to have click_lat_lng as both an input and an output--I expected that dash would complain about a circular dependency. But apparently it works ok.

That said, I did some reading and it looks like the way you're supposed to figure out which control triggered the callback is by using `dash.ctx.triggered_id` ([doc](https://dash.plotly.com/dash-html-components/button#determining-which-button-changed-with-dash.ctx)). I tried printing the value of that variable inside `pick_location` and it prints either `None` (on the initial load), `map`, or `submitLatLng`, depending on which control I used. If you use that then you won't need to set click_lat_lng to None.

aaron-kaplan (approved):
I think it would be best to switch to using `dash.ctx.triggered_id` as I commented above, but it's not critical and this otherwise looks good, so I'm approving.

remicousin:
I've just pushed something that revisits the process using ctx.trigger. Have a look.

As for the matter of taking dlf.Marker out of the CB and just return position values, let me know if you think it's probably just me being dumb and not knowing how to use CB properly just yet, and I'll look further into it.

Also, did this:
 lat = marker["props"]["position"][0]
    lng = marker["props"]["position"][1]
made sense to you? I had to print things to figure out what I could call from layers_group...

aaron-kaplan:
Argh, I I just left comments on a commit rather than on the PR, and they're not showing up here.

https://github.com/iridl/python-maprooms/commit/bb4297e6080e5b0e7322849e3fa025097316f1c4

subseas_flex_fcst/maproom.py
```diff
@@ -131,40 +131,63 @@ def write_mapTitle(startDate, leadTime, leadTimeOptions):
...
+        half_res = (fcst_mu["X"][1] - fcst_mu["X"][0]) / 2
+        tol = np.sqrt(2 * np.square(half_res)).values
+        try:
+            nearest_grid = fcst_mu.sel(X=lng, Y=lat, method="nearest", tolerance=tol)
+            lat = nearest_grid.Y.values
+            lng = nearest_grid.X.values
+        except KeyError:
+            lat = lat
+            lng = lng
+    return dlf.Marker(position=[lat, lng]), lat, lng
```
aaron-kaplan, line 172: Yes, this (switching on dash.ctx.triggered_id) feels better to me.

Reducing the level of if/then/else nesting would make it a bit easier to read.
```
if dash.ctx.triggered_id == None:
    ...
elif dash.ctx.triggered_id == "map":
    ...
else:
    ...
```

subseas_flex_fcst/maproom.py
```diff
@@ -131,40 +131,63 @@ def write_mapTitle(startDate, leadTime, leadTimeOptions):
...
+    fcst_mu, fcst_var, obs, hcst = read_cptdataset(leadTime, startDate, y_transform=CONFIG["y_transform"])
     # Errors handling
     try:
         half_res = (fcst_mu["X"][1] - fcst_mu["X"][0]) / 2
         tol = np.sqrt(2 * np.square(half_res)).values
-        nearest_grid = fcst_mu.sel(X=lng, Y=lat, method="nearest", tolerance=tol)
-        lat = nearest_grid.Y.values
-        lng = nearest_grid.X.values
         fcst_mu = fcst_mu.sel(X=lng, Y=lat, method="nearest", tolerance=tol)
         fcst_var = fcst_var.sel(X=lng, Y=lat, method="nearest", tolerance=tol)
```
aaron-kaplan, line 192: We already snapped the selection to the grid when setting the marker position, right? So can't we just do `.sel` without a tolerance here?

If we do need to re-snap it here, then factor out the logic that this function and pick_location have in common.

aaron-kaplan:
> Also, did this:
> lat = marker["props"]["position"][0]
> lng = marker["props"]["position"][1]
> made sense to you? I had to print things to figure out what I could call from layers_group...

I don't love it, but at the moment I don't know what else we could do. I think the cleanest solution will be to modify dlf.Marker such that if position is undefined it just makes it invisible, rather than throwing an error. Probably not a hard change to make, but I'll postpone looking into it until we figure out the right approach for generating the initial layout.

subseas_flex_fcst/maproom.py
remicousin: hmmm... maybe... If trigger is None, I don't do snapping because it's already done by picking the default central point.

subseas_flex_fcst/maproom.py
remicousin: I don't snap when the click/input is outside the data domain. When outside the data domain, pick_location returns whatever lon/lat was clicked/input   and local_plots triggers the errors. So either we rethink of the behavior of outside domain click/input (I don't have an idea right now; either I factor out, as you said.

subseas_flex_fcst/maproom.py
aaron-kaplan: Oh, yeah. You could do the snapping in all cases, just to make it simpler (it does no harm to re-snap a pre-snapped location), but leaving it as it is would also be ok with me.

subseas_flex_fcst/maproom.py
aaron-kaplan: Interesting--the same `.sel` operation does double duty: in one place it snaps the location to the grid, and in the other place it checks whether the location is in or out of bounds. 

We could change the behavior of clicking out of bounds--instead of moving the marker and displaying the error charts, we could just refuse to move the marker there. Not sure, that might be confusing/look like a bug.

Probably just factor it out.

remicousin: We could that. That is if click/input out of bounds: do nothing. We could also snap to the closest location still, that would be on the edge of the domain. It might look funny to a user... on the other hand there won't be out of bounds error graphs anymore.

In the future, we might have apps with layers of different resolution (e.g. grids, geometries, points), where we would have more complex behavior, e.g. snap to grid and get closest point to overlay time series from both (that might be the case already in DBflood...). So maybe it's safer to keep the current behavior... thinking out loud now.

subseas_flex_fcst/maproom.py
aaron-kaplan: Looking at this with fresh eyes, this still makes me uncomfortable. Pending a change to the Marker class that allows it to be created with no position, how about we use the center of the dataset as the initial position?

remicousin: I've just pushed 2 commits, the first one is trying to do that. Not sure I got what you had in mind so please have a look.

subseas_flex_fcst/maproom.py
remicousin: I've just pushed 2 commits, the second one is trying to do that. Not sure I got what you had in mind so please have a look.

subseas_flex_fcst/maproom.py
```diff
@@ -130,55 +130,81 @@ def write_mapTitle(startDate, leadTime, leadTimeOptions):
     targetPeriod = leadTimeOptions.get(leadTime)
     return f'{targetPeriod} {CONFIG["variable"]} Forecast issued {startDate}'
 
+
+def sel_center(spatial_array, lat, lng, dim_y="Y", dim_x="X"):
+    half_res = (spatial_array[dim_y][1] - spatial_array[dim_x][0]) / 2
+    tol = np.sqrt(2 * np.square(half_res)).values
+    return spatial_array.sel(method="nearest", tolerance=tol, **{dim_x:lng, dim_y:lat})
```
aaron-kaplan, line 137: The function looks good, I would just change the name. I thought "sel_center" was going to be a function that finds the center of the spatial array, i.e. the operation you're doing in layout.py line 42. Call it sel_snap? 

This is one of those things that should be built in to xarray, but isn't because it doesn't mange cell bounds. It would be interesting to see if xCDAT implements this operation.

remicousin:
Should I put sel_snap in pingrid since it's going to be used by onset maproom as well?

remicousin:
Then the Marker is now outside the CB.

aaron-kaplan:
> Should I put sel_snap in pingrid since it's going to be used by onset maproom as well?

Yes.

aaron-kaplan:
...with a docstring that explains what it does, in particular how it's different from `DataArray.sel(method=nearest)` (it raises an exception if the point is outside the domain of the dataset).

pingrid.py
```diff
@@ -27,6 +27,17 @@
 RAIN_PNE_COLORMAP = "[0xa020f0 [0x0000ff 35] [0x40e0d0 36] [0x32cd32 36] 0xffe465 [0xffe465 35] 0xffff00 [0xffa500 35] [0xa52a2a 36] [0x000000 36]]"
 CORRELATION_COLORMAP = "[0x000000 0x000080 [0x0000ff 25] [0x00bfff 26] [0x7fffd4 39] [0x98fb98 26] 0xffe465 [0xffe465 25] 0xffff00 [0xff8c00 38] [0xff0000 39] [0x800000 39] 0xa52a2a]"
 
+
+def sel_snap(spatial_array, lat, lng, dim_y="Y", dim_x="X"):
+    """"Selects the spatial_array's closest spatial grid center to the lng/lat coordinate.
+    Raises an excpetion if lng/lat is outside spatial_array domain.
```
aaron-kaplan, line 33: exception is misspelled

pingrid.py
```diff
@@ -27,6 +27,17 @@
...
+    Expects spatial grids to be square.
```
aaron-kaplan, line 34: True. I hadn't thought about that. We don't need to make that assumption: we could first check if the point is inside the bounds of the dataset, and if so, use `.sel(method=nearest)` with no tolerance limit.

That version would also be more correct than this one even in the case of square grid cells. The area defined by the cell center and the tolerance is a circle, not a square; for a pixel on the edge of the dataset, there are points that are inside the tolerance circle but outside the dataset.

aaron-kaplan: Oh, and once we've generalized it to non-square grid cells, we don't have to limit it to spatial dimensions anymore. We could just take arbitrary values for arbitrary coordinates. (I'm guessing this generalization might be hard for you to code. You can leave it for another time.)

aaron-kaplan:
You're right, it's not clear what tolerance is supposed to mean when selecting over multiple dimensions. If it's applied for each dimension independently then it's only meaningful when the grid cells are square.

There are tests for pingrid in fbfmaproom. You can add your tests there.

remicousin:
> You're right, it's not clear what tolerance is supposed to mean when selecting over multiple dimensions. If it's applied for each dimension independently then it's only meaningful when the grid cells are square.
> 
> There are tests for pingrid in fbfmaproom. You can add your tests there.

ah yeah... I didn't think of that: they don't offer tolerance to be some sort of array with the tolerance value per dimension involved...

pingrid.py
remicousin: That all makes sense. Testing whether the point is inside the bounds of the dataset may (or not) be tricky for boundless periodic domains, e.g. global longitude. xarray is not dealing with that, is it? It may have longitude going from -180 to 180 and other from 0 to 360, correct? Then -1 will be in bounds of the first case and out bounds of the 2nd case. It's probably fine to work within such a framework until we come up with periodicity considerations. Or maybe xarray takes care of that?

remicousin: How does that look now?

Obstacles to more generalization:
1/ dims must be evenly spaced for (half-)resolution to exist
2/ dims expected to be in increasing order -- that should be generalizable without much trouble
3/ generalize to apply to n dimensions? how would that work? the input is then a dict rather than a list of values and dims?

pingrid.py
```diff
@@ -27,6 +27,25 @@
...
+    """
+    the_method = None
+    half_res_y = (spatial_array[dim_y][1] - spatial_array[dim_y][0]) / 2
+    half_res_x = (spatial_array[dim_x][1] - spatial_array[dim_x][0]) / 2
+    min_y = spatial_array[dim_y][0] - half_res_y
+    max_y = spatial_array[dim_y][-1] + half_res_y
+    min_x = spatial_array[dim_x][0] - half_res_x
+    max_x = spatial_array[dim_x][-1] + half_res_x
+    if lat >= min_y and lat <= max_y and lng <= min_x and lng >= max_x:
+        the_method = "nearest"
```
aaron-kaplan, line 45: This is the kind of place where you could use the `shapely` approach for bounds checking that I suggested elsewhere, to handle longitude wrapping. But I suppose there's not much point doing that in this case, since the following line is a `.sel`, whcih doesn't handle longitude wrapping.

remicousin:
@aaron-kaplan , please have another look after rebasing and testing.

I put snap_sel and error_fig as public.

sel_snap was wrong (I hadn't tested it actually... my bad...):
besides a mistake in the tests, it turns out that CPT's Y goes from high to low. So not assuming that anymore: making resolution positive and using max/min to find the edges. If max/min too costly for big data, could read first and last and test who is greater.

aaron-kaplan:
I doubt we have any datasets where calculating the min/max of a coordinate would be a noticeable expense. That said, comparing the first and the last is cheaper and easy to program, so why not do it?

remicousin:
Maybe like this is better. High res data at subcontinental level (e.g. NDVI) can be really large...

subseas_flex_fcst/layout.py
```diff
@@ -28,7 +28,7 @@ def app_layout():
 
     # Initialization
     fcst_mu = predictions.sel_cpt_file(DATA_PATH,CONFIG["forecast_mu_file_pattern"],list(CONFIG["leads"])[0],startDates[-1])
-    center_of_the_map = [((fcst_mu.Y[0]+fcst_mu.Y[-1])/2).values, ((fcst_mu.X[0]+fcst_mu.X[-1])/2).values]
+    center_of_the_map = [((fcst_mu.Y[int(fcst_mu.Y.size/2)].values)), ((fcst_mu.X[int(fcst_mu.X.size/2)].values))]
```
aaron-kaplan, line 31: I've been trying to build the habit of using the notation `da['X']` rather than `data_array.X`. They're usually equivalent, but if the variable name happens also to be the name of a `DataArray` method, `da['sum']` gets to the variable whereas `da.sum` gets you the method.

aaron-kaplan:
Looks good to me.

Issue #174, opened 2022-09-19
Move read_cptdataset into separate predictions.py function file


drewmresnick:
Possibly... or make it an issue. The key here is that is we are going to use the same app for other cpt forecasts (biweekly, seasonal...) so all the functions associated with reading will either have to be more fancy to accommodate different cases, or have counterparts for the different cases, and the config file will tell which case we're in.
I am just not sure what shape that will all take as I review this now. So it may make sense to leave to other PR.

_Originally posted by @remicousin in https://github.com/iridl/python-maprooms/pull/156#discussion_r974675445_

remicousin:
This and issue #173  would probably benefit from trying and apply to seasonal CPT outputs. We need some sample files that hopefully we can find from AICCRA colleagues to play with.

aaron-kaplan:
We have lots of seasonal CPT forecasts in the FBF project.

remicousin:
Ah yeah... That's good news! Are they going to be read by ptio.open_cptdataset (I had the impression that this was for somewhat files generated by a somewhat recent pyCPT). We can start with some of those while I request from AICCRA team.

drewmresnick:
@remicousin is this something I should be working on now, or are we waiting until all the changes are merged for the subs maproom to apply it to seasonal data?

aaron-kaplan:
> Are they going to be read by ptio.open_cptdataset (I had the impression that this was for somewhat files generated by a somewhat recent pyCPT). 

I'm not aware that the CPT output format has changed recently.

You can find pointers to the datasets in [fbf-update-data.py](https://github.com/iridl/python-maprooms/blob/master/fbfmaproom/fbf-update-data.py). The ones with "pnep" (precipitation non-exceedance probabiilty) in the name are NextGen seasonal forecasts. The URLs in this file are Ingrid URLs; from there you can read the catalog entries to find the underlying data files.

remicousin:
@drewmresnick , is that not resolved?

PR #187, opened 2022-09-29
Factoring out functions

drewmresnick:
A couple of functions made to be reused.

### DIFF
--- a/pingrid.py
+++ b/pingrid.py
@@ -21,6 +21,7 @@ from shapely.geometry.polygon import LinearRing
 import flask
 import yaml
 import plotly.graph_objects as pgo
+from datetime import datetime, timedelta
 
 
 RAINBOW_COLORMAP = "[0x0000ff [0x00ffff 51] [0x00ff00 51] [0xffff00 51] [0xff0000 51] [0xff00ff 51]]"
@@ -793,3 +794,23 @@ def load_config(colon_separated_filenames):
         with open(fname) as f:
             config = deep_merge(config, yaml.safe_load(f))
     return config
+
+
+def targetRangeFormat(leadsValues,leadsKeys,startDate,periodLength):
+    ''' Formatting target range using leads and starts, and target range period length.
+    '''
+    optionsDict = {}
+    for idx, x in enumerate(leadsValues):
+        targetStart = startDate + timedelta(days=x)
+        targetEnd = targetStart + timedelta(days= periodLength - 1)
+        if (targetStart).strftime("%Y") == (targetEnd).strftime("%Y"):
+            if (targetStart).strftime("%b") == (targetEnd).strftime("%b"):
+                targetStartStr = targetStart.strftime("%-d")
+            else:
+                targetStartStr = (targetStart).strftime("%-d %b")
+        else:
+            targetStartStr = (targetStart).strftime("%-d %b %Y")
+        targetEndStr = targetEnd.strftime("%-d %b %Y")
+        dateRange = f"{targetStartStr} - {targetEndStr}"
+        optionsDict.update({leadsKeys[idx]:dateRange})
+    return optionsDict
--- a/pingrid.py
+++ b/pingrid.py
@@ -794,23 +794,3 @@ def load_config(colon_separated_filenames):
         with open(fname) as f:
             config = deep_merge(config, yaml.safe_load(f))
     return config
-
-
-def target_range_format(leads_values,leads_keys,start_date,period_length):
-    ''' Formatting target range using leads and starts, and target range period length.
-    '''
-    options_dict = {}
-    for idx, x in enumerate(leads_values):
-        target_start = start_date + timedelta(days=x)
-        target_end = target_start + timedelta(days= period_length - 1)
-        if (target_start).strftime("%Y") == (target_end).strftime("%Y"):
-            if (target_start).strftime("%b") == (target_end).strftime("%b"):
-                target_start_str = target_start.strftime("%-d")
-            else:
-                target_start_str = (target_start).strftime("%-d %b")
-        else:
-            target_start_str = (target_start).strftime("%-d %b %Y")
-        target_end_str = target_end.strftime("%-d %b %Y")
-        date_range = f"{target_start_str} - {target_end_str}"
-        options_dict.update({leads_keys[idx]:date_range})
-    return options_dict
--- a/pingrid.py
+++ b/pingrid.py
@@ -796,21 +796,21 @@ def load_config(colon_separated_filenames):
     return config
 
 
-def targetRangeFormat(leadsValues,leadsKeys,startDate,periodLength):
+def target_range_format(leads_values,leads_keys,start_date,period_length):
     ''' Formatting target range using leads and starts, and target range period length.
     '''
-    optionsDict = {}
-    for idx, x in enumerate(leadsValues):
-        targetStart = startDate + timedelta(days=x)
-        targetEnd = targetStart + timedelta(days= periodLength - 1)
-        if (targetStart).strftime("%Y") == (targetEnd).strftime("%Y"):
-            if (targetStart).strftime("%b") == (targetEnd).strftime("%b"):
-                targetStartStr = targetStart.strftime("%-d")
+    options_dict = {}
+    for idx, x in enumerate(leads_values):
+        target_start = start_date + timedelta(days=x)
+        target_end = target_start + timedelta(days= period_length - 1)
+        if (target_start).strftime("%Y") == (target_end).strftime("%Y"):
+            if (target_start).strftime("%b") == (target_end).strftime("%b"):
+                target_start_str = target_start.strftime("%-d")
             else:
-                targetStartStr = (targetStart).strftime("%-d %b")
+                target_start_str = (target_start).strftime("%-d %b")
         else:
-            targetStartStr = (targetStart).strftime("%-d %b %Y")
-        targetEndStr = targetEnd.strftime("%-d %b %Y")
-        dateRange = f"{targetStartStr} - {targetEndStr}"
-        optionsDict.update({leadsKeys[idx]:dateRange})
-    return optionsDict
+            target_start_str = (target_start).strftime("%-d %b %Y")
+        target_end_str = target_end.strftime("%-d %b %Y")
+        date_range = f"{target_start_str} - {target_end_str}"
+        options_dict.update({leads_keys[idx]:date_range})
+    return options_dict
--- a/subseas_flex_fcst/maproom.py
+++ b/subseas_flex_fcst/maproom.py
@@ -104,7 +104,7 @@ def targetStartOptions(start_date):
     for idx, lead in enumerate(leads_keys):
         target_range = predictions.target_range_format(leads_values[idx],leads_keys[idx],start_date,CONFIG["target_period_length"])
         leads_dict.update({lead:target_range})
-    return leads_dict, leads_keys[0]
+    return leads_dict, list(CONFIG["leads"])[0]
 
 
 @APP.callback(
--- a/subseas_flex_fcst/maproom.py
+++ b/subseas_flex_fcst/maproom.py
@@ -100,8 +100,11 @@ def targetStartOptions(start_date):
     leads_values = list(CONFIG["leads"].values())
     leads_keys = list(CONFIG["leads"])
     start_date = pd.to_datetime(start_date)
-    options_dict = predictions.target_range_format(leads_values,leads_keys,start_date,CONFIG["target_period_length"])
-    return options_dict, leads_keys[0]
+    leads_dict = {}
+    for idx, lead in enumerate(leads_keys):
+        target_range = predictions.target_range_format(leads_values[idx],leads_keys[idx],start_date,CONFIG["target_period_length"])
+        leads_dict.update({lead:target_range})
+    return leads_dict, leads_keys[0]
 
 
 @APP.callback(
@@ -164,9 +167,7 @@ def local_plots(n_clicks, click_lat_lng, startDate, leadTime, latitude, longitud
         errorFig = pingrid.error_fig(error_msg="Grid box out of data domain")
         return errorFig, errorFig, None, dlf.Marker(position=[lat, lng]), lat, lng
     #get issue date and target range for making plots titles
-    issue_date = (pd.to_datetime(startDate)).strftime("%-d %b %Y")
-    targets_dict = predictions.target_range_format(list(CONFIG["leads"].values()),list(CONFIG["leads"]),pd.to_datetime(startDate),CONFIG["target_period_length"])
-    target_range = targets_dict[leadTime]
+    target_range = predictions.target_range_format(CONFIG["leads"][leadTime],leadTime,pd.to_datetime(startDate),CONFIG["target_period_length"])
     # CDF from 499 quantiles
     quantiles = np.arange(1, 500) / 500
     quantiles = xr.DataArray(
@@ -245,7 +246,7 @@ def local_plots(n_clicks, click_lat_lng, startDate, leadTime, latitude, longitud
         xaxis_title=f'{CONFIG["variable"]} ({fcst_mu.attrs["units"]})',
         yaxis_title="Probability of exceeding",
         title={
-            "text": f"{target_range} forecast issued {issue_date} <br> at ({fcst_mu.Y.values}N,{fcst_mu.X.values}E)",
+            "text": f'{target_range} forecast issued {(pd.to_datetime(startDate)).strftime("%-d %b %Y")} <br> at ({fcst_mu.Y.values}N,{fcst_mu.X.values}E)',
             "font": dict(size=14),
         },
     )
@@ -298,7 +299,7 @@ def local_plots(n_clicks, click_lat_lng, startDate, leadTime, latitude, longitud
         xaxis_title=f'{CONFIG["variable"]} ({fcst_mu.attrs["units"]})',
         yaxis_title="Probability density",
         title={
-            "text": f"{target_range} forecast issued {issue_date} <br> at ({fcst_mu.Y.values}N,{fcst_mu.X.values}E)",
+            "text": f'{target_range} forecast issued {(pd.to_datetime(startDate)).strftime("%-d %b %Y")} <br> at ({fcst_mu.Y.values}N,{fcst_mu.X.values}E)',
             "font": dict(size=14),
         },
     )
--- a/subseas_flex_fcst/maproom.py
+++ b/subseas_flex_fcst/maproom.py
@@ -100,22 +100,9 @@ def targetStartOptions(startDate):
     leadsValues = list(CONFIG["leads"].values())
     leadsKeys = list(CONFIG["leads"])
     startDate = pd.to_datetime(startDate)
-    optionsDict = {}
-    for idx, x in enumerate(leadsValues):
-        endDate = x+CONFIG["target_period_length"]-1 #used to calculate the target end date (targetEnd = lead + period length - 1)
-        targetStart = startDate + timedelta(days=x)
-        targetEnd = startDate + timedelta(days=endDate)
-        if (targetStart).strftime("%Y") == (targetEnd).strftime("%Y"):
-            if (targetStart).strftime("%b") == (targetEnd).strftime("%b"):
-                targetStartStr = targetStart.strftime("%-d")
-            else:
-                targetStartStr = (targetStart).strftime("%-d %b")
-        else:
-            targetStartStr = (targetStart).strftime("%-d %b %Y")
-        targetEndStr = targetEnd.strftime("%-d %b %Y")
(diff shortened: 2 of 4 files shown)


drewmresnick:
@aaron-kaplan Should I be keeping files in this predictions.py file I created per Rémi's suggestion, or should they be moved into pingrid.py? Or, is pingrid.py for specific functions?

pingrid.py
```diff
@@ -770,3 +771,23 @@ def _proxy(fn, *args, **kwargs):
     if decode_times:
         ds = fix_calendar(ds)
     return ds
+
+
+def targetRangeFormat(leadsValues,leadsKeys,startDate,periodLength):
```
drewmresnick, line 776: I did not know if it was better to have the inputs be the separate lists of lead values / keys like this, or if it is better to have the input be the dictionary and then within the function, have the code that creates the lists. Or if it makes much of a difference either way.

drewmresnick:
@drewmresnick thoughts on addressing issue #174 in this PR?

aaron-kaplan:
> @aaron-kaplan Should I be keeping files in this predictions.py file I created per Rémi's suggestion, or should they be moved into pingrid.py?

Ultimately these functions will go in a library that's shared between projects, but predictions.py is good for now. We have a lot of reorganizing to do with pingrid.py.

BTW wouldn't it be more idiomatic to call it forecasts.py rather than predictions.py?

pingrid.py
aaron-kaplan: Personally I would choose neither of those options. I would have a function that formats a single date range, and call that function once for each value in the dictionary.

subseas_flex_fcst/predictions.py
```diff
@@ -69,3 +69,19 @@ def combine_cptdataset(dataDir,filePattern,dof=False):
...
     return dataCombine
+
+
+def cpt_startsList(dataPath,filePattern,dateSearchPattern,zeroPadding=False):
+    filesNameList = glob.glob(f'{dataPath}/{filePattern}')
+    startDates = []
+    for idx, i in enumerate(filesNameList):
+        startDate = re.search(dateSearchPattern,filesNameList[idx])
+        startDatedt = datetime.strptime(startDate.group(),"%b-%d-%Y")
+        startDateStr = startDatedt.strftime("%b-%-d-%Y")
```
aaron-kaplan, line 80: StartDateStr is unused?

aaron-kaplan:
> Ultimately these functions will go in a library that's shared between projects, but predictions.py is good for now.

Now I see that you put the date range formatting function in pingrid.py. That's fine too.

aaron-kaplan:
@drewmresnick variable names in python traditionally use `snake_case`, not `camelCase`. On a one-person project it doesn't really make a difference, but in a project with multiple people contributing it's good to have conventions so things don't get too inconsistent.

In JavaScript it's traditional to use camelCase, and dash is a python wrapper around some JavaScript  libraries, so in a dash project you may see camelCase bleeding into python code in some places. But in our own code, we should stick to snake_case.

I've not been too strict about this so far because I didn't want to get too nitpicky, but it's becoming apparent that if I don't start enforcing it now then it's going to get harder and harder to do later.

subseas_flex_fcst/predictions.py
```diff
@@ -69,3 +69,19 @@ def combine_cptdataset(dataDir,filePattern,dof=False):
...
+        startDates.append(startDatedt)
+    startDates = sorted(set(startDates))
```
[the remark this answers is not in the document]
drewmresnick: The way it stands currently, the code uses the file pattern in the config file. The file pattern thus does not include the part of the title that identifies the lead, so in this function there will be four copies of each date as the function finds four files for each start date. I can find a way to get around doing this, if you think it is preferred.

subseas_flex_fcst/predictions.py
```diff
@@ -69,3 +69,19 @@ def combine_cptdataset(dataDir,filePattern,dof=False):
         dataDic[idx] = ds
     dataCombine = xr.combine_by_coords([dataDic[x] for x in dataDic],combine_attrs="drop_conflicts")
     return dataCombine
+
+
+def cpt_startsList(dataPath,filePattern,dateSearchPattern,zeroPadding=False):
+    filesNameList = glob.glob(f'{dataPath}/{filePattern}')
+    startDates = []
+    for idx, i in enumerate(filesNameList):
+        startDate = re.search(dateSearchPattern,filesNameList[idx])
```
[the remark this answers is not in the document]
drewmresnick: Oh I did not realize this!

drewmresnick:
> @drewmresnick variable names in python traditionally use `snake_case`, not `camelCase`. On a one-person project it doesn't really make a difference, but in a project with multiple people contributing it's good to have conventions so things don't get too inconsistent.
> 
> In JavaScript it's traditional to use camelCase, and dash is a python wrapper around some JavaScript libraries, so in a dash project you may see camelCase bleeding into python code in some places. But in our own code, we should stick to snake_case.
> 
> I've not been too strict about this so far because I didn't want to get too nitpicky, but it's becoming apparent that if I don't start enforcing it now then it's going to get harder and harder to do later.

Thank you for pointing this out, I am not sure at what point I started adopting this practice... I will make sure to use snake_case moving forward. Should I go back and change this in the files I have been working on?

subseas_flex_fcst/predictions.py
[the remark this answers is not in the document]
aaron-kaplan: I see. I suggest that you either add a comment explaining what you just explained here, or pick one lead arbitrarily and just list the files for that lead

aaron-kaplan:
> Should I go back and change this in the files I have been working on?

I guess it would be good to get started on that. Doesn't all have to happen at once.

remicousin:
> > @aaron-kaplan Should I be keeping files in this predictions.py file I created per Rémi's suggestion, or should they be moved into pingrid.py?
> 
> Ultimately these functions will go in a library that's shared between projects, but predictions.py is good for now. We have a lot of reorganizing to do with pingrid.py.
> 
> BTW wouldn't it be more idiomatic to call it forecasts.py rather than predictions.py?

I probably suggested predictions over forecasts because I have a sense predictions is more generic and can include, e.g. projections, but don't quote on me on that.

remicousin:
> > @drewmresnick variable names in python traditionally use `snake_case`, not `camelCase`. On a one-person project it doesn't really make a difference, but in a project with multiple people contributing it's good to have conventions so things don't get too inconsistent.
> > In JavaScript it's traditional to use camelCase, and dash is a python wrapper around some JavaScript libraries, so in a dash project you may see camelCase bleeding into python code in some places. But in our own code, we should stick to snake_case.
> > I've not been too strict about this so far because I didn't want to get too nitpicky, but it's becoming apparent that if I don't start enforcing it now then it's going to get harder and harder to do later.
> 
> Thank you for pointing this out, I am not sure at what point I started adopting this practice... I will make sure to use snake_case moving forward. Should I go back and change this in the files I have been working on?

I am also guilty on that one: I know there are conventions, but I don't know them well enough to implement them automatically. One strategy is to cut on naming on objects :) , which often can be done combining code. That might make things harder to read, or easier, depending on one's literacy in Python and on how good the author is on picking names.

So anyway yes, I think it's fine to point us to those mistakes so that we progressively get the right habits.

remicousin:
Also, I think we should start and get the habit to write function documentation properly in new .py files. I guess it's fine not to do so in the functions in maproom/layout.py that are meant to be very specific to the given app; but when we factor things out in another file, it's because we expect wider use, thus we should write proper documentation. Probably true also for pingrid.py but Aaron can comment on that.

subseas_flex_fcst/predictions.py
```diff
@@ -69,3 +69,19 @@ def combine_cptdataset(dataDir,filePattern,dof=False):
...
+        startDatedt = datetime.strptime(startDate.group(),"%b-%d-%Y")
+        startDateStr = startDatedt.strftime("%b-%-d-%Y")
+        startDates.append(startDatedt)
+    startDates = sorted(set(startDates))
+    if zeroPadding == False:
+        startDates = [i.strftime("%b-%-d-%Y") for i in startDates] #needs to have date with no zero padding to match the file path namesme
+    if zeroPadding == True:
+        startDates = [i.strftime("%b-%d-%Y") for i in startDates]
```
remicousin, line 86: I don't think zeroPadding is going to help many cases. Some files will have 0s, some don't, that's covered. Some files will have the month expressed in numbers, with or without leading 0s... and it can go on like this. I would put the input of strftime as an input of the function instead.

remicousin:
selFile is basically an extension of open_cptdataset. Hopefully we can make it work at some point for anything open_cptdataset can use, but likely not for other set of forecast files that comes from something else. So I would call selFile more explicitly as something like open_cptfile

pingrid.py
remicousin: Additionally, there is also a getTargets function that is very similar in predictions.py. I think there is some rationalization to do with those 2 functions.   Maybe something like one function that gets target_start and target_end from issue, lead and length. Another that formats dates, whether they are a single date or a period described by a start and end.

subseas_flex_fcst/predictions.py
remicousin: In light of previous comments above. There might be a need to have a set of functions that manipulate time. We need:

1. functions that are given the strftime format of file names and convert that in datetime objects;
2. functions that manipulate the different datetime objects, for instance to retrieve target_start and target_end from issue and lead
3. functions that print our datetime objects in strftime formats we want in the app at various locations.

I am not pretending I have the solution but hopefully it helps rationalize the whole set.

drewmresnick:
> > > @aaron-kaplan Should I be keeping files in this predictions.py file I created per Rémi's suggestion, or should they be moved into pingrid.py?
> > 
> > 
> > Ultimately these functions will go in a library that's shared between projects, but predictions.py is good for now. We have a lot of reorganizing to do with pingrid.py.
> > BTW wouldn't it be more idiomatic to call it forecasts.py rather than predictions.py?
> 
> I probably suggested predictions over forecasts because I have a sense predictions is more generic and can include, e.g. projections, but don't quote on me on that.

I will leave the name for now; maybe once we have a better sense of which functions will stay in this file and which might be moved elsewhere we may come up with a different name, or decide this is the best name for the functions we are keeping here.

drewmresnick:
> Also, I think we should start and get the habit to write function documentation properly in new .py files. I guess it's fine not to do so in the functions in maproom/layout.py that are meant to be very specific to the given app; but when we factor things out in another file, it's because we expect wider use, thus we should write proper documentation. Probably true also for pingrid.py but Aaron can comment on that.

Yes I agree with this one, I was going to wait until the functions were finished before starting on the documentation, to not waste time in case things change.

subseas_flex_fcst/predictions.py
drewmresnick: Yes I think adding the zeroPadding option was my first attempt at trying to allow some flexibility in the way the user can handle the way the dates are output. Then I hoped by having the user be able to input the search pattern for the date that would handle at least flexibility in the way the dates are expressed initially in the file names when the search occurs (though this does assume the user knows enough about regex to figure out the pattern- though we can give a few common examples in the docs if this function gets approved.)

Maybe it is worth making an issue to address a set of functions that manipulate time. And speak a bit more about what functions might make the most sense for this.

I guess this all might not answer the question of if I should remove the zero_padding option from the file though.

drewmresnick:
> selFile is basically an extension of open_cptdataset. Hopefully we can make it work at some point for anything open_cptdataset can use, but likely not for other set of forecast files that comes from something else. So I would call selFile more explicitly as something like open_cptfile

Yes, both the selFile and combine_cptdataset functions are outdated functions I had written, you suggested I leave them in this file in case we want to reuse and update the code in the case we ever want to combine all of the datasets like I had done previously in the maproom. Perhaps it is a bit confusing to have these in the same file as functions we actually use are. So it would be best either to

1. move the functions we are using to pingrid.py and leave the unused functions in predictions.py, or
2. keep predictions.py to have these maproom specific functions we use and move the unused functions to yet another file, just to keep the functions around? 

Unless we just want to trash those functions altogether.

aaron-kaplan:
> Also, I think we should start and get the habit to write function documentation properly in new .py files. I guess it's fine not to do so in the functions in maproom/layout.py that are meant to be very specific to the given app; but when we factor things out in another file, it's because we expect wider use, thus we should write proper documentation. Probably true also for pingrid.py but Aaron can comment on that.

Completely agree, and while we're on the subject, let's also remember to write tests. The date range formatting function, for example, is a great candidate for some unit tests. They should be quite easy to write, will help document the intended behavior, and might even catch a bug or two.

aaron-kaplan:
> Yes, both the selFile and combine_cptdataset functions are outdated functions I had written, you suggested I leave them in this file

I think that was me, actually. I didn't mean that you should necessarily keep them in this project, just put them somewhere where you'll be able to find them later if you need them.

It occurs to me now that the subx-zarr project has a relatively sophisticated way of configuring and using forecast datasets. At some point we should generalize that and use it in places like this.

pingrid.py
drewmresnick: Latest commits address these comments. I thought about making the `target_range_format` function return the target start and target end separately instead of as a single string, what do you think? Or leave it as returning as a single string?

drewmresnick:
> > Yes, both the selFile and combine_cptdataset functions are outdated functions I had written, you suggested I leave them in this file
> 
> I think that was me, actually. I didn't mean that you should necessarily keep them in this project, just put them somewhere where you'll be able to find them later if you need them.

Ah okay, I have moved them out of the predictions.py file

> It occurs to me now that the subx-zarr project has a relatively sophisticated way of configuring and using forecast datasets. At some point we should generalize that and use it in places like this.

Is this something we should make an issue for to remember to come back to?

subseas_flex_fcst/maproom.py
```diff
@@ -99,26 +99,16 @@ def display_relevant_control(variable):
...
-    return optionsDict, leadsKeys[0]
+def targetStartOptions(start_date):
+    leads_values = list(CONFIG["leads"].values())
+    leads_keys = list(CONFIG["leads"])
+    start_date = pd.to_datetime(start_date)
+    leads_dict = {}
+    for idx, lead in enumerate(leads_keys):
+        target_range = predictions.target_range_format(leads_values[idx],leads_keys[idx],start_date,CONFIG["target_period_length"])
+        leads_dict.update({lead:target_range})
+    return leads_dict, leads_keys[0]
```
remicousin, line 110: you output leads_keys[0] here by in layout's leadTime, you point value directly to the CONFIG entry. You'd have to pick one or the other. I think pointing directly to the CONFIG file is good enough. If we need the callback to take of that at some point, we can revisit.

pingrid.py
remicousin: I think we almost never will need target_start and target_end separately so I am not sure there is added value outputing them separately. I think we will need at some point to split what calculates the datetime objects and what format them for printing. But we can deal with that later when concrete cases arise.

subseas_flex_fcst/predictions.py
remicousin: ok with setting up an issue

drewmresnick: I made a new issue to address making new functions, and changed the zero_padding option for the existing function to take the strftime format as an input.

drewmresnick:
@aaron-kaplan @remicousin should I create a new tests subdirectory and make a few tests for these functions in this same PR or should we make a new PR specifically for testing?

remicousin:
> @aaron-kaplan @remicousin should I create a new tests subdirectory and make a few tests for these functions in this same PR or should we make a new PR specifically for testing?

It think in principle I would build tests as you go, and that's also a mean for the reviewer to review and understand better the code. But since there are many things in this PR, I would not mind having this in another PR.

pingrid.py
```diff
@@ -793,3 +794,23 @@ def load_config(colon_separated_filenames):
         with open(fname) as f:
             config = deep_merge(config, yaml.safe_load(f))
     return config
+
+
+def target_range_format(leads_values,leads_keys,start_date,period_length):
```
remicousin, line 799: There is a version of this in predictions.py, that deals with one entry at a time. And the app seems to be using the predictions.py one. So maybe this one can go?

subseas_flex_fcst/predictions.py
```diff
@@ -8,64 +8,106 @@
...
-def selFile(dataPath, filePattern, leadTime, startDate):
+def sel_cpt_file(data_path, file_pattern, leadTime, startDate):
+    """ Select a single cpt file for a given start and lead.
+
+    Parameters
+    ----------
+    data_path : str
+        String of the path pointing to cpt datasets.
+    file_pattern : str
+        String of the file pattern name for a given variable's data file.
```
remicousin, line 19: filename pattern

subseas_flex_fcst/predictions.py
```diff
@@ -8,64 +8,106 @@
...
+        String of the start date to be selected for as is represented in the file name.
+    Returns
+    -------
+    fileSelected : xarray Dataset
+        Single CPT data file as multidimensional xarray dataset.
+    Notes
+    -----
+    `file_pattern` should be most common denominator for any group of datasets,
+    so that a sinlge file can be selected using only `leadTime` and `startDate`.
+    """
```
remicousin, line 32: Maybe add an example? I think that's an official documentation section.

subseas_flex_fcst/predictions.py
```diff
@@ -8,64 +8,106 @@
...
-        ds = ds.expand_dims(leadTimeDict)
-        dataDic[idx] = ds
-    dataCombine = xr.combine_by_coords([dataDic[x] for x in dataDic],combine_attrs="drop_conflicts")
-    return dataCombine
+    Parameters
+    ----------
+    data_path : str
+        String of the path pointing to cpt datasets.
+    file_pattern : str
+        String of the file pattern name for a given variable's data file.
```
remicousin, line 86: filename pattern

subseas_flex_fcst/maproom.py
```diff
@@ -96,26 +96,16 @@ def display_relevant_control(variable):
...
-    return optionsDict, leadsKeys[0]
+def targetStartOptions(start_date):
+    leads_values = list(CONFIG["leads"].values())
+    leads_keys = list(CONFIG["leads"])
+    start_date = pd.to_datetime(start_date)
+    leads_dict = {}
+    for idx, lead in enumerate(leads_keys):
+        target_range = predictions.target_range_format(leads_values[idx],leads_keys[idx],start_date,CONFIG["target_period_length"])
+        leads_dict.update({lead:target_range})
+    return leads_dict, list(CONFIG["leads"])[0]
```
remicousin, line 107: now, your callback outputs the exact a value that you explicitely give in your layout, instead of using the result of your CB. Do one or the other. That is to say either:
1/ Don't output leadTime/value
OR
2/ remove the line value=list(CONFIG["leads"])[0]
in the layout's leadTime
I guess I would do 1/ since the default leadTime value is always the same.

remicousin (changes requested):

subseas_flex_fcst/maproom.py
drewmresnick: So this part is a little buggy; if you do not have the leadTime/value as an output of the CB, then the value does not reset to be the first lead time when the use selects a different start date (for example if the user picks the thirs lead time and then selects a new start date, the leadtime value will not reset back to the first lead time). So as far as I can tell, you need that as an output in the callback. 

So I can do the second option of removing the initial value definition in the leadTime layout. And then leave the callback as-is.

drewmresnick:
> > @aaron-kaplan @remicousin should I create a new tests subdirectory and make a few tests for these functions in this same PR or should we make a new PR specifically for testing?
> 
> It think in principle I would build tests as you go, and that's also a mean for the reviewer to review and understand better the code. But since there are many things in this PR, I would not mind having this in another PR.

Okay I will work on a new branch for testing once this PR is merged.

subseas_flex_fcst/maproom.py
remicousin: My point was mostly that you are not using the output of the CB, unless it's overwritten? But anyways, it's very fine with me that the lead doesn't reset to the first one when users change the issue date. All that makes sense to me is that it is the first one on opening of the page.

drewmresnick: I think that it is overwritten each time. Personally I think it is better for it to reset each time a user chooses a new start date, for it to go back to the first lead time option. I think I would be a bit confused when first using the app if it it did not do so.

remicousin: I see... I think I came to realize that components are overwritten indeed if they have a value at first.

Resetting to the first lead time looks good. I think it was hard to do with uicore so didn't implement so.

So I guess this clarification resolves that.

remicousin (approved):
I believe this is it?

drewmresnick:
> I believe this is it?

Yes, for this PR. 

The next steps for these functions I will do in another PR, which will be writing documentation and tests.

PR #188, opened 2022-09-30
Fix regression in error handling when click outside forecast area

aaron-kaplan:
When you click outside the project area it's supposed to blank out the table, but currently it doesn't. I probably broke this in my recent PR where I refactored the code that pulls shapes from the db.

### DIFF
--- a/fbfmaproom/fbfmaproom.py
+++ b/fbfmaproom/fbfmaproom.py
@@ -407,7 +407,8 @@ def subquery_unique(base_query, key, field):
     with psycopg2.connect(**CONFIG["db"]) as conn:
         df = pd.read_sql(query, conn, params={"key": key})
     if len(df) == 0:
-        raise InvalidRequestError(f"invalid region {region_key}")
+        raise Exception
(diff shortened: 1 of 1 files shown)


fbfmaproom/fbfmaproom.py
```diff
@@ -406,7 +406,8 @@ def subquery_unique(base_query, key, field):
     with psycopg2.connect(**CONFIG["db"]) as conn:
         df = pd.read_sql(query, conn, params={"key": key})
     if len(df) == 0:
-        raise InvalidRequestError(f"invalid region {region_key}")
+        raise Exception
```
kgraaf, line 409: Why does this work if you have two adjacent `raise`s?

aaron-kaplan: Fixed. It doesn't actually matter what exception is raised here, but I removed the less specific one.

aaron-kaplan: @kgraaf please approve.

PR #189, opened 2022-09-30
Remove pyaconf from current projects

aaron-kaplan:
After the recent removal of queuepool, pyaconf was the only package left that we were still installing with pip. Having pip in the mix has been preventing us from using a more efficient workflow for managing conda environments (but I'm not changing the workflow yet in this PR). We weren't actually using much of pyaconf, so I added the functionality we need to pingrid.py and removed the rest.

I'm not modifying the PRISM or x24 maprooms, as they're not under active development. If at some point we revive them, we'll update them to follow the latest patterns.

Only one active project (onset_maproom) had an environment_macos.yml. I've deleted it rather than updating it. The new workflow will make macos development easier soon; in the meantime just build your environment from environment.yml if you're going to work on macos.


remicousin:
What's a useful review for us to approve here? Check that branch out, rebuild environment and test apps?

aaron-kaplan:
> What's a useful review for us to approve here? Check that branch out, rebuild environment and test apps?

That sounds good. It's mostly an FYI.

remicousin:
forecast maproom worked just fine. Your version relies on outdated config files for onset Maprooms. Given this all about how config file is read, I am not sure there is a point testing an older version. I am going to approve, then if you'd like me to test the onset maprooms, please rebase and let know.

remicousin (approved):

aaron-kaplan:
> Your version relies on outdated config files for onset Maprooms.

I just double-checked and as far as I can tell this branch is built on top of the current master. Are you sure the config file changes you're thinking of aren't in unmerged branches?

If you're about to merge something, I can hold this one back and rebase it after you merge.

remicousin:
hmm... never mind me... I'll test the onset maprooms now.

remicousin:
All works fine

aaron-kaplan:
This is merged, but the old environment should still work, and I have one more change in the queue that will affect how we build environments, so you can put off rebuilding your environments for another day or two if you want.

Please rebase all open branches onto this, otherwise we won't be able to test them with the new/current environment.

PR #191, opened 2022-10-01
Introduce conda-lock for fbfmaproom

aaron-kaplan:
Instead of `conda env export` we now use `conda-lock lock`; instead of `conda env create` we now use `conda create`. See README.md for details on the commands to run.

The advantages of this scheme are:
- Once the lock file is generated, you can build an environment from it without re-running the dependency solver, which means creating an environment now takes 10-20 seconds instead of 5-10 minutes.
- A developer on linux can generate the fully specified environment for macOS. (Not sure exactly how big an advantage this is--after generating a new macOS environment, the linux developer can't test to see if it actually works.)

@kgraaf please review and approve; the others are FYI.


remicousin (approved):
I approve in principle. Please let know if you'd like me to do something specific about it.

remicousin:
PS: does it change the way to add pytest to the environment once built?

aaron-kaplan:
> PS: does it change the way to add pytest to the environment once built?

Managing dev dependencies (pytest, netcdf, black, ...) is next on my list to address.

kgraaf (approved):

PR #193, opened 2022-10-05
simplest water balance step

remicousin:
The very first part of the water balance step-wise function.

One thing I am not sure of is that I think the final result should be a Dataset. Should it be so at this stage? and if so, should then the inputs that belong the to the water balance Dataset be a Dataset too? Or should we deal with building a Dataset later?

Also, when I run the building of the documentation, I only get calc.py module. I am not sure what I am missing to get the agronomy.py module...

Next part would be either to introduce Kc.

### DIFF
--- a/onset_maproom/agronomy.py
+++ b/onset_maproom/agronomy.py
@@ -1,0 +1,45 @@
+import xarray as xr
+
+def soil_plant_water_balance(
+sm_previous_day,
+rain,
+et,
+taw,
+runoff=0
+):
+    """Compute soil-plant-water balance from one day to the next.
+    The balance is defined as:
+    `sm`(t) + `drainage`(t) = `sm`(t-1) + `peffective`(t) - `et`(t)
+    where:
+    `sm` is the soil moisture and can not exceed total available water `taw`.
+    `drainage` is the residual soil moisutre occasionally excessing `taw` that drains through the soil.
+    `peffective` is the effective precipitation that enters the soil and is the `rain` minus a `runoff`.
+    `et` is the evapotranspiration consumed by the plant.
+    
+    Parameters
+    ------
+    sm_previous_day : DataArray
+        soil moisture of the previous day.
+    rain : DataArray
+        rainfall during the day.
+    et : DataArray
+        evapotransipiration of the plant during the day.
+    taw: DataArray
+        total available water that represents the maximum water capacity of the soil
+    runoff : DataArray
+        amount of rainfall lost to runoff during the day (default `runoff`=0).
+    Returns
+    -------
+    sm, peffective, drainage : Tuple of DataArray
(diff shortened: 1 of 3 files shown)


onset_maproom/agronomy.py
```diff
@@ -0,0 +1,45 @@
...
+et,
+taw,
+runoff=0
+):
+    """Compute soil-plant-water balance from one day to the next.
+    The balance is defined as:
+    `sm`(t) + `drainage`(t) = `sm`(t-1) + `peffective`(t) - `et`(t)
+    where:
+    `sm` is the soil moisture and can not exceed total available water `taw`.
+    `drainage` is the residual soil moisutre occasionally excessing `taw` that drains through the soil.
```
aaron-kaplan, line 15: "moisture" misspelled

aaron-kaplan: excessing --> exceeding?

onset_maproom/agronomy.py
```diff
@@ -0,0 +1,45 @@
...
+    et : DataArray
+        evapotransipiration of the plant during the day.
+    taw: DataArray
+        total available water that represents the maximum water capacity of the soil
+    runoff : DataArray
+        amount of rainfall lost to runoff during the day (default `runoff`=0).
+    Returns
+    -------
+    sm, peffective, drainage : Tuple of DataArray
+        next day soil moisture, effective precipitation and drainage
```
aaron-kaplan, line 34: You inconsistently refer to day t as either "the day" or "the next day". Could be misunderstood to mean that there are three days involved (the previous day, the current day, and the next day) instead of two.

aaron-kaplan (approved):
Nice and simple!

onset_maproom/agronomy.py
aaron-kaplan: :+1:

onset_maproom/agronomy.py
```diff
@@ -0,0 +1,45 @@
...
+runoff=0
+):
+    """Compute soil-plant-water balance from one day to the next.
+    The balance is defined as:
+    `sm`(t) + `drainage`(t) = `sm`(t-1) + `peffective`(t) - `et`(t)
+    where:
+    `sm` is the soil moisture and can not exceed total available water `taw`.
+    `drainage` is the residual soil moisutre occasionally excessing `taw` that drains through the soil.
+    `peffective` is the effective precipitation that enters the soil and is the `rain` minus a `runoff`.
+    `et` is the evapotranspiration consumed by the plant.
```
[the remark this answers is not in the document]
aaron-kaplan: :+1:

aaron-kaplan:
I reapprove.

aaron-kaplan:
To build the docs for the new modules, add them to docsrc/index.rst.

PR #194, opened 2022-10-05
Remic/dyntext

remicousin:
Historically in Ingrid Maprooms, the Maproom text has been static, so we've been writing what is viewed by default and then describing what can be changed. Now with the callbacks, we can easily have dynamic text and just describe what is currently viewed, and describe the other options.

This is a first attempt to do so. Doing it, I've been thinking of different things to implement this otherwise:

1. The map choice basically is associated with 3 items: the option description of the drop down menu; the more detailed description of what it is (showing in the text); and the map title. I am thinking I could create some sort of dictionary/list to have the whole info more tightly together
2. I wrote callback that output the text. I am thinking text could always be there and callbacks would change the value of display style attribute. This way, in the list of possible other map choices description, we could not display the description of what is in view, that is already displayed at the top. That would save some space
3. The Other choices description could be tooltips displayed from mousing-over the menu choices. I have mixed feelings about this idea because detailed description could be long and hard to read on a tooltip... or maybe not? That would have the advantage of saving space and no more callback for the other options.
4. All the info I talk about in 1. could be in the config file as I can easily imagine individual countries wanting to change to what they like
5. I think we should have some visual indicating that text is being changed as page reloads.

Thoughts?


drewmresnick:
I think it looks good to have it updated in the main description, I think it might be a little crowded to have that entire description in a mouseover unless it is maybe a summarized version of the description? 

I do agree that if you keep it to update only in the main description that it might be nice to have a visual indicator that it is updating; maybe there is a way to put that part of the text inside a div of some sort whose border gets highlighted when it is updated or something like that.

onset_maproom/layout.py
```diff
@@ -154,18 +154,19 @@ def controls_layout(lat_min, lat_max, lon_min, lon_max, lat_label, lon_label):
                         is critical to agriculture planification, in particular for planting.
                         """
                     ),
+                    dcc.Loading(html.P(id = "map_description"), type="dot"),
```
aaron-kaplan, line 157: PEP-8 says no spaces around the equals sign in a kwarg.
```
html.P(id="map_description")
```

onset_maproom/layout.py
```diff
@@ -284,13 +268,10 @@ def controls_layout(lat_min, lat_max, lon_min, lon_max, lat_label, lon_label):
...
-                            value="monit",
-                            # bs_size="sm",
+                            value=CONFIG["map_text"][0]["menu_value"],
                             options=[
-                                {"label": "Has germinating rain occured?", "value": "monit"},
-                                {"label": "When to prepare for planting?", "value": "mean"},
-                                {"label": "Climatological Standard deviation", "value": "stddev"},
-                                {"label": "How risky to plant...", "value": "pe"},
+                                {"label": text["menu_label"], "value": text["menu_value"]}
+                                for i, text in enumerate(CONFIG["map_text"])
```
aaron-kaplan, line 274: If you rewrite the config file as I suggested elsewhere, then this becomes
```
options=[
    {"label": val["menu_label"], "value": key}
    for key, val in CONFIG["map_text"]
]
```

onset_maproom/maproom.py
```diff
@@ -292,6 +292,16 @@ def write_map_title(search_start_day, search_start_month, map_choice, probExcThr
     return mytitle
 
 
+@APP.callback(
+    Output("map_description", "children"),
+    Input("map_choice", "value"),
+)
+def write_map_description(map_choice):
+    for i, text in enumerate(CONFIG["map_text"]):
```
aaron-kaplan, line 300: You never use `i`, so you don't need `enumerate`.
```
for text in CONFIG["map_text"]:
    if text["menu_value"] == map_choice:
        return text["description"]
```
Furthermore, if you rewrite the config file as
```
map_text:
    monit:
        menu_label: Has germinating rain occured?
        description: The map shows the result of the germinating rain date search
                (i.e. near-real time thus without checking of follow-up dry spells)
                from the most recent Early Start to now
                (or the last day with available rainfall data).
                Dates are expressed in days since an Early Start date.
```
Then you can eliminate the loop entirely and just write
```
 return CONFIG["map_text"][map_choice]["description"]
```

onset_maproom/layout.py
aaron-kaplan: I just edited the above. I had forgotten to add `.items()`.

aaron-kaplan:
Now that I see how much copy-paste is now involved in building a configuration file for a new country, I'm not as keen on this approach. For one thing, the more copy-paste people do, the more likely they are to make mistakes. Secondly, the more things are customizable, the harder it becomes for us to ensure that future changes are consistent with every country's customization.

To be honest, I was never really happy with the idea that Senegal's onset date maproom could have a different title and a different URL path than the other countries' onset date maprooms, either. At some point it makes more sense to make a whole new maproom, based on the original but not directly using it, than to make the original flexible enough to let everyone make all the customizations they want. Of course, it's hard to say exactly where that point lies.

If we're not yet to the point of giving one country an entirely custom version of the maproom, then I suggest we put all the default text values in one config file (per language, eventually), and use the multiple config file mechanism to let individual countries override those defaults if they want to. As I explained on another PR earlier today, the CONFIG environment variable can be a colon-separated list of files; values taken from files later in the list override values from earlier files.

remicousin:
The double-layered config files makes sense to me. How far customization should go is a tough question. The ultimate goal is to transfer technology, which would suggest that customization could go very far. The non-customizable parts facilitate replication, that is to say our work, less so the partners' work, who arguably don't care about what's going on in other countries. This being said, very very few partners have actually made any Ingrid/uicore Maprooms. LMS is not making new Python Maprooms as far as I can tell. I am hopeful that progressively they will, the democratic Python technology helps. But they still have resource scarcity issues to devote to new things (that would be less true with ICPAC and AGRHYMET). However, if we try and see further into the future and assume we reach a point where we build a community of developers/contributors who exchange on their production, and whether to make customizable, their own thing, or part of the new non-custom parts, would be an immense win, be cause then we could activate the very demanded keyword: South-South collaboration (as opposed to North-South traditional aid).

So I don't know where the fine line is. Maybe when we have more core things moved to pingrid or widgets, it will be easier to see things through. As for the specific case of Senegal, the URL may not have to be different. The title may have to. I would expect that the other countries eventually get a cessation date component. It's not as much a matter of design that of data/method availability to compute a cessation date.

remicousin:
I believe I fixed the syntax and use of lists/dicts correctly. I hope. And then we have now one single config-sample. I tested all countries with the overwriting method and that worked out. As a result, the config for the zarr conversion are in those smaller uncommitted overwriting config files...

aaron-kaplan:
Do you have customized config files for different countries? If so, do you have a place to keep them?

aaron-kaplan (approved):

remicousin:
They sit untracked in my repo on devi where I test...

PR #195, opened 2022-10-06
mousing over map prints label of highest checked admin overlay

remicousin:
This will print under the map the "label" of the highest level of admin overlay that is being checked. The "label" typically contains the name of the lower levels of admin levels they belong to -- to be formed in the config file where the sql query is.

So my code starts with a default text to print, and it is overwritten by the last hover_feature entry that is not None. Maybe it's not efficient?

This is using Dash's ALL and component id as dictionary, which allows both to:
-- not crash at start as the admin layers may not be known
-- get more than one input at a time

Also I didn't do any styling. Recommendations welcome.

### DIFF
--- a/onset_maproom/maproom.py
+++ b/onset_maproom/maproom.py
@@ -3,6 +3,7 @@ import flask
 import dash
 from dash import dcc
 from dash import html
+from dash import ALL
 import dash_bootstrap_components as dbc
 from dash.dependencies import Output, Input, State
 import dash_leaflet as dlf
@@ -91,15 +92,16 @@ def adm_borders(shapes):
 
 
 def make_adm_overlay(adm_name, adm_sql, adm_color, adm_lev, adm_weight, is_checked=False):
-    border_id = f'borders_adm{adm_lev}'
+    border_id = {"type": "borders_adm", "index": adm_lev}
     return dlf.Overlay(
         dlf.GeoJSON(
             id=border_id,
             data=adm_borders(adm_sql),
             options={
-                "fill": False,
+                "fill": True,
                 "color": adm_color,
                 "weight": adm_weight,
+                "fillOpacity": 0
             },
         ),
         name=adm_name,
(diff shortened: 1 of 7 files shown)


onset_maproom/maproom.py
```diff
@@ -255,6 +257,18 @@ def map_click(click_lat_lng):
...
 
 
+@APP.callback(
+    Output("hover_feature_label", "children"),
+    Input({"type": "borders_adm", "index": ALL}, "hover_feature")
+)
+def write_hover_adm_label(adm_loc):
+    location_description = "the map will return location name"
+    for i, adm in enumerate(adm_loc):
+        if adm is not None:
```
aaron-kaplan, line 267: Why would it be `None`?

onset_maproom/maproom.py
```diff
@@ -255,6 +257,18 @@ def map_click(click_lat_lng):
     ), round(lat_lng[0],4), round(lat_lng[1],4)
 
 
+@APP.callback(
+    Output("hover_feature_label", "children"),
+    Input({"type": "borders_adm", "index": ALL}, "hover_feature")
+)
+def write_hover_adm_label(adm_loc):
+    location_description = "the map will return location name"
+    for i, adm in enumerate(adm_loc):
```
aaron-kaplan, line 266: Only the last (topmost) input is used, so there's no point using a loop. Just use `adm_loc[-1]` (unless `len(adm_loc) == 0` in which case we use the default).

aaron-kaplan (approved):
Looks good.

When changing the UI, it can be helpful for reviewers if you add a screenshot to the PR.

drewmresnick:
I didn't know about these pattern matching callbacks until now, very nice.

onset_maproom/maproom.py
drewmresnick: So for this input it is not using all three of the layers to construct the output, it is only using a single adm layer? What determines the order of inputs that you know you could always use adm_loc[-1]? Are the layers always ordered in a certain way?

remicousin: From what I observed, let's say there are 3 admin boundaries, ie 3 components with id's type = borders_adm, hover_feature returns always an array of 3. However, it does the job of returning the feature only for those that are checked, and returns None otherwise. So there is no guarantee the last would not be None and so we want the last (because their index goes from 1 to 3 as I set it up with adm_lvl) that is not None.

The only flaw in the current implementation is that if the user unchecks all the admin boxes, then mousing over the map would return "Mousing over the map will return location name" even when the map is somewhere over Ethiopia...

onset_maproom/maproom.py
remicousin: if unchecked, hover_feature returns None

onset_maproom/maproom.py
remicousin: For instance in Ethiopia we have 3 levels, by default level 1 and 2 are checked, so hover_feature returns:
[feature_1, feature_2, None]

aaron-kaplan: I see.

PR #196, opened 2022-10-06
Manage dev dependencies

aaron-kaplan:
Dependencies that we use during development, like pytest and netcdf (for the zarrification script), are now listed in `environment-dev.yml`, separately from `environment.yml` which lists the dependencies that are needed in the production environment. Lock files generated according to the instructions in README.md now generate a lockfile for the dev environment, for developers' convenience. The Dockerfile uses a different command to build an environment that excludes the dev dependencies.


fbfmaproom/Dockerfile
```diff
@@ -12,17 +12,17 @@ RUN rm -r /etc/httpd/conf.d /etc/httpd/conf.modules.d
 
 # miniconda
 RUN curl -L \
-    https://repo.anaconda.com/miniconda/Miniconda3-py39_4.9.2-Linux-x86_64.sh \
+    https://repo.anaconda.com/miniconda/Miniconda3-py39_4.12.0-Linux-x86_64.sh \
```
aaron-kaplan, line 15: Updating to the latest version of conda has nothing to do with the intent of this PR, it's just something I do whenever I modify the Dockerfile so we don't fall too far behind.

remicousin (approved):
I approve but I've not done more than skimming through the Files changed section

PR #197, opened 2022-10-08
Fix bug in average_over when some data points are missing

aaron-kaplan:
While investigating values that unexpectedly changed when we extended the spatial domain of the Ethiopia observations, I discovered two distinct bugs in average_over. One is that when the shape to be averaged over is a single pixel, and we use all_touched=True, it ends up averaging together four pixels instead of just one. This is a relatively subtle bug and I'm not fixing it right now. The second is that the denominator of the average is wrong when some of the data values are NaN. The old Ethiopia datasets were missing some data at the extreme western edge, and as a result some pixels' displayed values were wrong by a factor of 0.5. That's what this PR fixes.


aaron-kaplan:
Pushed another commit that reimplements the weighted average using xarray's `weighted`, which I think makes the code easier to read.

remicousin (approved):

PR #199, opened 2022-10-11
Exports

aaron-kaplan:
I've set up some indirection in the pingrid import path such that anything we haven't explicitly declared as public (by listing it in `__all__`) isn't importable directly from the pingrid package. Now we can see at a glance what functions applications are using, and we can start reining them in where appropriate.

fbfmaproom uses some functions that are now marked private, but only in the vulnerability layer, which will be removed soon.

Read commit by commit.


aaron-kaplan:
Not sure I understand your question exactly, but I think you're trying to understand the purpose of this change.

As long as all the applications that use pingrid are together in this one repository, we can make changes to it relatively freely, because we know exactly what needs to be changed and how to test it. Once we publish pingrid as a library and invite other people to use it, there are going to be programs that we can't fix ourselves after we make a change. So we're going to have to be more careful about distinguishing backwards-compatible changes with non-backwards compatible ones, and publish release notes explaining what changes the users of the library need to make to their programs in order to upgrade to the latest version of the library.

If we make certain implementation details private, we ensure that clients of the library won't use them directly, only indirectly via the public functions. That gives us latitude to change implementation details without breaking anyone's code. We should expose as little as possible while still supporting the desired functionality.

remicousin:
Understood.

PR #202, opened 2022-10-12
simpler step

remicousin:
So... I was working on adding Kc to the step, until I realized it didn't really belong in the step. It belongs to the loop on doy that will feed another et to the step. Similarly, it made me realize that runoff also doesn't belong to the step. It doesn't even belong to the loop on doy since we know how to compute runoff from rainfall for all times.  So this PR removed runoff from the step.

The other element of the balance is Ks, that also doesn't belong to the step, but does belong to the loop.

So I believe next step from here is to make the loop function on doy. Then progressively incorporate the elements altered in loop.

Makes sense or I am starting to loose it here?

### DIFF
--- a/onset_maproom/agronomy.py
+++ b/onset_maproom/agronomy.py
@@ -1,13 +1,13 @@
 import xarray as xr
 
-def soil_plant_water_balance(
+def soil_plant_water_bucket(
     sm_yesterday,
     peffective,
     et,
     taw
 ):
     """Compute soil-plant-water balance from yesterday to today.
-    The balance is defined as:
+    The balance is thought as a bucket with water coming in and out:
     
     `sm` (t) + `drainage` (t) = `sm` (t-1) + `peffective` (t) - `et` (t)
     
--- a/onset_maproom/agronomy.py
+++ b/onset_maproom/agronomy.py
@@ -2,10 +2,9 @@ import xarray as xr
 
 def soil_plant_water_balance(
     sm_yesterday,
-    rain,
+    peffective,
     et,
-    taw,
-    runoff=0
+    taw
 ):
     """Compute soil-plant-water balance from yesterday to today.
     The balance is defined as:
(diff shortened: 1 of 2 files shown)


aaron-kaplan:
Runoff depends only on rainfall, not on how wet the soil is already? That's unintuitive to me, but if that's the model, this change seems appropriate.

Might want to reserve the name `soil_water_balance` for the higher-level function that calls this one.

onset_maproom/agronomy.py
```diff
@@ -2,10 +2,9 @@
 
 def soil_plant_water_balance(
     sm_yesterday,
-    rain,
+    peffective,
     et,
-    taw,
-    runoff=0
+    taw
```
aaron-kaplan, line 7: A little python tidbit: a comma is allowed after the last item in a list, and in multi-line lists I usually put one, because then when you add or remove the last element the resulting diff shows one line changed instead of two.

remicousin:
The model for runoff we've been told to use so far indeed relies only on past 7 days of rainfall. Eunjin, in her code, has another runoff function where a comment can be found saying:

# Determine runoff and effective precipitation based on SCS curve number method (EJ (12/20/2019))
def Peffective_2D(PCP, CN):  # CN should be pre-defined based on land cover, hydrologic soil groups, and antecent soil moisture condition

So that could depend on soil groups, thus space, thus not part of the steps or loops. Could depend on land cover: that could vary with time so part of the loop on days but not necessarily of the step function. Then antecedent soil moisture condition: also part of the loop, but not of the step.

But all that being said, in Eunjin code, that CN thing is set as a constant so the comment is wishful thinking at this point. It's going to be hard to predict what are all the possible options for modeling this or that part of the processes. Maybe the best strategy is to do the simplest thing with what we know we need and hope for the best? If the step function is simple, might be able to add component to its? or create another one that would fit in the loop function?

PR #203, opened 2022-10-13
passing the bucket into a loop

remicousin:
For now, ignoring the possibilities of multiple years/growing-seasons.

Turns out that I forgot to clip the wb to 0 in the bucket step. wb can not be negative. I guess it would be more logical to clip et so that it can't be greater than sm(t-1) + peff(t) but that is not how Eunjin does it. She also sets wb to 0 if negative.

Also corrected some typos or practicalities.

After this PR, we can start build up the loop. So next should be Kc this time for real.

### DIFF
--- a/onset_maproom/agronomy.py
+++ b/onset_maproom/agronomy.py
@@ -2,7 +2,7 @@ import xarray as xr
 import numpy as np
 
 
-def soil_plant_water_bucket(
+def soil_plant_water_step(
     sm_yesterday,
     peffective,
     et,
@@ -61,7 +61,7 @@ def soil_plant_water_balance(
     time_dim="T",
 ):
     """Compute soil-plant-water balance day after day over a growing season.
-    See `soil_plant_water_bucket` for the step by step algorithm definition.
+    See `soil_plant_water_step` for the step by step algorithm definition.
     
     Parameters
     ----------
@@ -83,14 +83,14 @@ def soil_plant_water_balance(
         
     See Also
     --------
-    soil_plant_water_bucket
+    soil_plant_water_step
     
     """
     
     # First Step
     if np.size(et) == 1:
         et = xr.DataArray(et)
-    sm0, drainage0 = soil_plant_water_bucket(
+    sm0, drainage0 = soil_plant_water_step(
         sminit,
         peffective.isel({time_dim: 0}, drop=True),
         et.isel({time_dim: 0}, missing_dims='ignore', drop=True),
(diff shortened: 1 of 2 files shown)


onset_maproom/agronomy.py
```diff
@@ -1,10 +1,12 @@
 import xarray as xr
+import numpy as np
+
 
 def soil_plant_water_bucket(
```
aaron-kaplan, line 5: Find a better name? Maybe something with the word "step" in it?

"Bucket" is kind of cute, because buckets hold water, but kind of confusing for the same reason.

onset_maproom/agronomy.py
```diff
@@ -29,19 +31,84 @@ def soil_plant_water_bucket(
...
     sm = wb - drainage
+    return sm, drainage
+
+
+def soil_plant_water_balance(
+    peffective,
+    et,
+    taw,
+    sminit,
+    dim="T",
```
aaron-kaplan, line 61: call it `time_dim`?

onset_maproom/agronomy.py
```diff
@@ -29,19 +31,84 @@ def soil_plant_water_bucket(
...
+    sm0, drainage0 = soil_plant_water_bucket(
+        sminit,
+        peffective.isel({dim: 0}, drop=True),
+        et.isel({dim: 0}, missing_dims='ignore', drop=True),
+        taw,
+    )
+    # Give time dimension to sm and drainage    
+    sm = sm0.expand_dims(
+        {dim: peffective[dim].size}
+    ).assign_coords({dim: peffective[dim]}).copy()
```
aaron-kaplan, line 102: If I understand the docs correctly, you can collapse `expand_dims` and `assign_coords` into a single step:
```
sm = sm0.expand_dims({dim: peffective[dim]})
```

Why did you add the `copy`?

onset_maproom/agronomy.py
aaron-kaplan: Will the user of the library need to know call this function, or is it an implementation detail? If an implementation detail, let's make it "private" by either starting its name with an underscore, or adding an `__all__` list and not including this in it.

remicousin: bucket is cute, but is also how they call it, e.g.
https://pressbooks.bccampus.ca/geoglabs2020/wp-content/uploads/sites/1032/2020/08/Figure-8.1-350x292.png
more at:
https://duckduckgo.com/?q=soil+plant+water+balance+bucket+algorithm&t=ffsb&iax=images&ia=images

remicousin: Can we answer the public/private question at the end? I assume private but I don't have a good vision of those things just yet. But yeah I am guessing that at best, we might want to program another kind of step if there are other models than this one, and then we would have some options in the loop function to choose which step-model to use. Or possibly also another loop-function if not compatible.

aaron-kaplan: Oh, then I don't understand why you used this word here. Bucket seems to refer here to a volume of space, whereas this function implements a step in time.

aaron-kaplan: > Can we answer the public/private question at the end? 

Sure.

onset_maproom/agronomy.py
remicousin: if no copy, at some point, when trying to assign other time steps of sm, will get:

ValueError: Assignment destination is a view.  Do you want to .copy() array first?

And that remains true even after collapsing the two functions.

onset_maproom/agronomy.py
remicousin: you do one bucket at a time...! + it's cute indeed. Anyway, it's called _step now

aaron-kaplan (approved):

PR #207, opened 2022-10-14
working case for a seasonal forecast

remicousin:
This is a rather quick and dirty generalization of the subseas forecast for a seasonal one.

I say quick and dirty because I feel some pieces need be broken down in smaller pieces to get a chance to work for other cases. In particular, I am thinking of breaking up cpt_starts_list into at least 2 functions: one that get datetime objects from the files list; and another that converts datetimes into strings of specific formats. The thing being that the format used in the filename is not necessarily the one we want for the app; but we do need it to select files.

We also need to be more flexible for more garbage in the filenames. What we have so far look like this:
subseas: CFSv2_SubXPRCP_CCAFCST_mu_Apr_Apr-8-2022_wk4.txt
seas: NextGen_PRCPPRCP_CCAFCST_mu_Aug-Oct_Jul2022_L1.tsv

To illustrate the previous comment, see how Jul2022 is not how we'd want to print dates in the app; whereas Apr-8-2022 was fine.
The seasonal files actually have some indication of the target in it _Aug-Oct_. That's useful for us not to have to  guess what is the relationship between start, lead and target, but that makes this part of filename pattern change when we don't really care for retrieving the file name (we just need 2 of them and it's easier to use start and lead). Note that there is actually an extra _Apr_ in the subseas one, that we actually ignored because all the files we have have it. But I suspect if we had a longer series of issues, it would have changed at some point.

I took care of that in this PR, but again quick and dirty as I don't have a sense of all the functionality available to identify patterns, replace text, etc. 

I am thinking of merging this PR, provided nothing terrible in it, so that we get something that works, then work in other PRs on the improvements described above. Unless you think it'd be better to tighten things up here in this PR.

Also note that there is no obs data for the issues of 2022 so the default landing page has empty map and graphs and you need to navigate to issues prior to 2022 to have stuff happem

This is config files entry for the subseasonal BD forecast:

# Forecast options
results_path: /data/xchourio/data/BMD-S2S-Data-Example/output
y_transform: true
start_regex: \w{3}-\w{1,2}-\w{4}
start_format_in: '%b-%d-%Y'
start_format_out: '%b-%-d-%Y'
forecast_mu_file_pattern: CFSv2_SubXPRCP_CCAFCST_mu_Apr_mystartandlead.txt
forecast_var_file_pattern: CFSv2_SubXPRCP_CCAFCST_var_Apr_mystartandlead.txt
obs_file_pattern: CFSv2_SubXPRCP_CCAFCST_obs_Apr_mystartandlead.txt
hcst_file_pattern: CFSv2_SubXPRCP_CCAFCST_xvPr_Apr_mystartandlead.txt

#S,L, target dates options
target_period_length: 7
leads: #provider_ID:leadTime_value
    wk1: 1
    wk2: 8
    wk3: 15
    wk4: 22

# Viz' set up
zoom: 7

These are for the new Guatemala seasonal forecast:

# Forecast options
results_path: /data/xchourio/data/ACToday/NextGen/Seasonal/GTM/PRCP_ENACTS/output
y_transform: false
start_regex: '[A-Z][a-z]{2}\d{4}'
start_format_in: '%b%Y'
start_format_out: '%b%Y'
forecast_mu_file_pattern: NextGen_PRCPPRCP_CCAFCST_mu_*_mystartandlead.tsv
forecast_var_file_pattern: NextGen_PRCPPRCP_CCAFCST_var_*_mystartandlead.tsv
obs_file_pattern: CanSIPSv2_PRCPPRCP_CCAFCST_Obs_*_mystartandlead.txt
hcst_file_pattern: CanSIPSv2_PRCPPRCP_CCAFCST_xvPr_*_mystartandlead.txt
variable: Precipitation

#S,L, target dates options
target_period_length: 3
time_units: months
leads: #provider_ID:leadTime_value
    L1: 1

# Viz' set up
zoom: 7


subseas_flex_fcst/maproom.py
```diff
@@ -212,7 +248,10 @@ def local_plots(marker_pos, start_date, lead_time):
 
     # Forecast CDF
     fcst_q, fcst_mu = xr.broadcast(quantiles, fcst_mu)
-    fcst_dof = int(fcst_var.attrs["dof"]) #int(dofVar)
+    try:
+        fcst_dof = int(fcst_var.attrs["dof"])
+    except:
+        fcst_dof = obs["T"].size - 1
```
aaron-kaplan, line 254: I'm wondering if it's really a good idea to silently default to a fallback value rather than raising an exception because the dof metadata were missing. Under what circumstances would it not be given?

subseas_flex_fcst/maproom.py
```diff
@@ -267,11 +306,15 @@ def local_plots(marker_pos, start_date, lead_time):
         )
     )
     cdf_graph.update_traces(mode="lines", connectgaps=False)
+    if CONFIG["time_units"] == "days":
+        start_date_pretty = (pd.to_datetime(start_date)).strftime("%-d %b %Y")
+    else:
```
aaron-kaplan, line 311: I would prefer to check for "months" (or whatever) explicitly, and raise an exception if the value of `time_units` is none of the expected values.

subseas_flex_fcst/maproom.py
```diff
@@ -422,7 +468,8 @@ def fcst_tiles(tz, tx, ty, proba, variable, percentile, threshold, start_date,le
         dims = fcst_mu.rename({"X": "lon", "Y": "lat"}).dims
     # pingrid.tile wants 2D data
     ).squeeze("T")
-    fcst_cdf = fcst_cdf.squeeze("S")
+    if "S" in fcst_cdf.dims:
+        fcst_cdf = fcst_cdf.squeeze("S")
```
aaron-kaplan, line 472: Why would the S dim not be present?

subseas_flex_fcst/predictions.py
```diff
@@ -8,7 +8,7 @@
 import pandas as pd
```
aaron-kaplan, line 8: I suggest renaming predictions.py to cpt.py, or if there's some stuff that's specific to CPT and some that's not, separating the two into different files.

subseas_flex_fcst/layout.py
```diff
@@ -22,12 +22,25 @@
...
+
 
 def app_layout():
 
     # Initialization
-    fcst_mu = predictions.sel_cpt_file(DATA_PATH,CONFIG["forecast_mu_file_pattern"],list(CONFIG["leads"])[0],startDates[-1])
+    fcst_mu = predictions.sel_cpt_file(
+        DATA_PATH,
+        CONFIG["forecast_mu_file_pattern"],
+        list(CONFIG["leads"])[0],
```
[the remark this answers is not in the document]
remicousin: yes, the leads are in the country-wise overriding config files. I pasted their content in introduction (even though I see that the indenting has gone doing that). The leads's have different children values (e.g. wk1 vs L1) so they can't be in the master config, since for instance L1 would not overwrite wk1 but add a new element to leads.

subseas_flex_fcst/maproom.py
remicousin: Historically, it hasn't been given. It's somewhat recent that it has been written into the files. But I guess not recent enough for those Guatemala ACToday forecasts not to have it...

subseas_flex_fcst/layout.py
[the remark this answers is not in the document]
aaron-kaplan: Oh yeah. I suggest you commit those files to this repo for now. Eventually they'll move to country-specific repos.

subseas_flex_fcst/maproom.py
remicousin: So... it's a bit messy as of now. Most of the time coords and dims put up in the cpt DataArrays are garbage (note that if they weren't, we probably would have much less work to do in retrieving/formatting dates).

In the subseasonal Bangladesh case, there is no S in the DataArray whatsoever, but Drew's sel_cpt_file adds it (expands_dim), so it needs be squeezed later on. in the seasonal Guatemala case, there is a S coord that is not a dim... so I could not use expands_dim in sel_cpt_file, but I need not squeeze it later on because not a dim.

This is sad. We could go simple and remove the expands_dim of sel_cpt_file and figure curating cleanly those files later. OR we have sel_cpt_file return a DataArray with a correct S value in it.

subseas_flex_fcst/maproom.py
remicousin: We can't figure out the correct value. It depends on the number of modes retained by CCA (if CCA used, if not, other considerations) for the predictor and predictand. If NextGen, there are cascading CCAs (models first then their combination). I was given a formula for that case that ended up having negative dofs (or ridiculously low) in some cases so I stepped back from that. Could not get answers from forecasters. And even if we had the right calculations, I am not sure if we have more changed to get the number of modes or the dof been given.
And the correct value is going to depend on month-of-S and L, so if you're thinking of putting dof in the config or something, it's actually going to have to be a DataArray.
So the occasion will be increasingly rare but incredibly painful to get right.

My take anyway is that it should not be the app's job to tells us whether the dof is missing or not, but rather whatever libraries are going to curate data properly in the first place. A somewhat not helpful comment since all is done together at present.

But Ingrid has used dof = n-1 for years, Simon knowing it, without raising many eyebrows (mostly raising my perfectionist ones), so at the stage we're at, it might be just fine a it is.

aaron-kaplan: OK. If there's nothing we can do about it then there's no point raising an exception.

subseas_flex_fcst/maproom.py
remicousin: I did something that doesn't raise an exception. I didn't want to break everything just because we can't get the right text printed. What do you think?

subseas_flex_fcst/maproom.py
aaron-kaplan: I see (mostly). I agree that the right approach is to modify sel_cpt_file so that the thing it returns always has the same dimensions, despite variability in the shape of the original datasets. I have no opinion on whether you should do that by always providing S or never providing S (for the time being).

I didn't understand the specifics of the BD and Guatemala datasets. Let me know if you need me to look closer.

subseas_flex_fcst/maproom.py
```diff
@@ -308,8 +308,10 @@ def local_plots(marker_pos, start_date, lead_time):
     cdf_graph.update_traces(mode="lines", connectgaps=False)
     if CONFIG["time_units"] == "days":
         start_date_pretty = (pd.to_datetime(start_date)).strftime("%-d %b %Y")
-    else:
+    elif CONFIG["time_units"] == "months":
         start_date_pretty = (pd.to_datetime(start_date)).strftime("%b %Y")
+    else:
+        start_date_pretty = "some day"
```
aaron-kaplan, line 314: Is this a legitimate case? I thought we would raise an exception if the units are none of the expected ones.

remicousin: it's not a legitimate case but I didn't feel it was worth stopping the program from making the graphs just because we don't know what dates to print... Right now we are using these units for printing dates only. Maybe I have the illegitimate cases print an error title?

aaron-kaplan: My philosophy is that when something unexpected happens, it's better to fail loudly and as early as possible, rather than silently falling back to some default value. It's easier to fix a problem immediately, when you still remember what you did.

subseas_flex_fcst/predictions.py
remicousin: started that. More might split when we extract from cpt functions what is purely date formating and such

subseas_flex_fcst/maproom.py
remicousin: I went for the simplest, ignoring the garbage in cpt DataArray, not trying to correct it, then squeezing everything. We'll have to figure that out at some point, but maybe first figure out if we'll get cpt DataArray that are not garbage on that front and figure out then how to identify the cases and deal with them.

subseas_flex_fcst/cpt.py
```diff
@@ -0,0 +1,95 @@
...
+        CFSv2_SubXPRCP_CCAFCST_mu_Apr_Apr-1-2022_wk1.txt
+    And where this file's `lead_time` and `start_date`:
+        `lead_time` == 'wk1' and `start_date` == 'Apr-1-2022'
+    `filename_pattern` == 'CFSv2_SubXPRCP_CCAFCST_mu_Apr_*.txt'
+    """
+    pattern = f"{start_date}_{lead_time}"
+    full_path = f"{data_path}/{filename_pattern}"
+    expanded_name = glob.glob(full_path.replace("mystartandlead",pattern))
+    if len(expanded_name) == 0:
+        file_selected = xr.Dataset().merge(xr.DataArray(np.nan).rename("missing_file"))
```
aaron-kaplan, line 44: What's the advantage of returning a 0-dimensional Dataset, compared to just returning None? I can imagine that it might simplify some code somewhere, but the way you've used it it seems equivalent, so I would prefer the simpler version.

subseas_flex_fcst/cpt.py
```diff
@@ -0,0 +1,95 @@
...
+    filename_pattern = filename_pattern.replace("mystartandlead", "*")
+    files_name_list = glob.glob(f'{data_path}/{filename_pattern}')
+    start_dates = []
+    for file in files_name_list:
+        start_date = re.search(regex_search_pattern, file)
+        start_date_dt = datetime.strptime(start_date.group(), format_in)
+        start_dates.append(start_date_dt)
+    start_dates = sorted(set(start_dates)) #finds unique dates in the case there are files with the same date due to multiple lead times
+    start_dates = [i.strftime(format_out) for i in start_dates]
+    return start_dates
```
aaron-kaplan, line 95: Seeing this CPT-specific stuff makes me lean more towards zarrification than I did when you asked earlier today. I'm ok with this version, but if you wanted to convert to zarr first and then use normal xarray operations instead of `sel_file` and `starts_list`, I would be ok with that too.

subseas_flex_fcst/cpt.py
```diff
@@ -0,0 +1,95 @@
...
+    `filename_pattern` == 'CFSv2_SubXPRCP_CCAFCST_mu_Apr_*.txt'
+    """
+    pattern = f"{start_date}_{lead_time}"
+    full_path = f"{data_path}/{filename_pattern}"
+    expanded_name = glob.glob(full_path.replace("mystartandlead",pattern))
+    if len(expanded_name) == 0:
+        file_selected = xr.Dataset().merge(xr.DataArray(np.nan).rename("missing_file"))
+    else:
+        file_name = expanded_name[0]
+        file_selected = cptio.open_cptdataset(file_name)
```
aaron-kaplan, line 47: Misleading variable name--I think it's a DataArray or Dataset, not a file?

subseas_flex_fcst/cpt.py
remicousin: That's a bit beyond what I can afford to do with this PR now.

subseas_flex_fcst/cpt.py
remicousin: yeah... must stuff was expecting a Dataset out of this. I changed that to handle None. There is still more to do here as maproom's read_cptdataset needs to move elsewhere and likely be revisited together with the function it uses. But I can't afford that now.

subseas_flex_fcst/cpt.py
remicousin: renamed along with function in last commit

subseas_flex_fcst/cpt.py
remicousin: yeah... this will have to be considered at another time/PR I am afraid.

remicousin:
I don't think I won't be able to make anything more on this this week and before the first User training. But I can create a list of issues to attend to.

subseas_flex_fcst/maproom.py
```diff
@@ -166,6 +194,14 @@ def pick_location(n_clicks, click_lat_lng, latitude, longitude):
     Input("lead_time","value"),
 )
 def local_plots(marker_pos, start_date, lead_time):
+    # Time Units Errors handling
+    if CONFIG["time_units"] == "days":
+        start_date_pretty = (pd.to_datetime(start_date)).strftime("%-d %b %Y")
+    elif CONFIG["time_units"] == "months":
+        start_date_pretty = (pd.to_datetime(start_date)).strftime("%b %Y")
+    else:
+        error_fig = pingrid.error_fig(error_msg="Unexpected time_units: check config")
```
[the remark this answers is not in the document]
remicousin: I forgot that one. I'll have to figure that out later some time this week somehow.

aaron-kaplan (approved):
One comment to be addressed later: https://github.com/iridl/python-maprooms/pull/207#discussion_r999821139

> Assuming the user and the developer are different people, this is an error that the user can't fix, so I would put the diagnostic information in the log file rather than the browser. For now, we do that either by raising an exception or with a simple print.

PR #218, opened 2022-11-01
Remic/zmdfcs2

remicousin:
This takes care of the ZMD case. It's not pretty but can't afford much more at this stage. I actually thought it was going to be uglier. The deal with ZMD is that their filenames contain no lead information (Start and target identify the forecasts). At first, I was trying to pretend make some list of leads anyway in the config file and adapt accordingly, but I could not wrap my head around it. So instead it's a case where S and target are given instead of S and leads...

I'd be interested to correct what is too gross, then leave more satisfying set up after the trainings.

### DIFF
--- a/subseas_flex_fcst/config-gt.yaml
+++ b/subseas_flex_fcst/config-gt.yaml
@@ -1,8 +1,7 @@
 # App set up
-server: 0.0.0.0
 
 # Forecast options
-results_path: /data/xchourio/data/ACToday/NextGen/Seasonal/GTM/PRCP_ENACTS/output
+forecast_path: /data/xchourio/data/ACToday/NextGen/Seasonal/GTM/PRCP_ENACTS/output
 y_transform: false
 start_regex: '[A-Z][a-z]{2}\d{4}'
 start_format_in: '%b%Y'
@@ -18,7 +17,7 @@ target_period_length: 3
 time_units: months
 leads: #provider_ID:leadTime_value
     L1: 1
-targets: none
+targets: null
 
 # Viz' set up
 zoom: 7
--- a/subseas_flex_fcst/config-gt.yaml
+++ b/subseas_flex_fcst/config-gt.yaml
@@ -1,4 +1,5 @@
 # App set up
+server: 0.0.0.0
 
 # Forecast options
 results_path: /data/xchourio/data/ACToday/NextGen/Seasonal/GTM/PRCP_ENACTS/output
@@ -6,10 +7,10 @@ y_transform: false
 start_regex: '[A-Z][a-z]{2}\d{4}'
 start_format_in: '%b%Y'
 start_format_out: '%b%Y'
-forecast_mu_file_pattern: NextGen_PRCPPRCP_CCAFCST_mu_*_mystartandlead.tsv
-forecast_var_file_pattern: NextGen_PRCPPRCP_CCAFCST_var_*_mystartandlead.tsv
-obs_file_pattern: CanSIPSv2_PRCPPRCP_CCAFCST_Obs_*_mystartandlead.txt 
-hcst_file_pattern: CanSIPSv2_PRCPPRCP_CCAFCST_xvPr_*_mystartandlead.txt
+forecast_mu_file_pattern: NextGen_PRCPPRCP_CCAFCST_mu_*_SLtarget.tsv
+forecast_var_file_pattern: NextGen_PRCPPRCP_CCAFCST_var_*_SLtarget.tsv
+obs_file_pattern: CanSIPSv2_PRCPPRCP_CCAFCST_Obs_*_SLtarget.txt 
+hcst_file_pattern: CanSIPSv2_PRCPPRCP_CCAFCST_xvPr_*_SLtarget.txt
 variable: Precipitation
 
 #S,L, target dates options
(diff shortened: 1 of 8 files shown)


remicousin:
ZMD next gen data you can reach at:
/data/remic/mydatafiles/ZMD/nexgenzmd/

remicousin:
And I forgot to mention that the "obs" files have a different file name pattern that the rest of the suite. So that is taken care of too...

remicousin:
I forgot to update the documentation of the functions that I have changed. Next on my to-do list but that might not happen today.

subseas_flex_fcst/config-zmd.yaml
```diff
@@ -0,0 +1,24 @@
+# App set up
+server: 0.0.0.0
+port: 8063
```
aaron-kaplan, line 3: Why does this one specify a port and the others don't?

subseas_flex_fcst/config-zmd.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+# Forecast options
+results_path: /data/remic/mydatafiles/ZMD/nextgenzmd
```
aaron-kaplan, line 6: It's confusing to call this "results" when it's the input to the application. `forecast_path`?

subseas_flex_fcst/config-gt.yaml
```diff
@@ -1,22 +1,24 @@
...
+obs_file_pattern: CanSIPSv2_PRCPPRCP_CCAFCST_Obs_*_SLtarget.txt 
+hcst_file_pattern: CanSIPSv2_PRCPPRCP_CCAFCST_xvPr_*_SLtarget.txt
 variable: Precipitation
 
 #S,L, target dates options
 target_period_length: 3
 time_units: months
 leads: #provider_ID:leadTime_value
     L1: 1
+targets: none
```
aaron-kaplan, line 21: Use `null` (the null value) instead of `"none"` (a string of four characters), here and in the other config files.

subseas_flex_fcst/layout.py
```diff
@@ -29,11 +29,18 @@
 def app_layout():
 
     # Initialization
+    if CONFIG["leads"] != "none":
```
aaron-kaplan, line 32: If you make the change I suggested in the config file, this will change to `is not None`.

subseas_flex_fcst/layout.py
```diff
@@ -29,11 +29,18 @@
...
+        use_leads = list(CONFIG["leads"])[0]
+        use_targets = None
+    else:
+        use_leads = None
+        use_targets = CONFIG["targets"][1]
```
aaron-kaplan, line 37: When there are two mutually-exclusive options, I like to raise an exception if both are specified, rather than silently ignoring one of them. It makes debugging easier.

subseas_flex_fcst/maproom.py
```diff
@@ -116,20 +126,29 @@ def display_relevant_control(variable):
     Input("start_date","value"),
 )
 def target_range_options(start_date):
-    leads_values = list(CONFIG["leads"].values())
-    leads_keys = list(CONFIG["leads"])
+    if CONFIG["leads"] != "none":
```
aaron-kaplan, line 129: becomes `is not None`

subseas_flex_fcst/maproom.py
```diff
@@ -42,29 +42,38 @@
 #Should I move this function into the predictions.py file where I put the other funcs?
 #if we do so maybe I should redo the func to be more flexible since it is hard coded to read each file separately..
 def read_cptdataset(lead_time, start_date, y_transform=CONFIG["y_transform"]):
+    if CONFIG["leads"] != "none":
```
aaron-kaplan, line 45: becomes `is not None`

subseas_flex_fcst/maproom.py
```diff
@@ -232,13 +258,16 @@ def local_plots(marker_pos, start_date, lead_time):
...
         return error_fig, error_fig
 
-    target_range = predictions.target_range_format(
-        CONFIG["leads"][lead_time],
-        lead_time,
-        pd.to_datetime(start_date),
-        CONFIG["target_period_length"],
-        CONFIG["time_units"],
-    )
+    if CONFIG["leads"] != "none":
```
aaron-kaplan, line 261: `is not None`

aaron-kaplan (changes requested):
The comment about the server address is important to fix (or discuss) now, everything else can wait.

I had been told that PyCPT output file names had been standardized so we didn't have to bend ourselves into pretzels like this anymore. Either that's not true, or ZMD is using a pretty old version of PyCPT.

subseas_flex_fcst/config-zmd.yaml
remicousin: no reason. I wasn't paying attention to changed to the country specific config files

subseas_flex_fcst/layout.py
remicousin: Let me know if that is not the way to go

remicousin:
It may be that the GT/BD files are old stuff and thus are the ones that have not standardized CPT filenames, while ZMD has.

aaron-kaplan (approved):
Just one remaining comment. I may merge this before you get to address it, so I can move ahead with the muli-page configuration.

PR #219, opened 2022-11-02
Fix incorrect display of values outside colormap range

aaron-kaplan:
Screenshots from the KMD monthly climo maproom (which is not in this repo; I already made the fix in the KMD copy):
Before:
Screenshot from 2022-11-01 15-14-41
After:
Screenshot from 2022-11-01 15-30-41


remicousin (approved):
I approve after skimming through. I don't have much bandwidth (both literally and internetly) to do much more. Can try and spend a bit more time on it once back at hotel this evening.

PR #220, opened 2022-11-02
Working forecast maproom config for BD

aaron-kaplan:
There are three config files in this directory (bd, gt, and sample) and none of them work. Each is missing different required fields. I got this one working in order to test a pingrid change.


remicousin:
I am sorry. You are supposed to use the country specific config to overwrite sample. This being said I didn't pay much attention of keeping the country specific config files straight as my open PR showed. Let me try and fix that and this PR should be unnecessary.

aaron-kaplan:
Oh yeah, of course.

PR #223, opened 2022-11-02
Remic/adm

remicousin:
I actually can't test this in my room... connection is too weak and I need to go through devi. That's why I didn't finished last night actually!!
Was tested this morning on gt and bd.
Rebased just now without significant conflict.
Pushing now for someone to put fresh eyes on it.
Going to dinner and then a place with better internet to wrap up.

### DIFF
--- a/subseas_flex_fcst/config-bd.yaml
+++ b/subseas_flex_fcst/config-bd.yaml
@@ -1,6 +1,4 @@
 # App set up
-mode: prod
-server: 0.0.0.0
 
 # Forecast options
 forecast_path: /data/xchourio/data/BMD-S2S-Data-Example/output
--- a/subseas_flex_fcst/config-bd.yaml
+++ b/subseas_flex_fcst/config-bd.yaml
@@ -1,4 +1,6 @@
 # App set up
+mode: prod
+server: 0.0.0.0
 
 # Forecast options
 forecast_path: /data/xchourio/data/BMD-S2S-Data-Example/output
@@ -23,3 +25,21 @@ targets: null
 
 # Viz' set up
 zoom: 7
+
+# Bangladesh
+shapes_adm:
+       - name: Divisions
+         color: black
+         sql: select gid as key, name_1 as label, ST_AsBinary(the_geom) as the_geom
+                   from bangladesh_bmd_divisions
+         is_checked: True
+       - name: Districts
+         color: black                      
+         sql: select gid as key, (name_2 || ', ' || name_1) as label, ST_AsBinary(the_geom) as the_geom
+                   from bangladesh_bmd_districts
+         is_checked: True
+       - name: Upazillas
+         color: grey
+         sql: select gid as key, (name_3 || ', ' || name_2 || ', ' || name_1) as label, ST_AsBinary(the_geom) as the_geom
+                   from bangladesh_bmd_upazilas
+         is_checked: False
--- a/subseas_flex_fcst/config-bd.yaml
+++ b/subseas_flex_fcst/config-bd.yaml
@@ -26,18 +26,18 @@ zoom: 7
 
 # Bangladesh
 shapes_adm:
-       - name: Divisions
-         color: black
-         sql: select gid as key, name_1 as label, ST_AsBinary(the_geom) as the_geom
-                   from bangladesh_bmd_divisions
-         is_checked: True
-       - name: Districts
-         color: black                      
-         sql: select gid as key, (name_2 || ', ' || name_1) as label, ST_AsBinary(the_geom) as the_geom
-                   from bangladesh_bmd_districts
-         is_checked: True
-       - name: Upazillas
-         color: grey
-         sql: select gid as key, (name_3 || ', ' || name_2 || ', ' || name_1) as label, ST_AsBinary(the_geom) as the_geom
-                   from bangladesh_bmd_upazilas
-         is_checked: False
+     - name: Divisions
+       color: black
+       sql: select gid as key, name_1 as label, ST_AsBinary(the_geom) as the_geom
+           from bangladesh_bmd_divisions
+       is_checked: True
+     - name: Districts
+       color: black                      
+       sql: select gid as key, (name_2 || ', ' || name_1) as label, ST_AsBinary(the_geom) as the_geom
+           from bangladesh_bmd_districts
+       is_checked: True
+     - name: Upazillas
+       color: grey
+       sql: select gid as key, (name_3 || ', ' || name_2 || ', ' || name_1) as label, ST_AsBinary(the_geom) as the_geom
+           from bangladesh_bmd_upazilas
+       is_checked: False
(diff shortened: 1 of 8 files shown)


drewmresnick:
I get a KeyError for `core_path` in the config files. Should that be the path that is currently called `forecast_path` in the config files?

aaron-kaplan:
@drewmresnick I had the same problem yesterday and Rémi set me straight in  https://github.com/iridl/python-maprooms/pull/220 . Use the sample config file *and* a country-specific file, e.g.
```
CONFIG=config-sample.yaml:config-bd.yaml
```

subseas_flex_fcst/config-bd.yaml
```diff
@@ -23,3 +23,21 @@ targets: null
 
 # Viz' set up
 zoom: 7
+
+# Bangladesh
+shapes_adm:
+       - name: Divisions
```
aaron-kaplan, line 29: This is indented by eight spaces. In other yaml files we mostly use four.

Are you using a different editor than usual? Maybe it can be configured.

subseas_flex_fcst/maproom.py
```diff
@@ -39,6 +46,53 @@
...
+        lambda x: x if isinstance(x, MultiPolygon) else MultiPolygon([x])
+    )
+    shapes = df["the_geom"].apply(shapely.geometry.mapping)
+    for i in df.index: #this adds the district layer as a label in the dict
+        shapes[i]['label'] = df['label'][i]
+    return {"features": shapes}
+
+
+def make_adm_overlay(adm_name, adm_sql, adm_color, adm_lev, adm_weight, is_checked=False):
+    border_id = {"type": "borders_adm", "index": adm_lev}
```
aaron-kaplan, line 79: Why did you use a compound id instead of a string? The only reason I know of is for use with a pattern-matching callback, but I don't see any of those in this PR.

I think you can just not assign this component an id at all. That would let you eliminate the `adm_lev` parameter.

subseas_flex_fcst/maproom.py
```diff
@@ -39,6 +46,53 @@
...
+                    """
+                ),
+            ]
+        )
+        df = pd.read_sql(s, conn)
+
+    df["the_geom"] = df["the_geom"].apply(lambda x: wkb.loads(x.tobytes()))
+    df["the_geom"] = df["the_geom"].apply(
+        lambda x: x if isinstance(x, MultiPolygon) else MultiPolygon([x])
+    )
```
aaron-kaplan, line 71: This step (69-71) should no longer be necessary, since https://github.com/iridl/python-maprooms/pull/198 . Please verify.

subseas_flex_fcst/maproom.py
```diff
@@ -39,6 +46,53 @@
...
+    shapes = df["the_geom"].apply(shapely.geometry.mapping)
+    for i in df.index: #this adds the district layer as a label in the dict
+        shapes[i]['label'] = df['label'][i]
```
aaron-kaplan, line 74: `shapes` is a dataframe, so I think 73-74 can be replaced with
```
shapes['label'] = df['label']
```

subseas_flex_fcst/config-bd.yaml
```diff
@@ -23,3 +23,21 @@ targets: null
...
+         color: black
+         sql: select gid as key, name_1 as label, ST_AsBinary(the_geom) as the_geom
```
aaron-kaplan, line 31: You don't seem to use the key anywhere, so you could drop it from the query. If you think we're likly to need it in the future, I have no objection to keeping it.

subseas_flex_fcst/maproom.py
```diff
@@ -39,6 +46,53 @@
...
+    with psycopg2.connect(**CONFIG["db"]) as conn:
+        s = sql.Composed(
+            [
+                sql.SQL("with g as ("),
+                sql.SQL(shapes),
+                sql.SQL(
+                    """
+                    )
+                    select
+                        g.label, g.key, g.the_geom
```
aaron-kaplan, line 60: If you remove the key from the SQL query in the config file, also remove `g.key` here.

aaron-kaplan (approved):
Make whatever changes you can make quickly, then merge so I can do the big reorg.

remicousin:
Starting looking into these now.

subseas_flex_fcst/maproom.py
remicousin: That's because it's coming form the onset maproom set up that is using this to print the name of the finest admin level label when mousing hover. So it will be needed or used when that part is implemented. It's just not there yet. That's likely a feature we'd want very often if not standard, so shall we keep this here when the mouse hover is put up?

subseas_flex_fcst/config-bd.yaml
remicousin: How are we going to select polygons in the future, to do for instance weight-average of the gridboxes over the polygons? Is that when it might be used?

subseas_flex_fcst/config-bd.yaml
remicousin: Only fixed this for now. I don't seem to be getting maps through devi on the hotel internet. Pushing to try my chance directly on my mac.

remicousin:
I don't think I can test any significant changed from the hotel... For some reason, when in prod mode, I get never ending 
129.236.110.237 - - [02/Nov/2022 14:54:09] "GET /flex-fcst-maproom/_reload-hash HTTP/1.1" 200 -
one after the other and no map loads.
I though that the connection was maybe too slow for working through devi to work... Adding the admins does seem to slow things down... so I pushed to work on my Mac... but then I realized again why I was working on devi: because I can't access iridb from my Mac...

So I don't know what to do... I pushed the cosmetics on the indenting for the config files.

I commented
    df["the_geom"] = df["the_geom"].apply(lambda x: wkb.loads(x.tobytes()))
    df["the_geom"] = df["the_geom"].apply(
        lambda x: x if isinstance(x, MultiPolygon) else MultiPolygon([x])
    )

and got an error right away so restored it. But other than that, I just get blank maps and the reload-hash without anything else happening... when on the Hotel wifi.

I got the bd and gt forecasts with their admin boundaries this morning from an ethernet internet connection at ZMD (somehow I got one -- that's the trick: they have internet but no wifi, and few working cabled connections....).

Let me know what you think we can do from there

subseas_flex_fcst/maproom.py
aaron-kaplan: ok

subseas_flex_fcst/config-bd.yaml
aaron-kaplan: yes

aaron-kaplan:
> I commented
> df["the_geom"] = df["the_geom"].apply(lambda x: wkb.loads(x.tobytes()))
> df["the_geom"] = df["the_geom"].apply(
> lambda x: x if isinstance(x, MultiPolygon) else MultiPolygon([x])
> )
> 
> and got an error right away so restored it. 

I didn't mean for you to delete the wkb.loads line, only the following ones. Not important right now. I will test on devi.

PR #231, opened 2022-11-16
Remic/conf

remicousin:
This is probably not what Aaron had in mine.... he will tell.

works with:

CONFIG=config-zmd.yaml:config.yaml FLEX_FCST_CONFIG=flex_fcst/config-zmd.yaml ONSET_CONFIG=onset/config-zmd.yaml MONTHLY_CONFIG=monthly/config-zmd.yaml python homepage.py

where config.yaml has an appropriate server, host and db/password.

I could not figure out how to control debug. Right now it is in prod mode. Sometimes convenient, sometimes not.


remicousin:
I'll update the README as final step after approval and before merging.

remicousin:
On a side note, I see also now that there "assets" in onset that are mostly country-specific rather than app specific and that should likely go one level down, if it's allowed.

aaron-kaplan:
What I had in mind was one top-level config file per country and no config files in the subdirectories, moving what's currently in the subdir configs into subsections of the top-level config. If you think that's a bad idea, I'm open to discussing it.

remicousin:
We are just going to have to be careful on how we name config parameters through maprooms. Unless we add a level of nesting for each maproom in the config?

remicousin:
Might have to be nested because once we have dozens of Maprooms it will be painfull to figure out what has been named...

aaron-kaplan:
Yes, nesting is what I meant when I said “section”. Have a top-level entry
called “onset”, whose children are config variables for the onset maproom.

remicousin:
Something more like this?

remicousin:
BTW I think I screwed up my commits by working both on Mac and devi... But hopefully I fell back on my feet.

aaron-kaplan (changes requested):
This looks good except for the fact that there are two variables called CONFIG and CFG, and it's going to be hard to remember what the difference is. I sugggest GLOBAL_CONFIG and CONFIG.

Personally, I would also prefer to have a single developer- and site-specific config file containing mode, server, port, and db rather than repeating that info in each country's config file, but this is a matter of taste that I won't insist on if you feel otherwise.

aaron-kaplan (approved):

Issue #151, opened 2022-07-28
Need something to make dynamic colorscales


remicousin:
Needed for flexible forecast maprooms. Has not been needed elsewhere so far in Ingrid's

remicousin:
Resolved thanks to PR #315

PR #244, opened 2022-11-30
Ingrid colormaps are BGR and so all will be

remicousin:
Turned out to be a bit tedious to apply the right changes and rebase revcs so here is a brand new PR for this. Tested on all Maprooms (fbf and 3 enacts ones)


pingrid/impl.py
```diff
@@ -51,11 +51,11 @@
 from datetime import datetime, timedelta
 
 
-RAINBOW_COLORMAP = "[0x0000ff [0x00ffff 51] [0x00ff00 51] [0xffff00 51] [0xff0000 51] [0xff00ff 51]]"
+RAINBOW_COLORMAP = "[null 1973790 16711680 [16776960 51] [65280 51] [65535 51] [255 51] [16711935 51] 12632256]"
```
aaron-kaplan, line 54: Besides switching the byte order, you also added an extra entry at the start and another at the end. I'm guessing these are the colors to be used for values below the minimum and above the maximum, respectively. But parse_colormap doesn't seem to have a special case for those, so I'm not sure they'll be interpreted as intended. I'm thinking Igor might have removed them on purpose. Have you looked into this?

remicousin: I haven't... you're probably right. And then there are 3 values added to be removed: 2 at the start and 1 at the end, who respectively are the colors for missing, below min and above max.

remicousin:
The colorscales look fine without the extra Ingrid tailing colors

enacts/onset/maproom.py
```diff
@@ -664,10 +663,9 @@ def onset_tile(tz, tx, ty):
 )
 def set_colorbar(search_start_day, search_start_month, search_days, map_choice):
     if map_choice == "pe":
-        thresholds = pingrid.CORRELATION_COLORMAP_KEY
         return (
             f"Probabily of onset date to be {search_days} past {search_start_day} {search_start_month}",
-            pingrid.to_dash_colorscale(pingrid.CORRELATION_COLORMAP, thresholds=thresholds),
```
aaron-kaplan, line 670: Why is this no longer needed?

remicousin: I think I had implemented this to have dynamic colorscales (which are needed for the flexible forecast) but we walked away from that for another reason but when walking away, it seems I missed out removing these.

We still need some sort of dynamic colorscale capacity. See Issue #151 . But these pieces are not part of it. Or if they will be, they'll come back but they don't belong here and now.

aaron-kaplan:
This looks fine, but as FIST has workshops coming up soon I want to do a careful double-check before approving. I hope to get to that later this afternoon.

remicousin:
absolutely

aaron-kaplan (approved):

PR #246, opened 2022-12-05
Defining colorscales as human can understand

remicousin:
Proposing a framework to define colorscales that is somewhat more human-readable friendly.

The colors definitions come from [ingrid](https://bitbucket.org/iridl/ingrid/src/master/src/ingridinit.tex) (search for instance DarkOrange).

Colors and colorscales can be pre-determined in pingrid or locally for a Maproom. 

I am not sure what to do of the alpha entry.

Note the changes in flex_fcst/maproom.py and some in pingrid are to enable testing without breaking the other colorscales.

CORRELATIONCS can be invoked in the Flex Fcst Maproom by asking the probability of non-/exceeding Precipitation (as opposed to a percentile) (you can run the multipage Maprooms on devi for ZMD configuration)

Next PR would be to convert all current colorscales to this framework. Then we should stick to it. And then I can implement the dynamically changing colorscales for Flex Fcst Maproom (for probability of non-/exceeding percentiles).

I am requesting everyone's review here: Aaron and Kevin for implementation; all others to comment on the concept.


aaron-kaplan:
Please take a look at the documentation for matplotlib colorscales. It would be nice to have interoperability with matplotlib if possible. (I'm not saying that what you've done is incompatible with matplotlib; I haven't looked yet.)

remicousin:
hmmm.... looking into matplotlib: why are we not just using matplotlib in the first place?

aaron-kaplan:
I used matplotlib in my initial uicore + python experiments, pre-dash, and it seemed fine. But when Igor migrated the FBF application to python, he tried matplotlib and found it far too slow, so he reimplemented a lot of its functionality himself on top of opencv. I haven't taken the time to see for myself how the performance compares, but I would like to when I get the chance, because I would rather use a mature, well-tested implementation than maintain our own geo plotting library. That just hasn't made it to the top of the priority list yet.

aaron-kaplan:
...but that's why I would like to use a colormap representation that will make it easy to switch to matplotlib if that is otherwise feasible.

remicousin:
I see... so matplotlib seems to be extremely rich so it seems likely that whatever we do, a few steps will convert that into something matplotlib-ish.

More specifically, with respect to colors definitions, it seems matplotlib understands a ton of [ways](https://matplotlib.org/stable/gallery/color/color_demo.html) to describe colors. Looking at the source code for, say aquamarine, the color is defined as 'aquamarine': '#04d8b2' and 'aquamarine': '#7FFFD4', but are actually 2 different colors: one from XKCD and the other from CSS4 "conventions."

But their colorscales are either defined as
-- 3x256 arrays of RGB values scales between 0 and 1
-- or 3xN with N<256 and I guess there is functionality to extrapolate equidistantly on 256 array
-- qualitative (or categorical, or banded) colorscales are also a 3xN but matplotlib is told not to blend from colors to colors but to make bands
-- a list of anchors or nodes with colors defined on each side of the anchor/node, e.g.:
_autumn_data = {'red':   ((0., 1.0, 1.0), (1.0, 1.0, 1.0)),
                'green': ((0., 0., 0.), (1.0, 1.0, 1.0)),
                'blue':  ((0., 0., 0.), (1.0, 0., 0.))}
-- but matplotlib doc indicated the following to create a colorscale through anchor points:
colors = ["darkorange", "gold", "lawngreen", "lightseagreen"]
nodes = [0.0, 0.4, 0.8, 1.0]
cmap2 = LinearSegmentedColormap.from_list("mycmap", list(zip(nodes, colors)))
-- while creating banded colorscale is a simple as:
cmap = ListedColormap(["darkorange", "gold", "lawngreen", "lightseagreen"])

So anyway... it seems to be very versatile.

pingrid/impl.py
```diff
@@ -402,6 +412,69 @@ def produce_shape_tile(
...
+    BLUE: [-0.8],
+    DEEPSKYBLUE: [-0.6],
+    AQUAMARINE: [-0.3],
+    PALEGREEN: [-0.1],
+    MOCCASIN: [-0.1, 0.1],
+    YELLOW: [0.1],
+    DARKORANGE: [0.4],
+    RED: [0.7],
+    DARKRED: [1],
+}
```
aaron-kaplan, line 438: A dictionary doesn't seem like the best representation. When you iterate through the items in the dictionary,  you get them in the order they were inserted, which is not necessarily the same as the order of the values. What happens if someone accidentally/unknowingly inserts values in the wrong order?

Let's define a ColorScale class, so we can put validation logic in the constructor and catch errors like that. The first argument could be a list of colors, and the second a list of thresholds, which is optional.

Secondarily, do we really need the special case where MOCCASIN gets two thresholds instead of one? Could we instead just repeat MOCCASIN twice in succession, with two different thresholds?

I don't know how familiar you are with defining classes. Let me know if you want more guidance.

pingrid/impl.py
```diff
@@ -402,6 +412,69 @@ def produce_shape_tile(
...
+DARKRED = BGRA(blue=0, green=0, red=128, alpha=255)
+DEEPSKYBLUE = BGRA(blue=255, green=191, red=0, alpha=255)
+MOCCASIN = BGRA(blue=181, green=228, red=255, alpha=255)
+NAVY = BGRA(blue=128, green=0, red=0, alpha=255)
+PALEGREEN = BGRA(blue=152, green=251, red=152, alpha=255)
+RED = BGRA(blue=0, green=0, red=255, alpha=255)
+YELLOW = BGRA(blue=0, green=255, red=255, alpha=255)
+
+
+CORRELATIONCS = {
```
aaron-kaplan, line 427: Call it CORRELATION_CS OR CORRELATION_COLORSCALE

aaron-kaplan (commented):
While we're working on this, I would like to make some other changes.

I would like pingrid's color order to be RGB instead of BGR. Igor used BGR because  he was using OpenCV and OpenCV uses BGR (apparently for historical reasons). But I don't want to tie us to OpenCV forever, and the rest of the software world has all standardized on RGB. Clients of pingrid should express their colors in RGB, and then parts of the code that are specific to OpenCV should convert it to BGR as necessary.

It would be nice not to have to specify alpha when defining a color. If we make RGBA a dataclass instead of NamedTuple, we can have alpha default to 255 when not specified. Something like
```
@dataclass
class RGBA:
    red: int
    green: int
    blue: int
    alpha: int = 255
```

aaron-kaplan:
I think it would be useful to retain the function for parsing Ingrid netcdf colorscale attributes. Don't get rid of that when you convert everything to the new color representation.

pingrid/impl.py
remicousin: I will give it a shot. I could not repeat MOCCASSIN with the dictionary format.

aaron-kaplan: Oh, right. Another reason not to use a dictionary.

remicousin: On a side note: are we not reinventing the wheel here? Even if matplotlib is slow to make maps, we could still use it just for colors and colorscales, couldn't we?

aaron-kaplan: It's a reasonable question. Some factors to consider:
- There is a maintenance burden either way. If we write our own code, we have to debug it and extend it ourselves, but if we use someone else's we have to upgrade periodically and test the upgrades to make sure they don't break anything. Generally if it's simple it's better to write it ourselves, if it's complicated and there's a popular, mature alternative then it's better to use that.
- Matplotlib's colormap functionality is quite full-featured, whereas opencv's is more limited. If we want to translate matplotlib colormaps in their full generality to opencv then it seems like the translation code, which we own, will have to be quite complex. If we want to restrict ourselves to a limited subset of matplotlib's functionality in order to keep the complexity down, the best way to do that is to write our own colormap class that supports only the functionality we want to allow.
- All else being equal, I would prefer to leave options open rather than committing to using any particular library. If we tell people that the way to create a colormap in python maprooms is to import matplotlib, then we can't drop matplotlib later without breaking everyone's code. If we tell them to use a pingrid class, then even if that class uses matplotlib behind the scenes, we can drop matplotlib and reimplement the pingrid class later.

Sometimes we will expose a third-party API directly rather than hiding it from our users because it's so extensive and powerful that there's no reasonable way to hide it. This is the case with xarray, for example. It's not clear to me yet whether matplotlib is in that category, but until we have evidence that it is, I'd like to assume that it's not.

remicousin: I gave it a shot. Still meant to applying to the CORRELATION colorscale used in flex. fcst maproom. I kept the function I created previously as I barely know what I am doing.
Will clean up and apply to all colorscales once we are done with that one as case study

pingrid/impl.py
```diff
@@ -529,8 +552,8 @@ def to_dash_colorscale(s: str) -> List[str]:
     return cs
 
 
-def to_dash_colorscale2(s: str) -> List[str]:
-    cm = colorscale2cv2colormap(s)
+def to_dash_colorscale2(s):
+    cm = s.cv2cm()
```
aaron-kaplan, line 556: Should be able to go straight from the pingrid RGBA representation to the dash representation without passing through the opencv representation. Interdependencies would make it harder to remove one of them in the future.

pingrid/impl.py
```diff
@@ -464,14 +486,15 @@ def colorscale2cv2colormap(colorscale):
     ])
     cv2cm_i = (255*(cs_val-cs_val[0])/(cs_val[-1]-cs_val[0])).astype(int)
     cv2cm_i = cv2cm_i + np.append([0], np.where(np.diff(cv2cm_i) == 0, 1, 0))
-    cv2cm_brga = np.array([
+    cv2cm_rgba = np.array([
         list(colorscale.keys())[i]
         for i in range(len(colorscale))
         for j in range(len(list(colorscale.values())[i]))
     ])
+    cv2cm_bgra = cv2cm_rgba[:,[2, 1, 0, 3]]
```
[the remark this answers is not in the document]
remicousin: colorscale2cv2colormap is going to the trash if the rest if good.

pingrid/impl.py
```diff
@@ -464,14 +486,15 @@ def colorscale2cv2colormap(colorscale):
     ])
     cv2cm_i = (255*(cs_val-cs_val[0])/(cs_val[-1]-cs_val[0])).astype(int)
     cv2cm_i = cv2cm_i + np.append([0], np.where(np.diff(cv2cm_i) == 0, 1, 0))
-    cv2cm_brga = np.array([
+    cv2cm_rgba = np.array([
         list(colorscale.keys())[i]
```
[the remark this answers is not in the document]
remicousin: colorscale2cv2colormap is going to the trash if the rest if good.

pingrid/impl.py
```diff
@@ -529,8 +552,8 @@ def to_dash_colorscale(s: str) -> List[str]:
     return cs
 
 
-def to_dash_colorscale2(s: str) -> List[str]:
-    cm = colorscale2cv2colormap(s)
+def to_dash_colorscale2(s):
+    cm = s.cv2cm()
     cs = []
     for x in cm:
         v = BGRA(*x)
```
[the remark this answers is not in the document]
remicousin: BGRA now irrelevant to all but fbfmaproom

pingrid/impl.py
```diff
@@ -529,8 +552,8 @@ def to_dash_colorscale(s: str) -> List[str]:
     return cs
 
 
-def to_dash_colorscale2(s: str) -> List[str]:
-    cm = colorscale2cv2colormap(s)
+def to_dash_colorscale2(s):
```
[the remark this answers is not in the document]
remicousin: There are now a few more ColorScale methods

pingrid/impl.py
remicousin: There are now a few ColorScale methods to get from one starting point to another ending point more cleanly and independently

remicousin:
The last push converts all enacts maprooms to the new framework. It will break fbfmaproom (because pingrid.tile now relies on ColorScale class). Once we are happy with the new framework, I can apply to fbfmaproom.

pingrid/impl.py
```diff
@@ -96,24 +87,40 @@ class ColorScale:
...
-        cv2cm_i = (255*(cs_val-cs_val[0])/(cs_val[-1]-cs_val[0])).astype(int)
-        cv2cm_i = cv2cm_i + np.append([0], np.where(np.diff(cv2cm_i) == 0, 1, 0))
-        cv2cm_rgba = np.array(self.colors)
-        cv2cm_bgra = cv2cm_rgba[:,[2, 1, 0, 3]]
-        cv2cm = np.full((256,4), np.nan)
-        for bgra in range(4):
-            cv2cm[:, bgra] = np.interp(np.arange(256), cv2cm_i, cv2cm_bgra[:, bgra])
-        return cv2cm.astype(int)
+               raise Exception(
+                   "if provided, scale must be same length as colors"
```
aaron-kaplan, line 97: Here is some ASCII art explaining why I think you need n-1 thresholds to go with n colors. If you have n thresholds, I don't know where to put the last one in this diagram.
```
       t1   t2   t3
<-- c1 | c2 | c3 | c4 -->
```

aaron-kaplan: Put in words, the thresholds represent points on the continuum, while the colors correspond to intervals. It takes n - 1 points to partition the continuum into n intervals.

remicousin: That's not how things are done right now, or intended to be done. Color ci is set at Ii that corresponds to ti and then interpolated for the rest of the array. If ti = tj, Ij = Ii +1

aaron-kaplan: I was (wrongly) thinking of bars of constant color rather than interpolation, but it seems like if you add interpolation to the mix, and you allow for color discontinuities at the boundaries, then you need even fewer thresholds for a given number of colors:

```
       t1         t2
<-- c1 | c2 -- c3 | c4 -->
```

As I mentioned yesterday, the colorscale code was originally written assuming a bounded range,  e.g. probabilities in [0-1], and then inappropriately used to display unbounded variables, e.g. precipitation amounts. Then I made a quick fix to make the precip maps not look like random noise when precip exceeded the expected maximum, but I was in "quick fix" mode at the time, and I may well have done it wrong or in a way that wasn't fully general. Let's try to get it right this time.

This kind of detail is why I thought it might be useful to do a survey of how each library treats colormaps, and see if we can pick a representation that can be mapped from ingrid and to any of plotly, opencv, and matplotlib.

aaron-kaplan: Oh, I see, the colormap I described above can be represented as follows in the current representation:

```
    t1   t1    t2   t2
<-- c1 | c2 -- c3 | c4 -->
```

So in terms of expressiveness it's equivalent, if a bit less efficient. But now I wonder if we're actually translating it correctly into opencv and plotly, including for these complicated cases like unbounded variables and discontinuities at boundaries. Do you know if we are?

pingrid/impl.py
```diff
@@ -96,24 +87,40 @@ class ColorScale:
...
+
+    def to_bgra(self):
+        return self.to_rgba()[:,[2, 1, 0, 3]]
+
+    def to_hex(self):
+        cm = self.to_rgba()
+        cs = []
+        for x in cm:
+            v = RGBA(*x)
+            cs.append(f"#{v.red:02x}{v.green:02x}{v.blue:02x}{v.alpha:02x}")
```
aaron-kaplan, line 121: We could have a `to_hex` method  on the `RGBA` class, and then call that method here.

The same might be true for the part of `to_rgba` that calculates the representation for a single color.

pingrid/impl.py
```diff
@@ -96,24 +87,40 @@ class ColorScale:
...
-    def cv2cm(self):
-        if self.scale is None:
-           cs_val = np.arange(len(colors))
+        if scale is None:
+            self.scale = list(np.arange(len(colors)))
         else:
-           if len(self.colors) == len(self.scale):
-               cs_val = np.array(self.scale)
+           if len(colors) == len(scale):
+               self.scale = scale
```
aaron-kaplan, line 94: Raise an exception if `scale` is not in sorted, non-decreasing order?

aaron-kaplan:
Whatever representation we settle on, I think it's important to have tests that confirm that it behaves the way we intend it to. Cases to test include: below the min, above the max, exactly at the min/max, interpolation between two points, exactly at one of the boundary points.

Given that we don't actually render colors ourselves, we only translate them to different representations for opencv and plotly to render, it seems like these tests are going to be somewhat hard to write.  Maybe I should do that myself, unless you feel like you have a handle on it.

pingrid/impl.py
remicousin: I don't know. I took inspiration from matplotlib representations that do allow this:

-- a list of anchors or nodes with colors defined on each side of the anchor/node, e.g.:
_autumn_data = {'red': ((0., 1.0, 1.0), (1.0, 1.0, 1.0)),
'green': ((0., 0., 0.), (1.0, 1.0, 1.0)),
'blue': ((0., 0., 0.), (1.0, 0., 0.))}

But also allows this:

-- but matplotlib doc indicated the following to create a colorscale through anchor points:
colors = ["darkorange", "gold", "lawngreen", "lightseagreen"]
nodes = [0.0, 0.4, 0.8, 1.0]
cmap2 = LinearSegmentedColormap.from_list("mycmap", list(zip(nodes, colors)))

and from the output of pingrid.parse_colormap.

I don't know if the threshold/interval is the way to think. If c1=c2, you don't want the color to take 2 consecutive spots in the array, don't you? On the other hand, since we are limited to 255 colors and integers from 0 to 255, and then scaling to whatever possible range of numbers, it would be more correct to rely on a threshold with colors on each side.

My understanding of things is superficial and approximative at this point an I only try to reproduce what I could see in pingrid, but starting from a different starting point. So I will have to look into more details into what plotly, opencv and matplotlib do and expect.

remicousin: and from what I can tell, pingrid.apply_colormap is rescaling data against so that scale_min becomes 0 and scale_max becomes 255 and the clips at 0 and 255 so there is no data point below/above the min/max.

aaron-kaplan: Yeah, I added the clipping. I don't know if that's the right way to deal with out-of-range values.

aaron-kaplan: > I don't know if the threshold/interval is the way to think. If c1=c2, you don't want the color to take 2 consecutive spots in the array, don't you? 

That's what I originally took issue with, but now I see that eliminating that redundancy would make it less expressive, so I'm ok with it.

What other way to think about it is there, other than threshold/interval?

remicousin: Ok. let's make a pause and rewind a bit.

We have data that goes from r0 to rn that are real numbers, ro <= rn
We want to put such data on a scale that goes from smin to smax, other real numbers, and smin <= smax. However, the relative position of r0 and rn with respect to smin and smax could be whatever (to generalize).

We are allowed 256 colors represented as an array of RGBA, or hex-stuff, or some other stuff I guess.

First color (0) should be assigned to data values < smin; last color (255) assigned to data values > smax. That leaves 254 colors from smin to smax, included.

So we are rescaling data from [smin, smax] to [1, 254].

Then we construct our array of 254 colors with some constraints and... I have to go take the shuttle...

remicousin: Then I guess that's where I am not sure what is supposed to happen. We want data value x assigned a specific color, thus assigned an integer between 1 and 254. Do we round the scaled x to the closest integer? Do we make small bins and assigned scaled x to one?

remicousin: Also we want to define colorscales inputting data values as thresholds (then be able to reuse an existing colorscale rescaling it for other data) so there is an issue of whether thresholds are inclusive or not, especially when making bands and 2 colors are assigned same value, which one should be of the color? And I suspect there might be 2 cases to consider: whether the data can take more or less than 256 values between smin and smax included. It's probably more subtle than that.

Then there is eventually the case of categorical data.

aaron-kaplan: I've spent several hours writing this comment, and come to understand things better in the process. The bottom line is that I think you're on the right track, it just took me some time to catch up with you.

Let's note that r0...rn may not be known at the moment that the colormap is defined. (I don't think that invalidates anything you said.)

> Do we round the scaled x to the closest integer? Do we make small bins and assigned scaled x to one?

Bins. Value x maps to colors[y] where y = floor((x-min) * 256 / (max-min)), assuming min <= x <= max. This is implemented in pingrid.apply_colormap. opencv requires its colormap to be in this 256-entry form.

I'm not sure we need to assume that n=256. That's a common value, and the only value allowed in opencv, but matplotlib seems to support other values. I like the anchor points representation because it doesn't constrain the number of values _a priori_; quantization can be delayed until we call into opencv, and if one day we replace opencv with something that supports higher color resolutions, we will be able to take advantage of it without changing most of our code.

Regarding out-of-range values, I surveyed the docs for the three packages.
- matplotlib uses n (256 by default) colors for in-range values, and an additional three for below min, above max, and masked. It doesn't reserve colors 0 and 255 for out-of-range values as you suggested.  By default it uses color 0 for below min and color 255 for above max (in addition to using those same colors for the first and last in-range intervals), but you can override that.
- Plotly maps below-min values to color 0 and above-max to color 255 with no option to override as far as I can see.
- Opencv doesn't seem to deal with out-of-range values at all. You're expected to normalize and clip your data before passing it to opencv for rendering, and if you don't, colors end up wrapping around, which is nonsensical. To avoid strange and library-dependent behavior, let's always clip data to min and max before passing it to any third-party library.

I see that plotly's colorscale representation is the same as the one you're using (n values, n colors), except it seems to assume that the values will all be between 0 and 1. I tried using values other than 0 and 1 for the first and last threshold, and plotly didn't complain, but I can't make sense of the result. Your approach of using arbitrary values and mapping them into [0, 1] or [0, 255] behind the scenes seems better.

> there is an issue of whether thresholds are inclusive or not, especially when making bands and 2 colors are assigned same value, which one should be of the color? 

The intervals should be considered open on one end and closed on the other, so that they don't overlap. Closed at the low end and open at the high end seems the most natural to me, but you could check what plotly and matplotlib do.

> And I suspect there might be 2 cases to consider: whether the data can take more or less than 256 values between smin and smax included. It's probably more subtle than that.

I don't understand what you're saying there. Could you elaborate or give an example?

Big picture:

Let's distinguish two things that we may have been conflating so far: the colormap representation our library uses internally, which can be translated to and from the representations used by other software; and the representation in which users of our library specify their colormaps. In the current implementation the two are the same, but they don't have to be. Plotly and matplotlib each offer multiple ways of specifying a colormap that are convenient in different cases, and we could do that too.

The anchor points representation you're using seems like a good internal representation. I objected to it at first because it seemed redundant, but now I understand that it gives greater expressiveness than the alternative I was suggesting, so I no longer object. (Specifically, it allows for interpolated color scales with discontinuities.)

Since the anchor points representation is redundant in simple cases, it would be nice to have the option to specify colormaps in a simpler format, as both matplotlib and plotly allow. But that can come later if we want.

I'm not sure this comment is going to help. Maybe we should talk.

remicousin: it's very helpful. I can see that the conversation can branch out on different sub-topics from here so maybe yes indeed it will be more efficient to talk. Maybe Tuesday January 10th? assuming we're both in the office and can use the blackboard? That is only 3 working-days from today, for me, after all!

aaron-kaplan:
There are some pre-existing tests of colormap functionality in fbfmaproom. They recently got broken, but I fixed them in https://github.com/iridl/python-maprooms/pull/268. They probably need to be adjusted on this branch.

pingrid/impl.py
remicousin: Shall we assume an in-person chat in the office next Tuesday (17th)?

remicousin:
These last set of commits intend to: get closer to matplotlib setup and address some of the comments above. They can be read in order, however later commits may overwrite earlier ones.

pingrid/impl.py
remicousin: I did something along those lines

pingrid/impl.py
```diff
@@ -96,24 +87,40 @@ class ColorScale:
...
+        cs_rgba = np.array(self.colors)
+        cs_rgba_full = np.full((256,4), np.nan)
+        for rgba in range(4):
+            cs_rgba_full[:, rgba] = np.interp(np.arange(256), cs_i, cs_rgba[:, rgba])
+        return cs_rgba_full.astype(int)
+
+    def to_bgra(self):
+        return self.to_rgba()[:,[2, 1, 0, 3]]
+
+    def to_hex(self):
```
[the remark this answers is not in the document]
remicousin: I did some renaming even though I am not too clear what originates from what ultimately.

pingrid/impl.py
```diff
@@ -90,6 +79,70 @@ def error_fig(error_msg="error"):
...
+    red: int
+    green: int
+    blue: int
+    alpha: int = 255
+
+    def to_hex(self):
+        return f"#{self.red:02x}{self.green:02x}{self.blue:02x}{self.alpha:02x}"
+
+
 class BGRA(NamedTuple):
```
aaron-kaplan, line 146: I don't think we should have both RGBA and BGRA classes. We could have one class called Color, with methods like `to_hex_rgba` and `to_hex_bgra`, if needed.

pingrid/impl.py
```diff
@@ -234,14 +286,14 @@ def _tile(da, tx, ty, tz, clipping):
...
-def empty_tile(width: int = 256, height: int = 256):
+def empty_tile(width: int = 256, height: int = 256, ncolors: int = 256):
     # If tile size were hard-coded, this could be a constant instead
     # of a function, but we're keeping open the option of changing
     # tile size. Also, numpy arrays are mutable, and having a mutable
     # global constant could lead to tricky bugs.
     im = apply_colormap(
         np.full([height, width], np.nan),
-        np.zeros((256, 4)),
+        np.zeros((ncolors, 4)),
```
aaron-kaplan, line 296: I don't think this is correct. `apply_colormap` is specific to opencv, so it requires a 256-element colormap.

We should eventually modify apply_colormap so that it takes a ColorScale instead of an opencv colormap; but when we do that, this function still won't need an `ncolors` parameter. The colormap resolution makes no difference here; all we need is a colormap that maps all values to (0, 0, 0, 0). We can just pick an arbitrary ncolors, like 2, and hard-code it.

aaron-kaplan (commented):
Tests would help with both correctness and readability. Other comments below.

remicousin:
I am not too motivated to write tests and docstrings until I feel I am on tracks, which I don't.

pingrid/impl.py
```diff
@@ -90,6 +79,70 @@ def error_fig(error_msg="error"):
...
+        else:
+           if (np.diff(scale) < 0).any():
+              raise Exception("scale must be monotically increasing")
+           elif len(colors) == len(scale):
+               self.scale = scale
+           else:
+               raise Exception("scale must be same length as colors")
+
+    def get_name(self):
+        return self.name
```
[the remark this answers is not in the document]
remicousin: I think I got influenced here by something I read in matplotlib

pingrid/impl.py
```diff
@@ -90,6 +79,70 @@ def error_fig(error_msg="error"):
 FuncInterp2d = Callable[[Iterable[np.ndarray]], np.ndarray]
 
 
+class ColorScale:
+    
+    def __init__(self, name, colors, scale=None, ncolors=256):
```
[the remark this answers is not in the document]
remicousin: I think I tried to get closer to [matplotlib](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.LinearSegmentedColormap.html#matplotlib.colors.LinearSegmentedColormap) but I am just getting confused often on how to do this.

So I guess no ncolors in ColorScale, all that matters is that to_rgba_array with lutsize with output an array that is what matplotlib typically utilizes.

pingrid/impl.py
```diff
@@ -90,6 +79,70 @@ def error_fig(error_msg="error"):
...
+    def rescaled(self, new_min, new_max):
+        cs_val = np.array(self.scale)
+        scale = (cs_val - cs_val[0]) * (new_max - new_min) / (cs_val[-1] - cs_val[0]) + new_min
+        return ColorScale(self.name, self.colors, scale, self.ncolors)
+
+    def to_rgba_array(self, lutsize=None):
+        if lutsize is None:
+            lutsize = self.ncolors
+        cs_i = self.rescaled(0, lutsize-1).scale.astype(int)
+        cs_i = cs_i + np.append([0], np.where(np.diff(cs_i) == 0, 1, 0))
```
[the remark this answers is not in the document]
remicousin: This is for the bands. Bands mean discontinuity so the anchor point is associated with 2 colors. One is assigned to the index rescaled to 0-lutsize, the other is assigned to the next index.
Then when all the colors are assigned to an index rescaled on 0-lutsize, np.interp fills the blanks, if I may say so, creating blending colors between anchor points with different colors, and a band of identical colors between anchor points of same colors.

pingrid/impl.py
```diff
@@ -90,6 +79,70 @@ def error_fig(error_msg="error"):
...
+
+    def rescaled(self, new_min, new_max):
+        cs_val = np.array(self.scale)
+        scale = (cs_val - cs_val[0]) * (new_max - new_min) / (cs_val[-1] - cs_val[0]) + new_min
+        return ColorScale(self.name, self.colors, scale, self.ncolors)
+
+    def to_rgba_array(self, lutsize=None):
+        if lutsize is None:
+            lutsize = self.ncolors
+        cs_i = self.rescaled(0, lutsize-1).scale.astype(int)
```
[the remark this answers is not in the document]
remicousin: Yes, will get back to it once we resolve the below.

pingrid/impl.py
```diff
@@ -90,6 +79,70 @@ def error_fig(error_msg="error"):
...
+        return self.to_rgba_array(lutsize=lutsize)[:,[2, 1, 0, 3]]
+
+    def to_dash_leaflet(self, lutsize=None):
+        if lutsize is None:
+            lutsize = self.ncolors
+        cm = self.to_rgba_array(lutsize=lutsize)
+        cs = []
+        for x in cm:
+            v = RGBA(*x)
+            cs.append(v.to_hex())
```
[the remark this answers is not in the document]
remicousin: cs and cm come from pingrid (Igor's I guess?)

pingrid/impl.py
remicousin: Right now there is still RGBA in pingrid and directly in fbfmaproom. I'll clean that up when we are happy with what we are doing and it all works with the enactsmulti maprooms. 

So what would the Color class look like?

pingrid/impl.py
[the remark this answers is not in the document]
aaron-kaplan: Indeed, matplotlib does it the way you did. A LinearColorScale is quantized and represented as a LUT, and then potenially resampled at a later time if a LUT of a different size is needed. It seems like that results in an unnecessary loss of information in some cases.

The authors of matplotlib clearly know more about this subject than I do, so maybe I'm missing something, but for now my recommendation is still to postpone sampling until a LUT is actually needed.

pingrid/impl.py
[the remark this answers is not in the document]
aaron-kaplan: That doesn't seem right. I'll come talk to you about it after looking over the other comments.

pingrid/impl.py
aaron-kaplan: Just like RGBA, except with `to_hex` renamed `to_hex_rgba`, and a new method `to_hex_bgra` added. When creating a Color from BGRA data, we'll have to be careful either to invert the order of the color arguments, or use kwarg syntax e.g. `Color(blue=b, green=g, red=r, alpha=a)`.

remicousin:
Last commit intents the piecewise implementation of colorscales. In monthly/maproom, I dropped the case drawing a categorical color bar because the map is using a lutsize=256 no matter what for now, so the color bar should be drawn the same way in case lutsize causes the loss of information. The categorical colorbar is nicer to the user's eye but could result being incorrect with respect to the map.

If it looks like we're on tracks, I'll clean up, add doc and tests, and then move on to following items.

pingrid/impl.py
```diff
@@ -90,6 +79,89 @@ def error_fig(error_msg="error"):
...
+        return ColorScale(name, self.colors[::-1], self.scale)
+
+    def rescaled(self, new_min, new_max):
+        cs_val = np.array(self.scale)
+        scale = (cs_val - cs_val[0]) * (new_max - new_min) / (cs_val[-1] - cs_val[0]) + new_min
+        return ColorScale(self.name, self.colors, scale)
+
+    def to_rgba_array(self, lutsize=256):
+        cs = self.rescaled(0, lutsize-1)
+        nc =  len(cs.scale)
```
aaron-kaplan, line 109: It would be clearer to call this variable `n_anchors`. The number of colors prior to quantization is potentially infinite, and the number of colors after quantization is lutsize. This variable represents neither of those.

pingrid/impl.py
```diff
@@ -90,6 +79,89 @@ def error_fig(error_msg="error"):
...
+        colors = np.array(cs.colors)
```
aaron-kaplan, line 110: Should we just change the ColorScale constructor to make cs.colors an np.array, rather than converting it to an array every time we use it?

pingrid/impl.py
```diff
@@ -90,6 +79,89 @@ def error_fig(error_msg="error"):
...
+                          delta_colors[i, band] / delta_scale[i],
+                      ]
+             ) for i in range(nc)]
+            # This is the lambda version of the Polynomial
+            # just can't get it to work... Leaving it here for someone smarter than me
+            ##[lambda x: (colors[i, band]
+            ##    + 0 if np.diff(cs.scale, append=cs.scale[-1])[i] == 0 else ((x - cs.scale[i]) * delta_colors[i, band] / delta_scale[i])
+            ##) for i in range(nc)]
+        # Same rescaling to all color bands
+        ) for band in range(4)])).astype(int)
```
aaron-kaplan, line 136: Do we really need to iterate over the bands explicitly? Is there a way to write it as a single calculation over a 2D array (variable value and band) instead of four 1D arrays?

If not, it's fine. Just looking to be more concise. Explicit loops in numpy code can be a problem for performance, but in this particular case it's not going to matter because the arrays are small.

pingrid/impl.py
```diff
@@ -90,6 +79,89 @@ def error_fig(error_msg="error"):
...
+        # Construct lutsize x 4 RGBA array
+        rgbaa = np.transpose(np.array([np.piecewise(
+            # Rescale lut indices to colors piece by piece
+            np.arange(lutsize),
+            # Rescale differently from one anchor point to the next
+            [np.arange(lutsize) >= cs.scale[i] for i in range(nc)],
+            # Rescaling is linear from one anchor to the next
+            [np.polynomial.polynomial.Polynomial(
+                  # Unless it's a discontinuity then there is no rescaling
+                  [colors[i, band], 0] if np.diff(cs.scale, append=cs.scale[-1])[i] == 0
```
aaron-kaplan, line 123: use delta_scale instead of repeating its definition.

pingrid/impl.py
```diff
@@ -90,6 +79,89 @@ def error_fig(error_msg="error"):
...
+        colors = np.array(cs.colors)
+        # append output is not used but saves writing a condition dedicated to last color
+        delta_colors  = np.diff(colors, axis=0, append=np.expand_dims(colors[-1,:], 0))
+        delta_scale = np.diff(cs.scale, append=cs.scale[-1])
+        # Construct lutsize x 4 RGBA array
+        rgbaa = np.transpose(np.array([np.piecewise(
+            # Rescale lut indices to colors piece by piece
+            np.arange(lutsize),
+            # Rescale differently from one anchor point to the next
+            [np.arange(lutsize) >= cs.scale[i] for i in range(nc)],
```
aaron-kaplan, line 119: Don't we need to compare to both lower and upper bounds to avoid having multiple true conditions for a single interval?

aaron-kaplan (commented):
Sending the comments I have so far because I have to stop for the afternoon. They're all local, line-by-line suggestions; my understanding of the gestalt isn't complete yet. I have a feeling it could be made simpler, but I'll have to work on it some more to see if that's really true.

np.polynomial has a constant called polyone that seems like it would be shorthand for a polynomial of degree one, but the docs don't make any sense (seem to be a copy of the docs for np.ndarray). I want to look into that.

aaron-kaplan:
I think it's time to add tests. Preferably tests that don't depend on internal implementation details--just put a list of anchor points in, and get a 256-element LUT out, so that the test remains valid if we change the implementation.

aaron-kaplan:
BTW good point about generating the colorbar and the map by the same method so they don't get out of sync, but I don't understand why that would make the colorbar less pleasing visually. Please give an example if you have one.

remicousin:
Where do I write pingrid tests?

aaron-kaplan:
> Where do I write pingrid tests?

There are some in fbfmaproom. You could put them there for now. I'll eventually move them somewhere else.

remicousin:
I am writing the documentation. Do you think the to_ methods of ColorScale and Color classes need be public? Right now the only one that is used explicitely in Maprooms files is to_dash_leaflet() for the colorbar. But ultimately it seems all those methods would be triggered by other public methods that make use of colorscales, right?
not sure...

aaron-kaplan:
At least the ColorScale methods need to be public, because we're going to want to adapt the tile generation code to use ColorScale. Not sure about Color, whether that should be considered an implementation detail of ColorScale or a public API in its own right.

pingrid/impl.py
remicousin: so its colors that has bands and colors is used to define the Polynomial functions. But Polynomials takes only 1-d array so I'd need to either make a loop at this level, or to decompose x and reform it. Either way, I would need to use the lambda x notation and as soon as I use that even without making any changes, it's not working anymore... (it's going through the last function).

pingrid/impl.py
remicousin: If we don't want multiple Trues, yes. Right now I am relying on the fact that the last True is the one that counts. Shall I make sure we have unique Trues?

remicousin:
> BTW good point about generating the colorbar and the map by the same method so they don't get out of sync, but I don't understand why that would make the colorbar less pleasing visually. Please give an example if you have one.

So now the categorical rainfall color bar looks like this:

Screen Shot 2023-02-08 at 12 40 58 PM

I shows this little band of white for the colors near 0 but it's hard to tell what data range it maps. Whereas the dash leaflet categorical color bar would have looked like this (search "categorical_colorbar"):

https://dash-leaflet-docs.onrender.com/

where equal size are given to categories regardless of the data ranges it represents.

remicousin:
I pushed documentation and tests. Might not have done it the right way for both. For Docs, not sure how to make html for pingrid, and for tests, I added stuff to be importable but not sure that's what should have been done.

remicousin:
Also, parse_color* related tests will fail because I removed BGRA but I didn't adapt those functions (now unused) accordingly. And fbfmaproom has been changed to use the new framework but you may want to double check that part.

pingrid/impl.py
aaron-kaplan: I think so, both for efficiency and to avoid relying on undocumented behavior.

aaron-kaplan:
I still don't get your point about the categorical color bar. The dash-leaflet example defines a scale with ten colors, each covering 1/10th of the range of the variable, whereas the rainfall example defines a scale where the first color (WHITE) covers from 0 to 0.2, and the second color coverse from 0.2 to 2. So naturally in the first case the bars are all the same width whereas in the second the second one is 9 times wider than the first.

If we want tick marks to show the boundary values, we can add them.

BTW the rainfall colormap demonstrates that this anchor point representation isn't a very convenient way to express a categorical colormap. It works, so it's good enough for now, but I think we should add alternative input representations in the future.

remicousin:
The Dash Leaflet example (last map of the page) as the first color band for 0-50 and the one to last for 500-1000 but are the same width regardless.

aaron-kaplan:
> The Dash Leaflet example (last map of the page) as the first color band for 0-50 and the one to last for 500-1000 but are   the same width regardless.

Oh yeah. First one is 0-10, not 0-50, but point taken. That is a purely aesthetic choice. We can do it that way if we wanted; that would be a strong argument for having separate classes for categorical and continuous colorscales.

pingrid/impl.py
```diff
@@ -402,14 +610,174 @@ def produce_shape_tile(
...
+        Color(128, 0, 0),
+        Color(128, 0, 0),
+    ],
+    [0, 1, 1, 1.2, 1.2, 1.4, 1.4, 1.6, 1.6, 1.7, 1.7, 1.8, 1.8,
+     2, 2, 2.2, 2.2, 2.4, 2.4, 2.6, 2.6, 2.8, 2.8,
+     3, 3, 3.2, 3.2, 3.4, 3.4, 3.6, 3.6, 3.8, 3.8, 5],
+)
+
+_FBF_PNE_CS = ColorScale(
+    "fbf_pnep",
```
aaron-kaplan, line 723: We shouldn't have an item called `fbf_pnep` here. Either it's reusable in other applications, in which case it should be called something more generic like `pnep`,  or it's specific to fbf, in which case it should be in the fbf application, not in pingrid.

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -88,11 +88,7 @@ countries:
                         lat: Y
                         lon: X
                         time: T
-                    colormap: &raincm "[16777215 16777215 12632256 12632256 14155730
-                        [14155730 21] 12841150 [12841150 23] 10215830 [10215830 23]
-                        7262835 [7262835 22] 6931300 [6931300 23] 5944410 [5944410 23]
-                        5285200 [5285200 23] 4625990 [4625990 22] 3966780 [3966780 23]
-                        3307570 [3307570 23] 2648360 [2648360 23] 1989150]"
+                    colormap: &raincm precip
```
aaron-kaplan, line 91: We don't need the `&raincm` anchor anymore. That was to save us copy-pasting the whole colormap representation every time (each occurrence of `*raincm` below refers to this same value), but now that we're using a name rather than a colormap, we can just use the name everywhere.

fbfmaproom/fbfmaproom.py
```diff
@@ -33,8 +33,8 @@
 
 import __about__ as about
 import pingrid
-from pingrid import ClientSideError, InvalidRequestError, NotFoundError, parse_arg
-from pingrid.impl import BGRA  # TODO stop using private BGRA class
+from pingrid import ClientSideError, CMAPS, InvalidRequestError, NotFoundError, parse_arg
+from pingrid.impl import Color # TODO stop using private Color class
```
aaron-kaplan, line 37: Remove the comment and the `.impl`. You made Color public.

fbfmaproom/fbfmaproom.py
```diff
@@ -220,7 +220,7 @@ def open_data_array(
             val_max = cfg["range"][1]
         else:
             assert False, "configuration doesn't specify range"
-    da.attrs["colormap"] = cfg["colormap"]
+    da.attrs["colormap"] = CMAPS[cfg["colormap"]]
```
aaron-kaplan, line 223: This is fine, but if you decide that the PNE colorscale should be FBF-specific, you can do something like
```
import pingrid
CMAPS = dict(pingrid.CMAPS, {
    'pne': pingrid.ColorScale(
        ...
    )
}
```

fbfmaproom/fbfmaproom.py
```diff
@@ -1169,9 +1169,10 @@ def vuln_tiles(tz, tx, ty, country_key, mode, year):
             (
                 r["the_geom"],
                 pingrid.impl.DrawAttrs(
-                    BGRA(0, 0, 255, 255),
+                    Color(255, 0, 0, 255),
                     pingrid.impl.with_alpha(
-                        pingrid.parse_colormap(cfg["colormap"])[
+                        CMAPS[cfg["colormap"]].to_rgba_array()[
+                        #pingrid.parse_colormap(cfg["colormap"])[
```
aaron-kaplan, line 1175: remove comment

fbfmaproom/fbfmaproom-sample.yaml
aaron-kaplan: i.e. both `&raincm precip` and `*raincm` get replaced with `precip` everywhere. Likewise for the other colormaps.

pingrid/impl.py
```diff
@@ -90,16 +81,234 @@ def error_fig(error_msg="error"):
...
-class BGRA(NamedTuple):
-    blue: int
-    green: int
+class ColorScale:
+    """A class to define and manipualte colorscales.
+
+    `ColorScale` instances are defined by colors at anchor points and are named.
+    Colors are graduated from one anchor to the next.
+    To make a band of the same color, following anchors should be equal.
+    To mark a discontinuity, following colors should differ and have the same anchor value.
```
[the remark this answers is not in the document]
remicousin: I would be nice then to print an image of the Examples colorscale. More generally speaking, it would be nice to print an image of all our registered colorscales. Can we do that?

aaron-kaplan:
@remicousin what's up with this PR?

aaron-kaplan:
> I don't know how to make html for the pingrid doc

Currently, only the wat_bal directory has documentation set up, for calc.py and agronomy.py. Does it make sense to move those two files into the pingrid module? If so, we'll generate the documentation there and it will cover all three; if not, I'll setup a separate Makefile for the pingrid documentation.

remicousin:
Yes. How do we go about this? All I really have to do here before merging is comment out the parse_color tests. Everything else we discussed i will file as issues. Shall I take care of those tests, then merge, then we can deal with the above in a separate PR?

aaron-kaplan:
Sounds good.

remicousin:
@aaron-kaplan  I finished rebasing on master. Would you like to test out a bit fbfmaproom? I could run it but I don't really know what to do to be thorough about it. Or let me know if/when good to merge into master.

aaron-kaplan:
It's closer, but I still see some differences in shades of orange in

http://shortfin01.iri.columbia.edu:9999/fbfmaproom/ethiopia?mode=0&season=season1&predictors=pnep+rain&predictand=bad-years&year=2023&issue_month=feb&freq=30&severity=0&include_upcoming=false&position=%5B6.875%2C43.875%5D&show_modal=false

vs

https://iridl.ldeo.columbia.edu/fbfmaproom2/ethiopia?mode=0&season=season1&predictors=pnep+rain&predictand=bad-years&year=2023&issue_month=feb&freq=30&severity=0&include_upcoming=false&position=%5B6.875%2C43.875%5D&show_modal=false

Is that what you meant by "the progressive change to DarkRed?" There's no color called DarkRed in `_RAIN_POE_CS`.

BTW in the code it looks like you're defining a continuous colorscale, but both the colorbar and the map seem to have no more than 16 distinct colors. Do you know why that is?

remicousin:
I can't see differences... But that doesn't mean they are the same. I think at some point, probably when they were only red colors, I made up the graduation and didn't try to reproduce the exact colors listed in fbf maproom config file.

with respect to the continuous vs 16 colors comment, I think the answer is that since opencv can only make 256 color arrays no matter what, then the map has to be so. Then we said at some point that it was wiser to make the colorbar the same way we make the map, to avoid potential discrepancies.

aaron-kaplan:
> with respect to the continuous vs 16 colors comment, I think the answer is that since opencv can only make 256 color arrays no matter what, then the map has to be so. Then we said at some point that it was wiser to make the colorbar the same way we make the map, to avoid potential discrepancies.

Right, I expected to see 256 different colors, not 16. (And BTW the old and new versions are the same in this respect.)

I compare the two maps by opening the two versions in different tabs in the same window, and flipping back and forth between the two tabs. Since the content is almost identical, the places where it's not identical pop out. It's not a big deal, but I'm curious why they're not identical. Is there wiggle room in the definition of linear interpolation?

aaron-kaplan (approved):

remicousin:
I am confused because I can't see a difference between the 2 maps.

Screenshot 2023-04-04 at 2 44 06 PM

And I don't understand your last comment on 256 vs 16 colors. We always have 256 bins because of opencv constrain, but colors can be the same in some bins.

aaron-kaplan:
I do see differences between your two snapshots. E.g. the dark orange block in the southwest tip has an incursion of light orange on its east edge in the image on the right, but not on the left. Anyway, the differences are easier to spot if you superimpose them as I described in my previous comment.

> And I don't understand your last comment on 256 vs 16 colors. We always have 256 bins because of opencv constrain, but colors can be the same in some bins.

They can be, but only if two successive anchor points have the same value. The POE colorscale doesn't have repeated anchor points except for Moccasin. Outside the Moccasin band, I expected to see more variation in color. No?

remicousin:
I saw the spot with a difference and I saw another one soon after. So I think it's likely because: i) I made up the anchors "by the eye" as they are not shown in the way they are defined the old way and ii) some colors grading in the reds I might also have done by manual linear interpolation form the lightest test to the darkest one. Then on top of that I wonder if there could be some inclusion dealt differently for the edges of the colors bins.

All that sounds acceptable, right?

Now I understand the other comment. You are looking at rain_poe that is going to be used by the flex. forecast maproom, and be transformed to adapt to choices of thresholds. The fbfmaproom colorscale is pne_25.

Let me know if I am good to rebase and merge

aaron-kaplan:
Good to rebase and merge.

I understand what you said about adapting the colorscale to the choice of threshold, but I don't see what that has to do with the number of distinct colors that appear on the map and the colorbar.  To be continued in another forum.

remicousin:
They are not the same colorscales. fbfmaproom is not using what you are looking at. fbfmaproom using some colorscale with 16 bands

PR #247, opened 2022-12-05
Drewr/unify widgets

drewmresnick:


enacts/onset/layout.py
```diff
@@ -5,7 +5,7 @@
 from dash import dash_table
 import dash_leaflet as dlf
 import plotly.express as px
-from .widgets import Block, Sentence, Date, Units, Number
+from .components import Block, Sentence, Date, Units, Number
```
remicousin, line 8: uses DateNoYear so import that instead of Date that is not used.
Also Units is not used so could be removed from the imports

remicousin (changes requested):

remicousin:
Block is missing the new optional ison argument that was set up for onset

https://github.com/iridl/python-maprooms/blob/master/enacts/onset/widgets.py

remicousin:
FYI I am taking over as Drew is going to be offline for the rest of the year, mostly.

remicousin:
@aaron-kaplan nad @kgraaf , please review.

remicousin:
Testable and tested on enacts-multipage for zmd config

components.py
```diff
@@ -1,11 +1,9 @@
 import dash
```
aaron-kaplan, line 1: I suggest we call it controls instead of components. ui_components was also better than components. Components could be anything at all.

components.py
```diff
@@ -79,6 +77,8 @@ def DateNoYear(id, defaultDay, defaultMonth):
         Default value that is displayed in the input box when user loads the page.
     defaultMonth : str
         Default value that is displayed in the dropdown when user loads the page.
+        Valid values are first 3 letters of the month with inital in upper caps.
+
     Returns
     -------
     [dbc.Input, dbc.Select] : list
```
aaron-kaplan, line 84: Can we reuse `Month` below instead of copy-pasting its definition?

components.py
```diff
@@ -15,23 +13,21 @@ def Number(id, default, min, max):
...
         dbc Input component with numerical inputs.
-    Notes
-    -----
-    Examples
-    --------
     """
-    return dbc.Input(id=id, type="number", min=min, max=max, size="sm",
-                     className="m-1 d-inline-block w-auto",debounce=True,value=str(default))
+    return [dbc.Input(id=id, type="number", min=min, max=max, size="sm",
+                     className="m-1 d-inline-block w-auto", debounce=True, value=str(default))]
```
aaron-kaplan, line 29: I feel like assigning these presentation-related CSS classes here is mixing content and presentation, but I really have no idea how bootstrap is meant to be used. @kgraaf any thoughts?

components.py
```diff
@@ -156,36 +158,46 @@ def Block(title, *body, width="100%"): #width of the block in its container
...
     """
+    if ison:
+        the_display = "inline-block"
+    else:
+        the_display = "none"
     return dbc.Card([
         dbc.CardHeader(title),
         dbc.CardBody(body),
-    ], className="mb-4 ml-4 mr-4", style={"display": "inline-block", "width": width})
+    ], className="mb-4 ml-4 mr-4", style={"display": the_display, "width": width})
```
aaron-kaplan, line 179: @kgraaf you've mentioned elsewhere that it's better to use CSS rather than in-line styles. Should we control visibility using a class instead of directly manipulating style here?

aaron-kaplan (commented):
I've had a look, but I would like @kgraaf to comment too.

components.py
```diff
@@ -1,11 +1,9 @@
 import dash
-from dash import dcc
 from dash import html
 import dash_bootstrap_components as dbc
-import dash_leaflet as dlf
 
 
-def Number(id, default, min, max):
+def Number(id, default, min=0, max=5):
```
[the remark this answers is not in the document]
remicousin: None were using those default.

remicousin:
@kgraaf please review as we need this one out of the way sooner than later. thanks!

controls.py
```diff
@@ -144,7 +135,7 @@ def Sentence(*elems):
 
     return dbc.Form(groups)
 
-def Block(title, *body, width="100%"): #width of the block in its container
+def Block(title, *body, ison=True, width="100%"): #width of the block in its container
```
kgraaf, line 138: What does `ison` mean here, and what is the use case of hiding the card in this way vs changing the style?

kgraaf, line 138: What is `ison` short for here, and what is the use case of hiding the card in this way vs changing the style?

controls.py
```diff
@@ -1,11 +1,9 @@
 import dash
-from dash import dcc
 from dash import html
 import dash_bootstrap_components as dbc
-import dash_leaflet as dlf
 
 
-def Number(id, default, min, max):
+def Number(id, default, min=None, max=None):
```
kgraaf, line 6: I'll take the blame for this, since I wrote this file original, but something I do wonder is if the order of the parameters and names of these components ought not to match the built-in dash components. My trainees found that very confusing that you set the id differently here.

enacts/onset/controls.py
```diff
@@ -0,0 +1 @@
+../../controls.py
```
kgraaf, line 1: When I set up the multi-page configuration initially, I was just trying to get something out as quickly as possible given the time constraints, but I'm not really sure if I did things the absolute best way. I wonder if there is any way we can put controls.py somewhere such that all the individual sub-apps can access it without having to symlink it.

controls.py
```diff
@@ -15,23 +13,21 @@ def Number(id, default, min, max):
...
     dbc.Input : component
         dbc Input component with numerical inputs.
-    Notes
-    -----
-    Examples
-    --------
     """
-    return dbc.Input(id=id, type="number", min=min, max=max, size="sm",
-                     className="m-1 d-inline-block w-auto",debounce=True,value=str(default))
+    return [dbc.Input(id=id, type="number", min=min, max=max, size="sm",
```
kgraaf, line 28: Do we think that always returning a list of components is the way to go? (This is an actual, no rhetorical question: I've gone back and forth)

kgraaf (commented):
I've made some comments on your code. But generally, I studied this very carefully and it looks really good

enacts/onset/controls.py
aaron-kaplan: This will be taken care of by packaging pingrid (working name) as a library.

controls.py
remicousin: ison stands for "is on." The [Onset Maproom](https://github.com/iridl/python-maprooms/blob/bd77571a6c27b15af4c13872f27f8f2ed26efa3a/enacts/onset/layout.py#L333) has an option to show cessation date analysis in addition to onset date analysis, but it's not available for all countries. So in countries where not available, the Block that provides the controls to define the parameters for cessation date is not displayed.

I think @aaron-kaplan wasn't too happy about that and would have preferred 2 Maprooms.

I didn't understand the last part of the question. I assume you mean hard-coding the style? well we can't in that set up since we use same layout.py and use config files to display it or not.

controls.py
remicousin: I am not sure I understand... Does dbc.Input cares about the order?

controls.py
remicousin: Number doesn't work without the []

aaron-kaplan:
@kgraaf it looks like the ball is back in your court on this. Please stay with it until it's finished--other branches that depend on this one are piling up.

controls.py
kgraaf: What I'm saying is that the order of the parameters is inconsistent with how the built-in components work, which is a tad confusing in my opinion.

controls.py
kgraaf: Well right, because the places that use it are expecting lists. I'm just wondering more broadly whether this is the right way to do things. It's probably fine for the sake of this PR though

controls.py
kgraaf: Ok makes sense. Any way we could come up with a better name for this parameter? (even `is_on` would be fine) I don't want to be too picky I think it just might be confusing as is.

components.py
kgraaf: Yeah it probably is, but at the same time we've already entangled presentation and content too much as it is in this manner so I don't know how much this matters.

kgraaf:
I believe I've gotten every outstanding question answered

controls.py
remicousin: I don't think I get it. I can see that you might want to use "value" for "default". But I don't think we want optional id and default so we would not be able to put value/default last. But I'm probably off topic.

aaron-kaplan: The `dbc.Input` constructor doesn't have any positional arguments, only keyword arguments. So (a) they're all optional, and (b) they can be provided in any order. In contrast, here we've made `id` and `default` positional arguments, which means they must come before everything else. I guess @kgraaf is asking whether we should make `id` and `default` optional?

Might be a good idea, but not worth holding up the PR.

aaron-kaplan (approved):

PR #249, opened 2022-12-08
Madagascar djf new datasets

kgraaf:


fbfmaproom/fbf-update-data.py
```diff
@@ -33,7 +33,27 @@
     ),
     (
         "madagascar/enacts-precip-djf",
-        "http://iridl.ldeo.columbia.edu/home/.rijaf/.Madagascar_v3/.ALL/.monthly/.rainfall/.rfe/T/(Dec-Feb)/seasonalAverage//units/(mm/month)/def/",
+        "http://map.meteomadagascar.mg/SOURCES/.Madagascar_v4/.ALL/.monthly/.rainfall/.rfe/T/(Dec-Feb)/seasonalAverage//units/(mm/month)/def/X/42.525/48.975/RANGE/Y/-25.9875/-20.025/RANGE/",
```
aaron-kaplan, line 36: Was this slow to download? It (and the DJF version) would be faster if we did the spatial subsetting before the seasonalAverage. But if it's already fast, no point going back and testing again.

aaron-kaplan (commented):
Looks good so far. Please add to data-conversion-scripts the script you used to generate the bad years zarr, or perhaps it would be simpler to modify the script that's already there to make it generate both the djf and ond zarrs. Then set up an internal test server so everyone can take a look.

fbfmaproom/fbf-update-data.py
aaron-kaplan: ditto for the following one.

kgraaf: The only one that was slow was NDVI which is just inherently slow so I'm leaving it as is

aaron-kaplan (changes requested):
I made comments on https://github.com/iridl/python-maprooms/commit/8013a0b3e5ddb90efbf2d20dd888b3b883339a5b (accidentally, meant to put them on the PR).

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -195,16 +206,104 @@ countries:
...
+                        obs: spi
+                        lat: Y
+                        lon: X
+                        time: T
+                    colormap: *raincm
+                    lower_is_worse: yes
+                    format: number3
+                subseas-dry-tercile-snov-l22-djf:
+                    label: November biweekly forecast, 3wk lead
+                    description: Biweekly rainfall forecast issued in early November for the two week period starting a week after the issue date. Probability of the driest tercile.
```
aaron-kaplan, line 256: "starting two weeks after the issue date". You copied the description from Niger which uses L=15 whereas here we're using L=22.

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -195,16 +206,104 @@ countries:
...
+                    path: madagascar/subseas-dry-tercile-snov-l22-djf.zarr
+                    var_names:
+                        obs: prob
+                        lat: Y
+                        lon: X
+                        time: T
+                    colormap: *raincm
+                    lower_is_worse: no
+                s2s-actual-rainfall:
+                    label: November actual rainfall
```
aaron-kaplan, line 266: Observed subseasonal rainfall

fbfmaproom/fbf-update-data.py
```diff
@@ -33,12 +33,44 @@
...
....rain/T/14/runningAverage/c%3A/14.0//units//days/def/%3Ac/mul/T/(1200%2021%20Nov%201999%20-%201200%205%20Dec%201999)/(1200%2019%20Nov%202000%20-%201200%203%20Dec%202000)/(1200%2018%20Nov%202001%20-%201200%202%20Dec%202001)/(1200%2017%20Nov%202002%20-%201200%201%20Dec%202002)/(1200%2016%20Nov%202003%20-%201200%2030%20Nov%202003)/(1200%2021%20Nov%202004%20-%201200%205%20Dec%202004)/(1200%2020%20Nov%202005%20-%201200%204%20Dec%202005)/(1200%2019%20Nov%202006%20-%201200%203%20Dec%202006)/(1200%2018%20Nov%202007%20-%201200%202%20Dec%202007)/(1200%2016%20Nov%202008%20-%201200%2030%20Nov%202008)/(1200%2022%20Nov%202009%20-%201200%206%20Dec%202009)/(1200%2021%20Nov%202010%20-%201200%205%20Dec%202010)/(1200%2020%20Nov%202011%20-%201200%204%20Dec%202011)/(1200%2018%20Nov%202012%20-%201200%202%20Dec%202012)/(1200%2017%20Nov%202013%20-%201200%201%20Dec%202013)/(1200%2016%20Nov%202014%20-%201200%2030%20Nov%202014)/(1200%2022%20Nov%202015%20-%201200%206%20Dec%202015)/(1200%2020%20Nov%202016%20-%201200%204%20Dec%202016)/VALUES/T//pointwidth/0/def/pop/T/(months%20since%201960-01-01)/streamgridunitconvert/T/toi4/use_as_grid/T/2.5/shiftGRID/X/42.525/48.975/RANGEEDGES/Y/-25.9875/-20.025/RANGEEDGES/",
```
aaron-kaplan, line 72: You've only covered the hindcast period. We need to cover the hindcasts and the forecasts, i.e. go all the way to 2022

fbfmaproom/fbfmaproom-sample.yaml
kgraaf: should that be the label or the name of the data set or both

fbfmaproom/fbf-update-data.py
kgraaf: Oh I see. The cross over point between the two is 2016?

fbfmaproom/fbfmaproom-sample.yaml
aaron-kaplan: I don't care as much about the name of the dataset, but since you have to download it again anyway (see my comment about extending the temporal range), might as well make it consistent.

fbfmaproom/fbf-update-data.py
aaron-kaplan: Yeah, unfortunately there's a gap between the hindcasts and the forecasts, which we can't do anything about. No reason there should be a gap in the obs, because those are available.

aaron-kaplan (changes requested):
Just looked back at this morning's email... Dan wants to start the history at 1991 instead of 1981. That's a one-line change in the config file.

Also need to bring in the new bad years data, as you mentioned.

- [ ] start history at  1991
- [ ] new bad years data

kgraaf:
Ok. Give me about 20-30 minutes to finish this up

kgraaf:
> Just looked back at this morning's email... Dan wants to start the history at 1991 instead of 1981. That's a one-line change in the config file.
> 
> Also need to bring in the new bad years data, as you mentioned.
> 
>     * [x]  start history at  1991
> 
>     * [x]  new bad years data

As far as I know this should be done. The new bad years are set as the default and live on the test server. They are located in `/data/kgraaf/madagascar/madagascar/bad-years-new.zarr/` for when you're ready to deploy

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -195,16 +206,115 @@ countries:
...
-                        obs: bad
+                        obs: rank
                         time: T
                     colormap: *raincm
-                    lower_is_worse: no
-                    format: bad
+                    lower_is_worse: yes
+                    format: number0
+                bad-years-new:
+                    label: Ranked bad years
```
aaron-kaplan, line 308: I assume they still want this column to be called "Drought humanitarian ranking." Let's call the old one v2 and the new one v3, so we don't have to change names when `new` is no longer new.

kgraaf: Makes sense

aaron-kaplan (approved):

PR #255, opened 2022-12-22
Niger ENACTS flood proxy

aaron-kaplan:
A duplicate of an existing entry but with lower_is_worse set to no instead of yes, as a first attempt at highlighting years where there were probably floods. Currently under review on shortfin01.


aaron-kaplan:
@kgraaf please review.

kgraaf (approved):

PR #268, opened 2023-01-05
fix colormap tests

aaron-kaplan:
These were broken by https://github.com/iridl/python-maprooms/pull/244/files, which changed how we interpret the ingrid colorscale representation.


remicousin (approved):
I approve but the current open PR is working on resorting back to RGB and possibly revisit the representation of colorscale in those fbf config files. But I guess regardless you need those tests to work in the meantime.

PR #269, opened 2023-01-12
One flask wsgi app with many dash apps attached

aaron-kaplan:
My main goal was to reduce the number of processes, because I saw something like forty processes running in prod on ZMD and I figured it was ten for each app. But I can't actually reproduce that on my laptop, so I can't tell if I've fixed it or not. I still think this version is worth merging because it eliminates the Werkzeug dispatcher, which is not meant for production use.


kgraaf:
- [ ] Test that multiple concurrent (or concurrent-ish) requests to different pages don't break anything

aaron-kaplan:
Turns out I had broken the tile URLs. I think I ignored that while testing locally because I didn't have the data, and then forgot to do a final test on devi before sending the PR. It's fixed now, and I tested with different apps open in different browser windows. @kgraaf please review again.

kgraaf:
Looks great

PR #271, opened 2023-01-18
Move Niger JAS forecast to /home/.aaron

aaron-kaplan:


aaron-kaplan:
Note that forecasts for other seasons and other variables are still coming from /home/.remic for now.

remicousin (approved):

PR #274, opened 2023-01-19
Remic/runoff2

remicousin:
Audrey FYI

Aaron for review

This adds the runoff computation. It is defined so that it can be done outside the water balance loop. it's usage is then to input to the water balance precip - runoff instead of just precip.

I am not sure what to do to test weekly_api_runoff. Maybe with dummy api_threshold and api_poly inputs? dummy enough that a calculation can be done on paper to be tested?
Corollary is how/whether to test the water balance with precip-runoff as input? Is there a point in doing so anyway? Not sure either...


enacts/wat_bal/agronomy.py
```diff
@@ -214,3 +214,105 @@ def soil_plant_water_balance(
...
+            api_poly,
+            xr.concat(
+                [np.power(daily_rain, 0), daily_rain, np.square(daily_rain)],
+                dim="powers",
+            ).where(daily_rain > no_runoff, 0),
+        ),
+        xr.concat(
+            [api <= api_thresh, ~np.isnan(api)],
+            dim="api_cat",
+        ).cumsum(dim="api_cat").where(lambda x: x <= 1, other=0),
```
aaron-kaplan, line 287: I don't understand what this concat/cumsum is doing. Please explain.

aaron-kaplan (commented):
I didn't manage to understand weekly_api_runoff. If there were a test for it, I would have stepped through the code using the test inputs, which might have helped my understanding. Yes, I think overriding the defaults with dummy/simple inputs could help with writing a comprehensible test.

I don't think we necessarily need a test that demonstrates that the output of this can be used as an input to the water balance function, unless perhaps for illustrative purposes.

remicousin:
I don't think I am going to figure out why I wrote a runoff function with numpy's select and then with xarray's dot. It looks like it was part of my initial giant commit so I might have made that move on my own thinking. Possibly I wanted to be xarray and there is no xarray's select. But is that irrelevant if I can implement the xarray's dot version with the modifications we discussed yesterday? that is switch the dot products around to realize only needed calculations?

enacts/wat_bal/agronomy.py
```diff
@@ -214,3 +214,105 @@ def soil_plant_water_balance(
...
+def api_sum(a, axis=-1):
+    """Weighted-sum for Antecedent Precipitation Index for an array of length n
+    applies weights of 1/2 for last element and 1/(n-i-1) for all others
+    Parameters
+    ----------
+    a : array_like
+        elements to weight and sum
+    axis : int, optional
+        Axis along which the weight and sum are performed.
+        if None, applies to last axis.
```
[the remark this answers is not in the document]
remicousin: I am not sure why I did it that way back then. What I can think of now:

- I was under the impression that rolling.reduce applied to the grouped dimension created by rolling by design but I may be wrong. 
- api_sum is extremely specific so I think I aimed at factoring it out for ease of read and not meant it to be reused by anything else, and thus figured a simpler numpy think was sufficient

enacts/wat_bal/agronomy.py
```diff
@@ -214,3 +214,105 @@ def soil_plant_water_balance(
...
+    api : ndarray
+        weighted sum of `a` along `axis` .
+    
+    See Also
+    --------
+    numpy.sum
+    """
+    api_weights = np.arange(a.shape[axis] - 1, -1, -1)
+    api_weights[-1] = 2
+    api_weights = 1 / api_weights
```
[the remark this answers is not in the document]
remicousin: yeah... that is what Ingrid does. I'll have to look back at old papers to retrieve the information. Or ask Amor.

enacts/wat_bal/agronomy.py
```diff
@@ -214,3 +214,105 @@ def soil_plant_water_balance(
...
+            [-1.14, 0.042, 0.0026],
+            [-2.34, 0.12, 0.0026],
+            [-2.36, 0.19, 0.0026],
+            [-2.78, 0.25, 0.0026],
+            [-3.17, 0.32, 0.0024],
+            [-4.21, 0.438, 0.0018],
+        ],
+        dims=["api_cat", "powers"],
+    ),
+    time_dim="T",
```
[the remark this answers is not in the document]
remicousin: Need to check back on all documents or to Amor.

enacts/wat_bal/agronomy.py
```diff
@@ -214,3 +214,105 @@ def soil_plant_water_balance(
...
+    -------
+    runoff : DataArray
+        daily Runoff.
+    See Also
+    --------
+    api_sum
+    Notes
+    -----
+    For instance with the default parameters, if rain is greater or equal to 12.5
+    and API is 18, then runoff is -1.14 + 0.042*x 0.0026*(x**2)
```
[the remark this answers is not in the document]
remicousin: Right... Should not those default values be in the documentation actually? In which case it will be visible to the doc reader.

Runoff < 0 set to 0 is documented in the first paragraph of the doc already.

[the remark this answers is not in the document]
remicousin: oh I didn't think of that: I thought I should write them down. Either way.

aaron-kaplan:
Put Eunjin's code in a `doc` directory or something like that, to make clear that it's not part of the application.

enacts/wat_bal/agronomy.py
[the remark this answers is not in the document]
remicousin: Actually, the API is used to define the categories that define which polynomial to apply to daily rain to compute runoff. It doesn't contribute to the actual calculation of the runoff. But I will right all that down as you asked for in your last comment.

remicousin:
[Quick sheets.pdf](https://github.com/iridl/python-maprooms/files/10468732/Quick.sheets.pdf)

Does this help?

remicousin:
and the runoff using numpy's select is here:
https://github.com/iridl/python-maprooms/pull/63/files

remicousin:
So now the way I see things is that using matricial calculations only does more work than needed. Current implementation realizes all 6 possibles polynomials for each t when we need just one. The other option would create a matrix of the Polynomial coefficients b of size t which is uselessly big (or is it) since there are only 6 of them.

Numpy's select would seem preferable. Or there may be a way to do this concisely with Xarray's where that I didn't see at the time.

enacts/wat_bal/agronomy.py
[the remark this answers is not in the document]
remicousin: It does happen automatically even though it's not pretty.

enacts/wat_bal/agronomy.py
```diff
@@ -214,3 +214,109 @@ def soil_plant_water_balance(
...
+            [0.858, 0.0895, 0.0028],
+            [-1.14, 0.042, 0.0026],
+            [-2.34, 0.12, 0.0026],
+            [-2.36, 0.19, 0.0026],
+            [-2.78, 0.25, 0.0026],
+            [-3.17, 0.32, 0.0024],
+            [-4.21, 0.438, 0.0018],
+        ],
+        dims=["api_cat", "powers"],
+    ),
```
aaron-kaplan, line 234: api_thresh and api_poly here are mutable defaults, which are something [we should avoid](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). Do this instead:
 ```
def weekly_api_runoff(
    ...
    api_thresh=None,
    ....
):
    if api_thresh is None:
        api_thresh = DEFAULT_API_THRESH.copy(deep=True)
```

File: enacts/wat_bal/agronomy.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (aaron-kaplan) [reply enacts/wat_bal/agronomy.py:274]: Resolve this unless you want a reminder to prettify the documentation.

File: enacts/wat_bal/agronomy.py [hunk 7]
```diff
@@ -214,3 +214,109 @@ def soil_plant_water_balance(
             - (planted_since - np.timedelta64(1, "D"))
         )
     return sm, drainage, et_crop, planting_date
+
+
+def weekly_api_runoff(
+    daily_rain,
+    no_runoff=12.5,
+    api_thresh=xr.DataArray([6.3, 19, 31.7, 44.4, 57.1, 69.8], dims=["api_cat"]),
```
aaron-kaplan, line 222: Isn't it strange that the default `api_thresh[0]` is smaller than the default `no_runoff`? Isn't that giving us a non-zero polynomial to calculate the runoff in cases where there's no runoff?

Should we avoid such inconsistencies and simplify the cold by folding no_runoff into api_poly, prepending an entry with api_thresh = no_runoff and api_poly = (0, 0, 0)?

enacts/wat_bal/agronomy.py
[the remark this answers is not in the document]
aaron-kaplan: > I was under the impression that rolling.reduce applied to the grouped dimension created by rolling

I checked the [docs](https://docs.xarray.dev/en/stable/generated/xarray.core.rolling.DataArrayRolling.html#xarray.core.rolling.DataArrayRolling). All the specific reduction functions like min, max, and mean say "Reduce this object's data windows by applying [min/max/mean] along its dimension," but rolling is different: it says "along some dimension(s)." By specifying axis=-1 you _might_ always be picking the right dimension, but it's not clear to me that that's the case. I would be more comfortable if you used `.reduce(api_sum, dim=time_dim)`.  I'm not sure if that will work directly with your numpy implementation of api_sum, or if api_sum will have to be rewritten with xarray.

enacts/wat_bal/agronomy.py
aaron-kaplan: You actually did fold no_runoff into api_thresh in the `select` version.

aaron-kaplan (commented):
Sorry, I hadn't realized that the ball was back in my court on this.

I think using [numpy.piecewise](https://numpy.org/doc/stable/reference/generated/numpy.piecewise.html) and [Polynomial](https://numpy.org/doc/stable/reference/generated/numpy.polynomial.polynomial.Polynomial.html#numpy.polynomial.polynomial.Polynomial) will let us create a more readable and hopefully more efficient implementation. And if it's not more efficient, then I think making it more efficient is more work than it's worth right now.

Let me know  if you want to work together to write that. Preferably today, before I forget what's going on again--it took me the better part of the morning just to reconstruct the understanding I arrived at two weeks ago.

enacts/wat_bal/agronomy.py
```diff
@@ -214,3 +226,153 @@ def soil_plant_water_balance(
...
+    >>>         [-2.34, 0.12, 0.0026],
+    >>>         [-2.36, 0.19, 0.0026],
+    >>>         [-2.78, 0.25, 0.0026],
+    >>>         [-3.17, 0.32, 0.0024],
+    >>>         [-4.21, 0.438, 0.0018],
+    >>>     ],
+    >>>     dims=["api_cat", "powers"],
+    >>> )
+
+    Which means dor instance that, if rain is greater or equal to 12.5
```
aaron-kaplan, line 293: "for" is misspelled.

enacts/wat_bal/agronomy.py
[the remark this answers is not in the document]
remicousin: I've left this aside for now. The documentation's [examples](https://docs.xarray.dev/en/stable/generated/xarray.core.rolling.DataArrayRolling.reduce.html#xarray.core.rolling.DataArrayRolling.reduce) suggest it applies only to the rolled dimension/window.
Will get back to it once we are happy with the main function.

enacts/wat_bal/agronomy.py
remicousin: Is that irrelevant now with the new setup?

enacts/wat_bal/agronomy.py
```diff
@@ -214,3 +226,153 @@ def soil_plant_water_balance(
...
+    if api_thresh is None:
+        api_thresh = DEFAULT_API_THRESHOLD
+    if api_poly is None:
+        api_poly = DEFAULT_API_POLYNOMIALS
+    print([daily_rain <= no_runoff]
+            + [api <= api_v for k, api_v in enumerate(api_thresh)])
+    func = lambda dr, i, nr:  np.piecewise(
+        dr,
+        [dr <= nr]
+            + [i <= api_v for k, api_v in enumerate(api_thresh)],
```
aaron-kaplan, line 341: No reason to use `enumerate` here.
```
i <= api_v for api_v in api_thresh
```

remicousin:
I pushed with another function that uses piecewise and polynomial. I wrote a test. I am under the impression that I did right but the result seems to apply everywhere the last polynomial coefficents, unless daily_rain <= no_runoff.
I can't figure out why: I might need some duck help. I printed in the code the condlist that seems to be right. And in the test, I printed the output of the function and the expected.

The [doc](https://numpy.org/doc/stable/reference/generated/numpy.piecewise.html) says funclist should take a 1d array as input. Is that what the problem is? Am I too naive in my application of apply_ufunc?

We can talk if easier.

enacts/wat_bal/agronomy.py
```diff
@@ -214,3 +226,153 @@ def soil_plant_water_balance(
...
+        [0]
+            + [lambda x: np.polynomial.polynomial.polyval(x, api_coeff)
+                for k, api_coeff in enumerate(api_poly)],
```
aaron-kaplan, line 344: I think this can be written more simply using `Polynomial` instead of `polyval`.
```
[np.polynomial.polynomial.Polynomial(api_coeff) for api_coeff in api_poly]
```
because an instance of the `Polynomial` class can be called as a function:
```
>>> p = np.polynomial.polynomial.Polynomial([1, 1])
>>> p
Polynomial([1., 1.], domain=[-1,  1], window=[-1,  1])
>>> p(1)
2.0
>>> p(2)
3.0
>>> p(3)
4.0
```

aaron-kaplan:
This definition makes the test pass:
```
def api_runoff(
    daily_rain,
    api,
    no_runoff=12.5,
    api_thresh=None,
    api_poly=None,
):
    if api_thresh is None:
        api_thresh = DEFAULT_API_THRESHOLD
    if api_poly is None:
        api_poly = DEFAULT_API_POLYNOMIALS
    conds = (
        [(daily_rain <= no_runoff).values] +
        [
            (
                (daily_rain > no_runoff) &
                ((i == 0) | (api > api_thresh[i-1])) &
                ((i == len(api_thresh)) | (api <= api_thresh[i]))
            ).values
            for i in range(len(api_thresh))
        ]
    )
    funcs = [0] + [
        np.polynomial.polynomial.Polynomial(coeffs)
        for coeffs in api_poly
    ]

    runoff = (
        xr.DataArray(
            np.piecewise(daily_rain, conds, funcs),
            dims=daily_rain.dims,
        )
        .where(~np.isnan(api), drop=True)
        .clip(min=0)
        .rename("runoff")
    )
    runoff.attrs = dict(description="Runoff", units="mm")
    return runoff
```
I figured that out after reading the [source code](https://github.com/numpy/numpy/blob/v1.24.0/numpy/lib/function_base.py#L640-L758) of `np.piecewise`, which is surprisingly readable. Our previous attempts had two problems:
- When we tried to use DataArrays as the conds, we hit the case in line 732 of the above-linked numpy file. It's intended to promote a scalar value to a list with one element, but it's badly written such that it incorrectly considers a DataArray to be a scalar. This is a bug in numpy, for which I intend to submit a PR. We can work around this by using `.values` on the DataArrays to convert them to numpy arrays.
- We assumed/hoped that when multiple conditions were true for the same position, `piecewise` would use the func for the first one, because the doc says that `piecewise` is similar to `select`, and that's how `select` handles that situation. Unfortunately that's not the case. As we can see clearly from the source code, with `piecewise` if multiple conditions are true then the last one wins, not the first. I worked around this by making the conditions more specific so it's impossible for more than one to be true.

Note that this version doesn't use `apply_ufunc`. It's not needed--numpy can translate a DataArray to a np.array. It can't translate in the other direction automatically, though, so I had to do that by hand.

This is not as succinct and easy to read as I had hoped it would be, but I think it does avoid computing the values of polynomials at points where they're not relevant.

enacts/wat_bal/agronomy.py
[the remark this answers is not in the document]
remicousin: There is now a dedicated api function as opposed to api_sum to feed an rolling calculation. The whole point was in case someone wanted to calculate runoff differently. The way it is, they now can by computing their own api, provide their own api tresholdds and polynomials.

remicousin:
Or: instead of listing all conditions, could do as before and add a step to turn all Trues after first to False

remicousin:
Oops: there was a bug with vi, so commit "typo" also contains the previous work which is dropping api_sum function and makes an api function of its own. So we can feed api_runoff with other apis if we wish.

remicousin:
I am wondering if the fact that funcs of funclist must take a 1darray as stated in [doc](https://numpy.org/doc/stable/reference/generated/numpy.piecewise.html) is not going to be a problem when applying this to bigger arrays with X and Y dimensions for instance? And I see in the [code](https://github.com/numpy/numpy/blob/v1.24.0/numpy/lib/function_base.py#L640-L758) that y[cond] is "flatenning" y and cond. Is that the piece of code you are looking to correct?

I am writing the colorscale code with piecewise now and I am trying to do it for all bands vectorially but either the above won't let me, either I am doing it wrong.

enacts/wat_bal/agronomy.py
```diff
@@ -214,3 +226,115 @@ def soil_plant_water_balance(
...
         )
     return sm, drainage, et_crop, planting_date
+
+
+def api_runoff(
+    daily_rain,
+    api,
+    no_runoff=12.5,
+    api_thresh=None,
+    api_poly=None,
```
aaron-kaplan, line 236: If you make the defaults immutable as I suggested above, then we can just say
```
api_thresh=DEFAULT_API_THRESHOLD,
api_poly=DEFAULT_API_POLYNOMIALS,
```
here and eliminate the `if var is None` stuff later on.

If you don't make them immutable, then the current version is not safe. You need to make a copy of `DEFAULT_API_POLYNOMIALS` instead of using it directly.

aaron-kaplan:
The documentation doesn't specify what happens when multiple conditions are true, so let's not rely on that. Maybe in the future someone will change it to match `select`; we don't want our code to break in that case. Let's just make sure only one condition is true per point. I don't have a strong opinion about whether that should be done the way I did it or by explicitly switching all but the first True to False.

The above may be moot given what comes next, but I've already written it so I'll leave it there.

> I see in the [code](https://github.com/numpy/numpy/blob/v1.24.0/numpy/lib/function_base.py#L640-L758) that y[cond] is "flatenning" y and cond.

Oh, I see. Also `x[cond]`. No, that's not the thing I wanted to fix. I just wanted to allow `cond` to be a DataArray. Here's the [doc](https://numpy.org/doc/stable/user/basics.indexing.html#boolean-array-indexing) about the flattening. I think what we need is to convert each cond array to a list of indices, so we can map the flattened values back into the right places in y. I can work on that. In the meantime, you could go back to the inefficient version that evaluates all the polynomials at all the points.

aaron-kaplan:
Actually np.piecewise works fine!
```
def test_piecewise():
    x = np.array([
        [1, 2, 3],
        [4, 5, 6]
    ])
    conds = [
        x % 2 == 0, # even
        x % 2 == 1  # odd
    ]
    funcs = [
        lambda x: 0, 
        lambda x: 1
    ]
    result = np.piecewise(x, conds, funcs)
    expected = [
        [1, 0, 1],
        [0, 1, 0]
    ]
    assert np.array_equal(result, expected)
```
`vals = x[cond]` flattens it, but then `y[cond] = func(vals, ...)` unflattens it back to its original shape. The problem you're having with the colorscales must be something different.

remicousin:
hmmm... I don't know... if funcs inolve objects with dimensions, it might not be able to figure out what to broadcast and what not.

aaron-kaplan:
I'm pretty sure it's ok, as long as x, y, and cond all have the same shape, which they do. x[cond] is a flat 1d array, whose i'th element is the value of x at the i'th true value of cond. And then y[cond] = foo, where foo is a flat 1d array, maps the i'th element of foo back to the position of the i'th True element of cond.

aaron-kaplan:
Going into a meeting, I can talk through this with you more another time.

remicousin:
I have something that works in part for the colorscale business but I don't understand yet the part that doesn't work.

I have a question though of which answer might help. Initially here when I rewrote your solution, I wrote:
[
       lambda x: np.polynomial.polynomial.Polynomial(coeffs)(x)
        for coeffs in api_poly
    ]
instead of 
[
        np.polynomial.polynomial.Polynomial(coeffs)
        for coeffs in api_poly
    ]

and that didn't work (It would go back to running the last function to all elements or something like that). Why is that?

Currently my colorscale solution uses lambda but I don't trust that to work anymore, what is the syntax to write a list of callables, f(x,*args,**kw). The *args are going to be the arguments that vary per condition and generated into a list of them with for. I would write my f outside piecewise and see what happens. But I don't how that's written...

Anyway, I am probably done for today.

aaron-kaplan:
I don't know. `lambda x: np.polynomial.polynomial.Polynomial(coeffs)(x)` is equivalent to `np.polynomial.polynomial.Polynomial(coeffs)` as far as I can see. I'll have to look at the full program to figure out what's going on.

enacts/wat_bal/agronomy.py
```diff
@@ -214,3 +226,122 @@ def soil_plant_water_balance(
...
+    Parameters
+    ----------
+    daily_rain : DataArray
+        daily precipitation
+    api : DataArray
+        an Antecedent Precipitiona Index. Must be same size as daily_rain.
+    no_runoff : DataArray, optional
+        `runoff` is 0 if `daily_rain` is lesser or equal to `no_runoff`
+        (default `no_runoff` =12.5)
+    api_thresh : tuple, optional
```
aaron-kaplan, line 253: No change needed, but just to be sure you understand what's going on, the argument provided by the user doesn't have to be a tuple, we just needed to make the default value a tuple. This method will work if passed a tuple, list, np.array, xr.Dataset, or anything else that can be iterated over with `for v in ...`. The term that generalizes over that class of objects is "iterable."

aaron-kaplan (approved):
Looks good to me, other than the unresolved conversation about citing a reference.

enacts/wat_bal/agronomy.py
remicousin: Then you would describe api_poly as iterable(iterable) ? or iterable of 2 dimensions?

aaron-kaplan: Yes, iterable(iterable) or iterable of iterables.

PR #275, opened 2023-01-24
New AA (née FBF) Layout

kgraaf:
A test server is currently located [here](http://devi.iri.columbia.edu:8077/fbfmaproom/niger). I would appreciate testing if possible.

### DIFF
--- a/fbfmaproom/fbflayout.py
+++ b/fbfmaproom/fbflayout.py
@@ -10,7 +10,7 @@ SEVERITY_COLORS = ["#fdfd96", "#ffb347", "#ff6961"]
 
 
 def app_layout():
-    return html.Div(
+    return dbc.Container(
         [
             dbc.Modal(
                 [
@@ -21,12 +21,15 @@ def app_layout():
                 centered=True,
             ),
             dcc.Location(id="location", refresh=True),
-            map_layout(),
             logo_layout(),
-            table_layout(),
-            command_layout(),
+            dbc.Row(command_layout()),
+            dbc.Row([
+                dbc.Col(map_layout(), id="lcol"),
+                dbc.Col(table_layout(), id="rcol"),
+            ]),
             disclaimer_layout(),
-        ]
+        ],
+        fluid=True,
     )
 
 
@@ -149,8 +152,9 @@ def map_layout():
         center=None,
         style={
             "width": "100%",
-            "height": "100%",
-            "position": "absolute",
+            # "height": "100%",
+            "height": "100vh",
+            # "position": "absolute",
         },
         closePopupOnClick=False,
     )
@@ -160,7 +164,7 @@ def logo_layout():
     return html.Div(
         [html.H4("FBF—Maproom"), html.Img(id="logo")],
         id="logo_panel",
-        className="info",
+        # className="info",
         style={
             "position": "absolute",
             "top": "10px",
@@ -171,6 +175,7 @@ def logo_layout():
             "pointerEvents": "auto",
             "paddingLeft": "10px",
             "paddingRight": "10px",
+            "display": "none",
         },
     )
 
@@ -179,7 +184,7 @@ def disclaimer_layout():
     return html.Div(
         [html.H5("This is not an official Government Maproom.")],
         id="disclaimer_panel",
-        className="info",
+        # className="info",
         style={
             "position": "absolute",
             "width": "fit-content",
@@ -190,6 +195,7 @@ def disclaimer_layout():
             "pointerEvents": "auto",
             "paddingLeft": "10px",
             "paddingRight": "10px",
+            "display": "none",
         },
     )
 
@@ -322,6 +328,23 @@ def command_layout():
                     "verticalAlign": "top",
                 },
             ),
+            html.Div(
+                [
+                    dcc.Checklist(
+                        ['Map', 'Table',],
+                        ['Map', 'Table',],
+                        id="fbf_display",
+                    ),
+                ],
+                style={
+                    "position": "relative",
+                    "width": "95px",
+                    "display": "inline-block",
+                    "padding": "10px",
+                    "verticalAlign": "top",
+                },
+            ),
+
             dbc.Alert(
                 "No data available for selected month and year",
                 color="danger",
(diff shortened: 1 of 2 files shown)


fbfmaproom/fbflayout.py
```diff
@@ -171,6 +175,7 @@ def logo_layout():
             "pointerEvents": "auto",
             "paddingLeft": "10px",
             "paddingRight": "10px",
+            "display": "none",
```
aaron-kaplan, line 178: Need to reenable logo, disclaimer

fbfmaproom/fbflayout.py
```diff
@@ -335,18 +358,18 @@ def command_layout():
...
+        #     "position": "absolute",
+        #     "top": "10px",
+        #     "right": "10px",
+        #     "left": "230px",
+        #     "zIndex": "1000",
+        #     "height": "fit-content",
+        #     "pointerEvents": "auto",
+        #     "paddingLeft": "10px",
+        #     "paddingRight": "10px",
+        # },
```
aaron-kaplan, line 372: Font has changed, text no longer fits in menus.

aaron-kaplan: The fonts in the control bar look good now, but the fonts in the table controls are still weird.

aaron-kaplan:
@kgraaf do you have more recent changes to push?

fbfmaproom/fbflayout.py
```diff
@@ -21,12 +21,16 @@ def app_layout():
                 centered=True,
             ),
             dcc.Location(id="location", refresh=True),
-            map_layout(),
-            logo_layout(),
-            table_layout(),
-            command_layout(),
+            # logo_layout(),
```
aaron-kaplan, line 24: Logo needs to be re-enabled.

fbfmaproom/fbflayout.py
```diff
@@ -337,15 +379,15 @@ def command_layout():
...
-            "paddingRight": "10px",
+        #     "position": "absolute",
+        #     "top": "10px",
+        #     "right": "10px",
+        #     "left": "230px",
+        #     "zIndex": "1000",
+        #     "height": "fit-content",
+        #     "pointerEvents": "auto",
+        #     "paddingLeft": "10px",
+        #     "paddingRight": "10px",
```
aaron-kaplan, line 390: Uncomment or remove.

fbfmaproom/fbflayout.py
```diff
@@ -414,31 +457,33 @@ def table_layout():
...
+            # "height": "80vh",
+        #     "position": "absolute",
+        #     "top": "110px",
+        #     "right": "10px",
+        #     "zIndex": "1000",
+        #     "bottom": "50px",
+        #     "width": "600px",
+        #     "pointerEvents": "auto",
+        #     "paddingLeft": "10px",
+        #     "paddingRight": "10px",
```
aaron-kaplan, line 486: Uncomment or remove.

fbfmaproom/fbflayout.py
```diff
@@ -149,8 +154,9 @@ def map_layout():
         center=None,
         style={
             "width": "100%",
-            "height": "100%",
-            "position": "absolute",
+            # "height": "100%",
+            "height": "85vh",
+            # "position": "absolute",
```
aaron-kaplan, line 159: Remove comments

fbfmaproom/fbfmaproom.py
```diff
@@ -745,11 +745,37 @@ def calculate_bounds(pt, res, origin):
...
+            rclass = "d-none"
+        }
+
+        return [ lclass, rclass, zoom ]
+    }
+
+    """,
+    Output("lcol", "className"),
+    Output("rcol", "className"),
+    Output("map", "zoom"),
```
aaron-kaplan, line 768: I think you must not have this commit in the version that's running on devi. When I run this version, I get an error in the js console about multiple callbacks trying to update the same output, which is not allowed.

aaron-kaplan (changes requested):
I'm 99% certain that what's in this PR is not what you have running on devi. Please get a working version into the PR, and also address the other comments I made earlier and some more I made below.

aaron-kaplan:
We need to pin protobuf to 3.20.x. When I create an environment from environment.yml I get protobuf 4.x, which is not compatible with any version of geobuf.

Protobuf 4 isn't recent, so why didn't we have this problem before? I haven't confirmed, but I'm guessing that the conda package metadata for geobuf has the appropriate constraint on protobuf, but in this PR we switch to installing protobuf with pip instead of conda, and the pip metadata is probably missing that constraint.

fbfmaproom/fbflayout.py
```diff
@@ -149,179 +168,128 @@ def map_layout():
...
-            "left": "90px",
-            "zIndex": "1000",
-            "height": "fit-content",
-            "pointerEvents": "auto",
-            "paddingLeft": "10px",
-            "paddingRight": "10px",
-        },
-    )
+def disclaimer_layout():
+    return
```
aaron-kaplan, line 177: Did you have a reason to keep this?

fbfmaproom/fbflayout.py
```diff
@@ -149,179 +168,128 @@ def map_layout():
...
 
 
-def disclaimer_layout():
+def command(label, tool, cmd, width="105px"):
```
aaron-kaplan, line 180: Why did you name this "command?" I suggest "control".

aaron-kaplan:
labels in table still use different front from those in control bar.

aaron-kaplan:
When the window is tall, there's a lot of unused space at the bottom. Can we make map and table go to the bottom of the window, and let the disclaimer overlap?

kgraaf:
> When the window is tall, there's a lot of unused space at the bottom. Can we make map and table go to the bottom of the window, and let the disclaimer overlap?

I did this but now the disclaimer blocks part of the table but perhaps that's ok. Can't overlap the map because then it blocks the colorbar

aaron-kaplan:
Was there a reason we couldn't just leave a fixed number of pixels below the table for it, rather than a percentage of viewport height? If that's not feasible, then blocking part of the table seems ok.

aaron-kaplan (approved):

aaron-kaplan (commented):
Seems like the conflicts are significant, so I'd like to review again after it's rebased. Unapproving.

aaron-kaplan:
Replaced by https://github.com/iridl/python-maprooms/pull/313

Issue #253, opened 2022-12-19
clean up enactstozarr.py and its config entries


remicousin:
Indeed, it's dirty. It would be nice to have one config entry that provides all the enacts dataset info, with a consistent structure.

I'm making the scope of this PR creep larger and larger, just because I wasn't reviewing as carefully when we were in a rush ahead of the training. We can make an issue for this and fix it later.

_Originally posted by @aaron-kaplan in https://github.com/iridl/python-maprooms/pull/236#discussion_r1051176424_

PR #277, opened 2023-01-26
Enactstozarr

drewmresnick:
improving the file for enacts zarr conversion

### DIFF
--- a/enacts/config-gmet.yaml
+++ b/enacts/config-gmet.yaml
@@ -28,12 +28,23 @@ chunks:
     X: 29
 zarr_path: /data/remic/mydatafiles/ #top level directory to zarr data
 vars:
-   #var name in file name: [input file path, output file path, var name def in file]
-    precip: [Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/, Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/, precip]
-    tmax: [Ghana/version3/GHANA_ENACTS/ALL/mrg_tmax_daily/, Ghana/version3/GHANA_ENACTS/ALL/mrg_tmax_daily/, temp]
-    tmin: [Ghana/version3/GHANA_ENACTS/ALL/mrg_tmin_daily/, Ghana/version3/GHANA_ENACTS/ALL/mrg_tmin_daily/, temp]
+   #var name in file name:
+   #    - input file path
+   #    - output file path
+   #    - var name def in file
+    precip:
+        - Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/ 
+        - Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/ 
+        - precip
+    tmax: 
+        - Ghana/version3/GHANA_ENACTS/ALL/mrg_tmax_daily/
+        - Ghana/version3/GHANA_ENACTS/ALL/mrg_tmax_daily/
+        - temp
+    tmin: 
+        - Ghana/version3/GHANA_ENACTS/ALL/mrg_tmin_daily/
+        - Ghana/version3/GHANA_ENACTS/ALL/mrg_tmin_daily/
+        - temp
 
-Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/
 # Onset
 onset:
 
--- a/enacts/config-gmet.yaml
+++ b/enacts/config-gmet.yaml
@@ -34,15 +34,15 @@ vars:
    #    - var name def in file
     precip:
         - Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/ 
-        - Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/ 
+        - null 
         - precip
     tmax: 
         - Ghana/version3/GHANA_ENACTS/ALL/mrg_tmax_daily/
-        - Ghana/version3/GHANA_ENACTS/ALL/mrg_tmax_daily/
+        - null
         - temp
     tmin: 
         - Ghana/version3/GHANA_ENACTS/ALL/mrg_tmin_daily/
-        - Ghana/version3/GHANA_ENACTS/ALL/mrg_tmin_daily/
+        - null
         - temp
 
 # Onset
--- a/enacts/config-gmet.yaml
+++ b/enacts/config-gmet.yaml
@@ -29,8 +29,8 @@ chunks: #if regridding, consider updating sizes
 zarr_path: /data/remic/mydatafiles/ #top level directory to zarr data
 vars:
    #var name in file name:
-   #    - input file path
-   #    - output file path ## set to null if same as input file path
+   #    - input file path relative to nc_path
+   #    - output file path relative to zarr_path (set to null if same as input file path)
    #    - var name def in file
     precip:
         - Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/ 
--- a/enacts/config-gmet.yaml
+++ b/enacts/config-gmet.yaml
@@ -30,7 +30,7 @@ zarr_path: /data/remic/mydatafiles/ #top level directory to zarr data
 vars:
    #var name in file name:
    #    - input file path
-   #    - output file path
+   #    - output file path ## set to null if same as input file path
    #    - var name def in file
     precip:
         - Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/ 
--- a/enacts/config-gmet.yaml
+++ b/enacts/config-gmet.yaml
@@ -20,14 +20,20 @@ shapes_adm:
       is_checked: True
           
 # Daily ENACTS Zarr conversion
-rr_mrg_nc_path: /Data/data23/Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/
-rr_mrg_resolution: .0375
-rr_mrg_chunks:
+nc_path: /Data/data23/ #top level directory to nc data
+resolution: 0 #set to 0 for no regridding
+chunks:
     T: 600
     Y: 15
     X: 29
-rr_mrg_zarr_path: /data/remic/mydatafiles/Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/
+zarr_path: /data/remic/mydatafiles/ #top level directory to zarr data
+vars:
+   #var name in file name: [input file path, output file path, var name def in file]
+    precip: [Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/, Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/, precip]
+    tmax: [Ghana/version3/GHANA_ENACTS/ALL/mrg_tmax_daily/, Ghana/version3/GHANA_ENACTS/ALL/mrg_tmax_daily/, temp]
+    tmin: [Ghana/version3/GHANA_ENACTS/ALL/mrg_tmin_daily/, Ghana/version3/GHANA_ENACTS/ALL/mrg_tmin_daily/, temp]
 
+Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/
 # Onset
 onset:
 
--- a/enacts/config-gmet.yaml
+++ b/enacts/config-gmet.yaml
@@ -21,8 +21,8 @@ shapes_adm:
           
 # Daily ENACTS Zarr conversion
 nc_path: /Data/data23/ #top level directory to nc data
-resolution: 0 #set to 0 for no regridding
-chunks:
+resolution: null #set to null for no regridding
+chunks: #if regridding, consider updating sizes
     T: 600
     Y: 15
     X: 29
--- a/enacts/config-nma.yaml
+++ b/enacts/config-nma.yaml
@@ -28,7 +28,7 @@ shapes_adm:
 # Daily ENACTS Zarr conversion
 rr_mrg_zarr_path: /data/aaron/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/rr_mrg_ALL-staging/ #this I could delete and we could use the new file path defs in the maprooms?
 nc_path: /Data/data23/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to find input data
-resolution: .0375
+zarr_resolution: .0375
 chunks:
     T: 578
     Y: 25
--- a/enacts/config-nma.yaml
+++ b/enacts/config-nma.yaml
@@ -26,14 +26,7 @@ shapes_adm:
       is_checked: False
 
 # Daily ENACTS Zarr conversion
-rr_mrg_nc_path: /Data/data23/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/
-rr_mrg_resolution: .25
-rr_mrg_chunks:
-    T: 578
-    Y: 25
-    X: 31
-rr_mrg_zarr_path: /data/aaron/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/rr_mrg_ALL-staging/
-
+rr_mrg_zarr_path: /data/aaron/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/rr_mrg_ALL-staging/ #this I could delete and we could use the new file path defs in the maprooms?
 nc_path: /Data/data23/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to find input data
 resolution: .0375
 chunks:
@@ -42,7 +35,7 @@ chunks:
     X: 31
 zarr_path: /data/drewr/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to put output file
 data_src: 
-   #var name: [input file path, output file path]
+   #var name in file name: [input file path, output file path, var name def in file]
     precip: [Rainfall/daily/, Rainfall/daily/r0375/rr_mrg_ALL/, precip]
     tmax: [Temperature/daily/tmax/, Temperature/daily/tmax/r0375/rr_mrg_ALL/, temp]
     tmin: [Temperature/daily/tmin/, Temperature/daily/tmin/r0375/rr_mrg_ALL/, temp]
--- a/enacts/config-nma.yaml
+++ b/enacts/config-nma.yaml
@@ -28,7 +28,7 @@ shapes_adm:
 # Daily ENACTS Zarr conversion
 nc_path: /Data/data23/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to find input data
 zarr_resolution: null #set to null for no regridding
-chunks: 
+chunks: #consider changing chunk sizes if regridding 
     T: 578
     Y: 25
     X: 31
--- a/enacts/config-nma.yaml
+++ b/enacts/config-nma.yaml
@@ -35,8 +35,8 @@ chunks: #consider changing chunk sizes if regridding
 zarr_path: /data/drewr/Ethiopia/ #top level directory path to put output file
 vars: 
    #var name in file name: 
-   #    - input file path
-   #    - output file path ## set to null if it is identical to input file path
+   #    - input file path relative to nc_path
+   #    - output file path relative to zarr_path (set to null if it is identical to input file path)
    #    - var name def in file
     precip: 
         - NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/
--- a/enacts/config-nma.yaml
+++ b/enacts/config-nma.yaml
@@ -27,8 +27,8 @@ shapes_adm:
 
 # Daily ENACTS Zarr conversion
 nc_path: /Data/data23/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to find input data
-zarr_resolution: 0 #set to 0 for no regridding
-chunks:
+zarr_resolution: null #set to null for no regridding
+chunks: 
     T: 578
     Y: 25
     X: 31
--- a/enacts/config-nma.yaml
+++ b/enacts/config-nma.yaml
@@ -36,7 +36,7 @@ zarr_path: /data/drewr/Ethiopia/ #top level directory path to put output file
 vars: 
    #var name in file name: 
    #    - input file path
-   #    - output file path
+   #    - output file path ## set to null if it is identical to input file path
    #    - var name def in file
     precip: 
         - NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/
--- a/enacts/config-nma.yaml
+++ b/enacts/config-nma.yaml
@@ -26,9 +26,8 @@ shapes_adm:
       is_checked: False
 
 # Daily ENACTS Zarr conversion
-rr_mrg_zarr_path: /data/aaron/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/rr_mrg_ALL-staging/ #this I could delete and we could use the new file path defs in the maprooms?
 nc_path: /Data/data23/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to find input data
-zarr_resolution: .0375 #set to None if you don't want regridding
+zarr_resolution: 0 #set to 0 for no regridding
 chunks:
     T: 578
     Y: 25
@@ -36,9 +35,9 @@ chunks:
 zarr_path: /data/drewr/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to put output file
 vars: 
    #var name in file name: [input file path, output file path, var name def in file]
-    precip: [Rainfall/daily/, Rainfall/daily/r0375/rr_mrg_ALL/, precip]
-    tmax: [Temperature/daily/tmax/, Temperature/daily/r0375/tmax/tmax_mrg_ALL/, temp]
-    tmin: [Temperature/daily/tmin/, Temperature/daily/r0375/tmin/tmin_mrg_ALL/, temp]
+    precip: [Rainfall/daily/, Rainfall/daily/rr_mrg_ALL/, precip]
+    tmax: [Temperature/daily/tmax/, Temperature/daily/tmax/tmax_mrg_ALL/, temp]
+    tmin: [Temperature/daily/tmin/, Temperature/daily/tmin/tmin_mrg_ALL/, temp]
 
 # Onset
 onset:
--- a/enacts/config-nma.yaml
+++ b/enacts/config-nma.yaml
@@ -34,6 +34,19 @@ rr_mrg_chunks:
     X: 31
 rr_mrg_zarr_path: /data/aaron/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/rr_mrg_ALL-staging/
 
+nc_path: /Data/data23/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to find input data
+resolution: .0375
+chunks:
+    T: 578
+    Y: 25
+    X: 31
+zarr_path: /data/drewr/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to put output file
+data_src: 
+   #var name: [input file path, output file path]
+    precip: [Rainfall/daily/, Rainfall/daily/r0375/rr_mrg_ALL/, precip]
+    tmax: [Temperature/daily/tmax/, Temperature/daily/tmax/r0375/rr_mrg_ALL/, temp]
+    tmin: [Temperature/daily/tmin/, Temperature/daily/tmin/r0375/rr_mrg_ALL/, temp]
+
 # Onset
 onset:
 
--- a/enacts/config-nma.yaml
+++ b/enacts/config-nma.yaml
@@ -34,10 +34,22 @@ chunks:
     X: 31
 zarr_path: /data/drewr/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to put output file
 vars: 
-   #var name in file name: [input file path, output file path, var name def in file]
-    precip: [Rainfall/daily/, Rainfall/daily/rr_mrg_ALL/, precip]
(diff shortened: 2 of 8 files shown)


enacts/enactstozarr.py
```diff
@@ -11,34 +11,35 @@
...
+    netcdf = list(sorted(Path(nc_path).glob("*.nc")))
+    
+    data = xr.open_mfdataset(
+        netcdf,
+        preprocess = set_up_dims,
+        parallel=False
+    )[var_name]
+    
+    if not np.isclose(data['X'][1] - data['X'][0], RESOLUTION):
     # TODO this method of regridding is inaccurate because it pretends
```
drewmresnick, line 43: @remicousin not sure if I was to keep this regridding code in the function since you had not originally included in the function you started working on. I put it in there for now. Mybe we can make it a parameter that can be defined so you can choose if you want to use the regridding or not?

enacts/config-nma.yaml
```diff
@@ -20,13 +20,19 @@ shapes_adm:
...
     T: 578
     Y: 25
     X: 31
-rr_mrg_zarr_path: /data/aaron/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/rr_mrg_ALL-staging/
+zarr_path: /data/drewr/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to put output file
+data_src: 
+   #var name in file name: [input file path, output file path, var name def in file]
+    precip: [Rainfall/daily/, Rainfall/daily/r0375/rr_mrg_ALL/, precip]
+    tmax: [Temperature/daily/tmax/, Temperature/daily/tmax/r0375/rr_mrg_ALL/, temp]
+    tmin: [Temperature/daily/tmin/, Temperature/daily/tmin/r0375/rr_mrg_ALL/, temp]
```
aaron-kaplan, line 35: Maybe we should we just mirror the structure of the original ENACTS dataset, so that element 0 and element 1 of each config are the same and can be combined? I imagine someone had a reason to add those extra directory levels.

enacts/enactstozarr.py
```diff
@@ -11,34 +11,35 @@
...
-RR_MRG_ZARR_PATH = CONFIG["onset"]["rr_mrg_zarr_path"]
-RESOLUTION = CONFIG["onset"]["rr_mrg_resolution"]
-CHUNKS = CONFIG["onset"]["rr_mrg_chunks"]
-
-#Read daily files of daily rainfall data
-#Concatenate them against added time dim made up from filenames
-
-RR_MRG_PATH = Path(RR_MRG_NC_PATH)
-RR_MRG_FILE = list(sorted(RR_MRG_PATH.glob("*.nc")))
+RESOLUTION = CONFIG["resolution"]
```
aaron-kaplan, line 14: Rename this ZARR_RESOLUTION to avoid confusion?

enacts/config-nma.yaml
```diff
@@ -20,13 +20,19 @@ shapes_adm:
...
-rr_mrg_chunks:
+rr_mrg_zarr_path: /data/aaron/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/rr_mrg_ALL-staging/ #this I could delete and we could use the new file path defs in the maprooms?
+nc_path: /Data/data23/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to find input data
+resolution: .0375
+chunks:
     T: 578
     Y: 25
     X: 31
-rr_mrg_zarr_path: /data/aaron/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/rr_mrg_ALL-staging/
+zarr_path: /data/drewr/Ethiopia/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to put output file
```
aaron-kaplan, line 30: I don't understand why there are two zarr paths. Is that what you meant by your comment "this I could delete..." above? If rr_mrg_zarr_path is unused then definitely get rid of it.

enacts/config-nma.yaml
```diff
@@ -20,13 +20,19 @@ shapes_adm:
...
+data_src:
```
aaron-kaplan, line 31: Call this `vars` or `variables` instead of `data_src`?

enacts/config-nma.yaml
drewmresnick: I added the r0375 directory level because when the previous script was run it was run at a different resolution, so I thought it would make sense to have the option to create other directories if someone chooses to rerun the script at a different resolution like I did. As for the rr_mrg_ALL directory level, I cannot say. I just kept that because the previous path included it.

drewmresnick:
@aaron-kaplan @remicousin once it is confirmed that the structure I proposed in the config-nma.yaml file for the vars dict is okay, then I will go back in and update all of the config files with that structure and will also go and update all the maproom files where the old file path configuration def is found to reflect the change. I just did not want to touch them until I know that the structure makes the most sense.

enacts/config-nma.yaml
aaron-kaplan: Oh, I had it backwards. I thought the one with r0375 in it was the original. I agree with having subdirs for different zarr resolutions.

Not a fan of rr_mrg_ALL. I don't even know what the rr means.

aaron-kaplan:
At some point we should come up with a coherent specification of a full ENACTS 
 site, including both ALL and MON, and covering multiple versions. ENACTS methodology seems to be pretty well standardized, but for whatever reason, the directory naming patterns are not, so it would be good to have a standard dataset structure that our maprooms all use internally, and then the job of the config file would be to map from the idiosyncratic directory structure of an individual ENACTS instance and the standard internal data structure.

This doesn't feel particularly urgent to me, though. I'm fine with the structure you've made here for now. I'm not sure what @remicousin had in mind when he asked you to work on this so I'll wait for him to comment.

remicousin:
The goal of this PR is to make enactstozarr (re)usable again. My work on the wat_bal_monit maproom triggered rationalization of multiple copies of enactstozarr (among other things) which I did poorly as documented in this issue #253 

Historically, we try to be careful with our Data Cataloguing structure since: i) it's public facing; ii) it's what is ultimately invoked by Ingrid programs and iii) we have control of that. But not that careful with structure of how data files are organized. With Python Maprooms invoking data files directly and without an intermediary catalog, that frontier has vanished. And in order to keep track of things, we have had zarr directory structure correspond to data files directory structure (thus the less cautiously curated organization). Now, the resolution version 0.25 vs 0.0375 is not from ENACTS products but from us producing different zarr products; and thus by definition that resolution substructure is not part of the structure organizing ENACTS data files.

Long story short to say: it's a bit messy as enactstozarr involves a bunch of different things: nc files we read, zarr files we write, zarr files used by this or that app. And so as a consequence I unresolved the conversation about this and added a comment as the current proposition is probably not ok.

enacts/enactstozarr.py
```diff
@@ -48,53 +49,27 @@ def set_up_dims(xda):
...
+    
+    print(f"conversion for {variable} complete.")
     return zarr
 
-
-convert("rfe")
-convert("tmin")
-convert("tmax")
+for i in CONFIG['vars']:
+    convert(i)
```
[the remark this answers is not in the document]
aaron-kaplan: I have something in progress that will let us incrementally add data to a zarr store as the upstream dataset gets extended in time, rather than re-converting the entire dataset on every run of the script. Once that's available, we'll rewrite this script in that paradigm; in the meantime, don't worry about it.

enacts/enactstozarr.py
aaron-kaplan: Sure, making `resolution` a parameter of the `convert` function makes sense.

enacts/config-nma.yaml
aaron-kaplan: I really think the path of the native resolution ENACTS data should be global, not page-specific. If certain pages need parts of their calculations done ahead of time for speed reasons (e.g. regridding in this case), those can be page-specific.

But my first thought here to wonder whether we really need to do the regridding ahead of time. The Ingrid version does the regridding on the fly, so maybe we can do that here too, and avoid the complexity of having multiple copies of the data at different resolutions. @drewmresnick are you up to trying that?

remicousin: Then we should first try if no regridding at all is acceptable for the Ethiopia Onset maproom first. The regridding in Ingrid was only for performance issue. That test might have to be done on a machine more performant than devi to be relevant.

aaron-kaplan: We did try the onset calculation at the native resolution, and that was too slow. There have been some other performance improvements since then, so I guess it's worth another try.

enacts/enactstozarr.py
drewmresnick: Just to confirm, it would be a parameter where if it is set to true then it will compare the resolution such as in line 42, to attempt the regridding?

aaron-kaplan: I imagined it would be a number. It could be optional--if it's not provided, then don't regrid. If it is provided, then check the resolution of the original dataset, and if it's not the same as the desired resolution, then regrid.

enacts/config-nma.yaml
drewmresnick: I restructured / renamed the temp data to address Rémi's comment.

enacts/config-nma.yaml
drewmresnick: I am unclear exactly what you are asking me to do @aaron-kaplan. Do you want me to work to create a function that can be applied to the data on-the-fly inside of a maproom instead of running the regridding on the data during conversion such as it is inside this file?

aaron-kaplan: That's what I was suggesting, yes. But then Rémi made a different suggestion, which was to try not regridding at all, just use the dataset's native resolution in the onset maproom. I don't know if that will work, but it's easy to try, so let's try it.

enacts/enactstozarr.py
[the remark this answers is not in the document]
drewmresnick: Yes and I guess my thought was someone could just comment out variables they wouldn't want for now.

drewmresnick:
> The goal of this PR is to make enactstozarr (re)usable again. My work on the wat_bal_monit maproom triggered rationalization of multiple copies of enactstozarr (among other things) which I did poorly as documented in this issue #253
> 
> Historically, we try to be careful with our Data Cataloguing structure since: i) it's public facing; ii) it's what is ultimately invoked by Ingrid programs and iii) we have control of that. But not that careful with structure of how data files are organized. With Python Maprooms invoking data files directly and without an intermediary catalog, that frontier has vanished. And in order to keep track of things, we have had zarr directory structure correspond to data files directory structure (thus the less cautiously curated organization). Now, the resolution version 0.25 vs 0.0375 is not from ENACTS products but from us producing different zarr products; and thus by definition that resolution substructure is not part of the structure organizing ENACTS data files.
> 
> Long story short to say: it's a bit messy as enactstozarr involves a bunch of different things: nc files we read, zarr files we write, zarr files used by this or that app. And so as a consequence I unresolved the conversation about this and added a comment as the current proposition is probably not ok.

I am not exactly sure where this conversation leaves me in terms of what I am working on in this pull request.. are we to discuss further the regridding, and as a result also the file structure of the data?

enacts/enactstozarr.py
drewmresnick: So then do you have in mind a different test than line 42?

`if not np.isclose(data['X'][1] - data['X'][0], ZARR_RESOLUTION):`

`ZARR_RESOLUTION` would be the number it would check against. So then if we are to use `ZARR_RESOLUTION` as the parameter I would think we would need to 1) probably rename that var because it is vague that this is what it is being used for and 2) be more specific as to what you would set `ZARR_RESOLUTION` as in the config file if you do not want to test for regridding.

aaron-kaplan: The name of the parameter can just be `resolution`, and the configuration variable would be `CONFIG['onset']['resolution']`. Set it to `null` if you want to use the dataset's native resolution.

remicousin:
I would say:
1/ move the current zarr data to a 0p25 subrepo (better to use 'p' to remove some ambiguity so r0375 should be r0p0375)
2/ create the native resolution zarr data without any resolution suprepo
3/ test onset maproom on that zarr data (maybe not on devi)
4/ if it seems to work ok: remove the regridding from the enactstozarr code
5/ if it doesn't: have the regridding done inside the onset maproom directly -- test performance of that
6/ if works ok. done
7/ if doesn't work ok: put back regridding option in enactstozarr code

yeah?

aaron-kaplan:
> better to use 'p' to remove some ambiguity

Why not "." instead of "p"?

Otherwise we are in agreement.

remicousin:
I am not sure if "." is too much of a special character but if not, that is all fine with me to use it.

aaron-kaplan:
> I am not sure if "." is too much of a special character

Only in DOS and Windows prior to Windows 95, AFAIK.

drewmresnick:
> I would say: 1/ move the current zarr data to a 0p25 subrepo (better to use 'p' to remove some ambiguity so r0375 should be r0p0375) 2/ create the native resolution zarr data without any resolution suprepo 3/ test onset maproom on that zarr data (maybe not on devi) 4/ if it seems to work ok: remove the regridding from the enactstozarr code 5/ if it doesn't: have the regridding done inside the onset maproom directly -- test performance of that 6/ if works ok. done 7/ if doesn't work ok: put back regridding option in enactstozarr code
> 
> yeah?

I tested the Zambia onset maproom using the regridded data and the data at the native resolution without any regridding. The runtime is not significantly greater (for example to return a new onset graph the difference is maybe 50-100 ms.) So it seems to me that it should work fine at the native resolution.

remicousin:
I am a bit confused because Zambia didn't have a 0.25 version, only Ethiopia. Did you create one for your test? Also I would think the testing would matter slightly more on the time to map rather than the time to draw graphs... but maybe not...

Regardless, it seems you are all set to create the Ethiopia ENACTS 0.0375 zarr files and you should go ahead with that. I would put no extra directory in the folder path to reflect resolution if it's the native one. Then if someone really needs to create a regridded version, then they will add a   r0.25/ to their structure. And so, we can assume now that the Ethiopia onset Maproom will work ok at 0.0375 but Aaron should still add r.025/ to his zarr files path structure for the record and in case we need to revert back to it. And if we do then we'll revisit the config file set up.

remicousin (approved):

drewmresnick:
So to finish this PR, I will update the file structure so that there is only a subdirectory for resolution if we have something regridded (such as the r0p25 subd) but will move forward assuming that the original resolution will work. 

I am going to edit all of the config files to reflect the changes since so far I only applied it to one config file.

Also @remicousin I know you said we could remove the regridding function but maybe it makes sense to keep it in the file just in case we ever need it? I can just make it so that the conversion to zarr will never attempt to regrid by default. But at least in case someone ever needs it, it could still be there.

remicousin:
yes I think we can probably leave the regridding function invokable by the enactstozarr code.

drewmresnick:
@remicousin I know you already approved this PR but I am wondering if you wanted to look at the last changes I pushed according to the comments that were made, before I merge this?

enacts/enactstozarr.py
```diff
@@ -48,53 +36,45 @@ def set_up_dims(xda):
...
-def set_up_dims(xda):
-    datestr = Path(xda.encoding["source"]).name.split("_")[2]
-    year = int(datestr[0:4])
-    month = int(datestr[4:6])
-    dekad = int(datestr[6:7])
-    day = (dekad - 1) * 10 + 1
-    xda = xda.expand_dims(T = [dt.datetime(year, month, day)])
-    xda = xda.rename({'Lon': 'X','Lat': 'Y'})
-    return xda
+        print("Your data will be regridded. Refer to function documentation for more information on this.")
```
remicousin, line 39: @aaron-kaplan should we use raise warnings for this kind of stuff? not sure what that entails.

enacts/enactstozarr.py
```diff
@@ -48,53 +36,45 @@ def set_up_dims(xda):
...
+    netcdf = list(sorted(Path(nc_path).glob("*.nc")))
+    
     data = xr.open_mfdataset(
         netcdf,
         preprocess = set_up_dims,
         parallel=False
-    )#[variable]
-    print(data)
+    )[var_name]
+    if ZARR_RESOLUTION != 0:
```
remicousin, line 58: So I believe if you literally define it as 
null
in the config file, then it becomes None when you read the config file and assigne to ZARR_RESOLUTION and I assume that's probably better than using 0 (even though 0 is a good choice since it's impossible to have such a resolution but still).

enacts/enactstozarr.py
```diff
@@ -11,34 +11,22 @@
...
     return xda
 
-rr_mrg = xr.open_mfdataset(
-    RR_MRG_FILE,
-    preprocess = set_up_dims,
-    parallel=False
-).precip
-
-if not np.isclose(rr_mrg['X'][1] - rr_mrg['X'][0], RESOLUTION):
+def regridding(data, resolution):
```
remicousin, line 28: For the record, I would create an issue about regridding, unless already there but I don't think so, including the TODO comment below and including the desire to generalize to other dimensions.

enacts/config-nma.yaml
```diff
@@ -20,13 +20,18 @@ shapes_adm:
...
 # Daily ENACTS Zarr conversion
-rr_mrg_nc_path: /Data/data23/NMA_Ethiopia_v7/ALL_NEW/Rainfall/daily/
-rr_mrg_resolution: .25
-rr_mrg_chunks:
+nc_path: /Data/data23/NMA_Ethiopia_v7/ALL_NEW/ #top level directory path to find input data
+zarr_resolution: 0 #set to 0 for no regridding
+chunks:
     T: 578
     Y: 25
     X: 31
```
remicousin, line 28: Should not the chunks vary with resolution? since we want to strike a balance between number of chunks and their size? and avoid chunks of very different sizes? How to deal with that? Maybe leave a comment about it here so that someone who's going to regrid doesn't forget to think about their chunking?

remicousin (changes requested):
Not sure how to unapprove... so posting an additional empty review requesting changes to see if that would change my approval status.

enacts/enactstozarr.py
drewmresnick: Ah yes thank you, I was confused as to why defining it as `None` was not working..

enacts/enactstozarr.py
drewmresnick: an issue for improving the method used for the regridding?

remicousin: yes. so that it is on the "issues" radar and so that whoever is going to tackle it is going to be reminded there is this piece of code here.

enacts/config-gmet.yaml
```diff
@@ -15,14 +15,20 @@ shapes_adm:
...
     T: 600
     Y: 15
     X: 29
-rr_mrg_zarr_path: /data/remic/mydatafiles/Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/
+zarr_path: /data/remic/mydatafiles/ #top level directory to zarr data
+vars:
+   #var name in file name: [input file path, output file path, var name def in file]
+    precip: [Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/, Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/, precip]
+    tmax: [Ghana/version3/GHANA_ENACTS/ALL/mrg_tmax_daily/, Ghana/version3/GHANA_ENACTS/ALL/mrg_tmax_daily/, temp]
+    tmin: [Ghana/version3/GHANA_ENACTS/ALL/mrg_tmin_daily/, Ghana/version3/GHANA_ENACTS/ALL/mrg_tmin_daily/, temp]
```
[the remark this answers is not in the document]
aaron-kaplan: This is equivalent to
```
vars:
    precip:
        - Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/
        - Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/
        - precip
    tmax:
        ...
```
Either is acceptable, but in this case I have a slight preference for breaking it up, because the line is rather long and hard to scan this way.

enacts/config-nma.yaml
aaron-kaplan: Yes

drewmresnick: I do think the idea would be to change the chunks at different resolutions - however is it so simple a concept that we are expecting users to know what sizes to be chunking the data? I have not been part of the process for deciding on sizes for chunking so I cannot say for sure, though from what I know so far it seems like something of a case-by-case basis when it comes to deciding on sizes. 

I can at least make a note that says something like: 
"Be aware that changes to resolution may also require different chunking sizes. if regridding, consider updating chunk sizes"

I guess this also brings up the question: should I change the chunk sizes for the Ethiopia maproom, since we are using the native resolution now instead of the regridded data?

enacts/enactstozarr.py
aaron-kaplan: If we go with the suggestion I made [elsewhere](https://github.com/iridl/python-maprooms/pull/277/files/3251997db01508771fb8000718d78c54db9c284b#r1091296468) , `resolution` will be null unless the config author intended to regrid it. In that case I don't think a warning is necessary.

enacts/config-nma.yaml
aaron-kaplan: If it's fast enough, then it's good enough for now. Someday we'll do some experimentation to identify good rules for picking a chunking, and hopefully automate it.

remicousin: For now the "users" are just us really so I think just a comment to remind the next "user" would do. Plus all that is temporary as well as Aaron mentioned.

drewmresnick:
@remicousin I applied the changes you had suggested and based on comments you and Aaron made so it should be ready for a final review now

enacts/config-sng.yaml
```diff
@@ -30,13 +30,30 @@ shapes_adm:
...
-rr_mrg_resolution: .0375
-rr_mrg_chunks:
+nc_path: /Data/data23/ANACIM/anacim_v3/ALL_block2/rr_mrg_daily/
+zarr_resolution: null #set to null for no regridding
+chunks: #consider changing chunk sizes if regridding
     T: 592
     Y: 31
     X: 29
-rr_mrg_zarr_path: /data/aaron/ANACIM/anacim_v3/ALL_block2/rr_mrg_daily
+zarr_path: /data/aaron/ANACIM/anacim_v3/ALL_block2/rr_mrg_daily
```
remicousin, line 39: here it's not going to work out: ANACIM/anacim_v3/ALL_block2/ need be either with the paths, or with the vars.

enacts/enactstozarr.py
```diff
@@ -48,53 +36,45 @@ def set_up_dims(xda):
...
+    )[var_name]
+    if ZARR_RESOLUTION != None:
+        print("attempting regrid")    
+        data = regridding(data, ZARR_RESOLUTION)
 
     data = data.chunk(chunks=CONFIG['chunks'])
-
-    zarr = f"{CONFIG['data_dir']}/{variable}.zarr"
+    
+    zarr = f"{CONFIG['zarr_path']}{CONFIG['vars'][variable][1]}"
```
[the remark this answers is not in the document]
remicousin: or generally speaking is there anything that checks that we are not going to overwrite something pre-existing?

remicousin (changes requested):

remicousin:
As we work on this, I am now seeing two other problems that will need to be addressed but maybe not in this PR.

One is that we are ignoring the case that zarrifies dekadal data (used by monthly maproom). I must have run it when at ZMD to generate the zarr files for their monthly version but I don't quite recall what I used and I don't see how set_up_dims could set the time correctly for dekads there.

The 2nd problem is that set_up_dims is pretty optimistic on how the ENACTS data files names are with a specific split made.

enacts/enactstozarr.py
```diff
@@ -48,53 +36,45 @@ def set_up_dims(xda):
...
-if not os.access(CONFIG["data_dir"], os.W_OK | os.X_OK):
-    sys.exit("can't write to output directory")
-
-def set_up_dims(xda):
-    datestr = Path(xda.encoding["source"]).name.split("_")[2]
-    year = int(datestr[0:4])
-    month = int(datestr[4:6])
-    dekad = int(datestr[6:7])
-    day = (dekad - 1) * 10 + 1
-    xda = xda.expand_dims(T = [dt.datetime(year, month, day)])
```
remicousin, line 71: AH! Actually: this is the dekad data case... that's why there were 2 set_up_dims... One came from enactstozarr working on daily data, the other from dekadal...

:(

Sorry Drew, I am afraid you are going to have to reconcile this. Maybe 2 functions: set_up_daily_dims and set_up_dekad_dims and a new argument to convert to tell which one to use?

Then you probably want the config files that use these to be set up like you're setting up the daily data. I think only kmd and zmd have it. And then monthly/maproom.py and layout.py will have to use the right CONFIG info...

Sorry I didn't see that earlier...

enacts/enactstozarr.py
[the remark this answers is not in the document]
drewmresnick: It looks to me like I forgot to add the `.zarr` at the end of this line of code. I would think that if I added the file type definition back to this line of code we would not run into an issue here where the data would be overwritten even if the output is to the same directory. But I can also look into how to check if a file already exists under the same name and if so, to throw an error.

enacts/enactstozarr.py
drewmresnick: Okay, yes in this case I can resolve the other things and then maybe you can take over this part.

enacts/config-gmet.yaml
```diff
@@ -20,13 +20,30 @@ shapes_adm:
...
-rr_mrg_zarr_path: /data/remic/mydatafiles/Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/
+zarr_path: /data/remic/mydatafiles/ #top level directory to zarr data
+vars:
+   #var name in file name:
+   #    - input file path
+   #    - output file path
+   #    - var name def in file
+    precip:
+        - Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/ 
+        - Ghana/version3/GHANA_ENACTS/ALL/mrg_precip_daily/
```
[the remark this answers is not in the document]
drewmresnick: I can do this yes. I will wait for your response on the other conversation regarding the structure of the config file before I do anything about this part

enacts/config-nma.yaml
```diff
@@ -26,13 +26,30 @@ shapes_adm:
...
+        - Rainfall/daily/rr_mrg_ALL/
+        - precip
+    tmax: 
+        - Temperature/daily/tmax/ 
+        - Temperature/daily/tmax/tmax_mrg_ALL/
+        - temp
+    tmin: 
+        - Temperature/daily/tmin/
+        - Temperature/daily/tmin/tmin_mrg_ALL/
+        - temp
```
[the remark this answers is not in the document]
remicousin: I think it makes sense to reproduce the path structure from the nc files to the zarr ones. In that case we need only nc_path and zarr_path to indicate the top part of the path that changes.

drewmresnick:
@remicousin just finished my sanity check, the data looks to be saving correctly. :) I got nervous because something went wrong and one of the zarr files I made was empty when I tried to use it, but now the files I re-made are filled with the data so I think there was something I did on my end when trying to generate data that messed up the conversion.

drewmresnick:
@remicousin can you look at the commit I just pushed to see if this is what you had in mind for the file paths defs?

As per your second comment, are you asking me to update the dictionary as you suggested where the output file path by default is none, meaning that the input and output file dirs are the same?

drewmresnick:
@remicousin I think I am understanding more now what you are asking, are you hoping it would be like this:

nc_path: top level path to where nc files are
zarr_path: top level path to where zarr files will go

vars:
    var_name:
        - common file path 
        - alternate output file path (if not identical to the common file path)
        - var name in data

remicousin:
yes that is what I had in mind. Please apply to all other countries.

drewmresnick:
@remicousin I added the change to the remaining config files

remicousin (approved):
I approve but we can wait for Aaron's

drewmresnick:
@aaron-kaplan when you have some time could you do a final look through of this PR so we can merge it?

aaron-kaplan (approved):
Just minor comments.

drewmresnick:
@remicousin this branch is merged now for your use.

enacts/enactstozarr.py
```diff
@@ -48,53 +37,48 @@ def set_up_dims(xda):
...
+            Y=np.arange(data.Y.min(),data.Y.max() + resolution, resolution),
+        )    
+    return data
 
 def convert(variable):
-    netcdf = list(sorted(Path(CONFIG['data_src'][variable]).glob("*.nc")))
-
+    print(f"converting files for: {variable}")
+    
+    var_name = var_name[2]
```
aaron-kaplan, line 50: This doesn't look right. Did you test?

drewmresnick: @aaron-kaplan No I did not test because of how long it takes for the code to run, and I thought it was trivial. Rémi is going to open a new PR to continue work on this so should he just correct it in his PR? or should I open a quick PR to fix the issue?

Sorry about that..

aaron-kaplan: Not urgent, since this isn't being deployed anywhere.

drewmresnick: Okay, then when @remicousin opens his PR to continue working on this, he can make the change - just for clarification, when you suggested this

`input_path, output_path, var_name = CONFIG['vars'][variable]`

Were you implying making three separate variables directly calling the config var such as ex:

`var_name = CONFIG['vars'][variable][2]`

aaron-kaplan:
```
input_path, output_path, var_name = CONFIG['vars'][variable]

```
is a more concise way of writing
```
input_path = CONFIG['vars'][variable][0]
output_path = CONFIG['vars'][variable][1]
var_name = CONFIG['vars'][variable][2]
```

drewmresnick: Oh I see, so then you do not have to put the [0], [1], [2] at the end of each variable. Thank you for the clarification. :)

PR #279, opened 2023-01-27
update readme so that the example runtime cmd will work

drewmresnick:


enacts/wat_bal/README.md
```diff
@@ -29,7 +29,7 @@
 
 * Start the development server using both config files, e.g.:
 
-    `CONFIG=../config.yaml:../config-zmd.yaml:config.yaml python maproom_monit.py`
+    `CONFIG=../config.yaml:../config-sng.yaml:../config.yaml python maproom_monit.py`
```
remicousin, line 32: then you have to give it another name like myconfig.yaml.

enacts/config.yaml and config-sng.yaml are both part of the git repo, whereas "myconfig.yaml" is your own that you create to overwrite whatever you need to overwrite to make the app work in your environment.

drewmresnick: Are you just saying that in this example we should specify for the user to create a copy of `config.yaml` - `enacts/myconfing.yaml` for their own use? Would they not also then need to make a copy of `config-sng.yaml` such as `enacts/myconfig-sng.yaml` if this is the case since they theoretically could be changing both files?

remicousin: no: each yaml overwrites/adds to the previous yaml file(s). So if/when you want to change both files for your local work, you can do so all into your one myconfig.yaml. And your personal myconfig.yaml needs not be a copy: it just needs to contain what you want to change/add to the other 2 config files.

drewmresnick: Okay I edited the file to reflect this hopefully clearly.

remicousin (approved):

PR #280, opened 2023-01-27
Crop suitability

drewmresnick:


enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,545 @@
...
+            lat = nearest_grid["Y"].values
+            lng = nearest_grid["X"].values
+        except KeyError:
+            lat = lat
+            lng = lng
+    return [lat, lng], lat, lng
+
+def sel_year_season(data,target_season,target_year):
+    seasonal = data.sel(T=data['T.season']==target_season)
+    if target_season == "DJF": #need december of previous year right now this does not work
```
aaron-kaplan, line 263: Yeah, I think we need to label each day with an identifier of the season it belongs to. E.g. define the season as a start date and a length (in months, to avoid issues with leap years?), and label each day with the start date of its season, then filter on the month of that, and use the year of the start of the season as the coordinate of the resulting subset.

You don't have to invent this all from scratch. @remicousin has done a lot of it in the onset maproom.

aaron-kaplan (commented):
Not a thorough review, but here are some comments to get you started.

As I said by email, you should check out the dry spell functionality in the onset maproom. It's not exactly what you need, but I think it will help.

remicousin:
For dry spells defined as Tufa did, there is longest_run_length that could be a source of inspiration. Instead of taking the max at the end, could divide by the length that defines a dry spell, that would give a count of spells within a series of continuous spells, and then sum that to count all the spells in the season. I think... need to check what goes into the max is what I claim here, it's been a while...
Dividing (or using modulo) should be done on integers so that, if the threshold for spells is 5 days, 6 will return 1, 14 will return 2, etc.. which sounds appropriate.
Now there could be an approach outside longest_run_length that is better but I'd have to think more about it.

enacts/crop_suitability/layout_crop_suit.py
```diff
@@ -0,0 +1,374 @@
...
+                [
+                    html.H5(
+                        [
+                            CONFIG["crop_suit_title"],
+                        ]
+                    ),
+                    html.P(
+                        f"""
+                        """
+                    ),
```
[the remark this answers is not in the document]
drewmresnick: I think there was additional inputs from the onset maproom that I emptied while I was playing around with the layout and then forgot to fully delete them. I will do it now.

enacts/crop_suitability/layout_crop_suit.py
```diff
@@ -0,0 +1,374 @@
...
+from pathlib import Path
+import calc
+import pingrid
+import pandas as pd
+
+
+GLOBAL_CONFIG = pingrid.load_config(os.environ["CONFIG"])
+CONFIG = GLOBAL_CONFIG["crop_suit"]
+
+DR_PATH = GLOBAL_CONFIG["rr_mrg_zarr_path"]
```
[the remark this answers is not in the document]
remicousin: Yes I am not able to run the app at present because of this. But @aaron-kaplan , I am also not able to run the other app (wat_bal) that is not part of the multi-page I think because I don't know how to make it know globals_....

remicousin:
Is the analysis meant to be performed on a single year chosen by the user? Or will there be analysis through years at some point (either some reduction to make map or some repetition to make time series)? Asking for a friend. :) I mean: asking to think thoroughly of the dry spell problem.

drewmresnick:
> Is the analysis meant to be performed on a single year chosen by the user? Or will there be analysis through years at some point (either some reduction to make map or some repetition to make time series)? Asking for a friend. :) I mean: asking to think thoroughly of the dry spell problem.

@remicousin As of now the crop suitability analysis is used both to generate the map for a single year and a time series plot over all available years of data. So I am trying to build it so that it can work both for a single year and for multiple years.

remicousin:
As a general comment, can we name the app similarly to the CSMT [Maproom](http://iridl.ldeo.columbia.edu/maproom/Health/Regional/Africa/Malaria/CSMT/index.html)? I think the inspiration comes from there. So something like Climate Suitability fro Crop ... something-something.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,553 @@
...
+    try:
+        if data_choice == "precip_map":
+            data_var = pingrid.sel_snap(data.precip, lat1, lng1)
+            isnan = np.isnan(data_var).sum().sum()
+        elif data_choice == "suitability_map":
+            data_var = pingrid.sel_snap(data.crop_suit, lat1, lng1)
+            isnan = np.isnan(data_var).sum().sum()            
+        else:
+            data_var = pingrid.sel_snap(data.temp, lat1, lng1)
+            isnan = np.isnan(data_var).sum().sum()
```
remicousin, line 399: I think we concluded elsewhere that one sum() was enough

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,553 @@
...
+        if isnan > 0:
+            error_fig = pingrid.error_fig(error_msg="Data missing at this location")
+            germ_sentence = ""
+            return error_fig, error_fig, germ_sentence
+    except KeyError:
+        error_fig = pingrid.error_fig(error_msg="Grid box out of data domain")
+        germ_sentence = ""
+        return error_fig, error_fig, germ_sentence
```
remicousin, line 407: germ_sentence belongs to onset maproom and so you can remove it.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,553 @@
...
+
+    if data_choice == "suitability_map":
+        seasonal_suit = data_var
+        timeseries_plot = pgo.Figure()
+        timeseries_plot.add_trace(
+            pgo.Scatter(
```
remicousin, line 413: user Bars for time series with potential missing values in it. See current master onset date for examples

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,553 @@
...
+        ) 
+    else:
+        data_var.load()
+    
+        seasonal_var = data_var.sel(T=data_var['T.season']==target_season)
+        seasonal_mean = seasonal_var.groupby("T.year").mean("T")
+        
+        timeseries_plot = pgo.Figure()
+        timeseries_plot.add_trace(
+            pgo.Scatter(
```
remicousin, line 433: use Bars

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,553 @@
...
+        timeseries_plot.update_layout(
+            xaxis_title = "years",
+            yaxis_title = f"{CONFIG['map_text'][data_choice]['data_var']} ({CONFIG['map_text'][data_choice]['units']})",
+            title = f"{CONFIG['map_text'][data_choice]['menu_label']} seasonal climatology timeseries plot"
+        ) 
+    else:
+        data_var.load()
+    
+        seasonal_var = data_var.sel(T=data_var['T.season']==target_season)
+        seasonal_mean = seasonal_var.groupby("T.year").mean("T")
```
remicousin, line 429: it looks like this is the only difference with the case crop_suitability, since you are checking on the case earlier, maybe you can do that there and thus you need not have if cases to make the graph?

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,553 @@
...
+        )
+
+    return timeseries_plot
+
+@APP.callback(
+    Output("map_title","children"),
+    Input("target_year","value"),
+    Input("target_season","value"),
+)
+def write_map_title(target_year,target_season):
```
remicousin, line 453: to reconsider depending whether we really only want those 4 seasons as choices.

remicousin (changes requested):
I did a first review. Key things:

- what seasons do we want as options;
- consider my proposition on how to select seasons/years
- note that when rebasing, the colorscales won't work as the new system has merged
- consider leaving the dry spell component to another PR to limit scope of this PR
- I would like to review more when I am able to build

drewmresnick:
@remicousin I just pushed the changes so that you should be able to build the maproom by running 
`CONFIG=../config.yaml:../config-nmal:../myconfig.yaml python maproom_crop_suit.py`

Where myconfig.yaml just has your db password. Note it is slow so it will take a minute to load the map and time series plot.

Going back now to look at the additional comments and will respond to them / push additional commits as it appropriate.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,545 @@
...
+    target_season,
+    dry_spell_rain,
+    dry_days_in_row,
+    number_dry_spells,
+):  
+    #seasonal_year_precip = sel_year_season(rainfall_data,target_season,target_year)
+    #seasonal_year_tmax = sel_year_season(tmax_data,target_season,target_year)
+    #seasonal_year_tmin = sel_year_season(tmin_data,target_season,target_year)
+    seasonal_precip = rainfall_data.sel(T=rainfall_data['T.season']==target_season).load()
+    seasonal_tmax = tmax_data.sel(T=tmax_data['T.season']==target_season).load()
```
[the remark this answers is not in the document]
drewmresnick: Yes, while xarray does have the grouping by season functionality, we will want to move away from this. I just wanted to get a simple example working and then work out from there. The goal, according to the parameters outlined by Tufa, would be that the user can select the season and also the maximum season length. So then I suppose the best way to do that would be to allow the user to choose a season start date and a season length and from that we would find the season start and end dates to define the season.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,545 @@
...
+        pingrid.to_dash_colorscale(pingrid.RAINBOW_COLORMAP),
+        mymap_max,
+        [0,0.5,1],
+    )
+
+if __name__ == "__main__":
+    APP.run_server(
+        debug=GLOBAL_CONFIG["mode"] != "prod",
+        processes=GLOBAL_CONFIG["dev_processes"],
+        threaded=False,
```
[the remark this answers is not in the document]
drewmresnick: To me those all look different aside from which options are defined. What you see here I copied from the [onset maproom](https://github.com/iridl/python-maprooms/blob/947d5d075782ab5f76ff15eec85273b3445250f5/enacts/onset/maproom.py#L951)

Then in ../app.py it is:

    if __name__ == "__main__":
        if GLOBAL_CONFIG["mode"] != "prod":
            import warnings
            warnings.simplefilter("error")
            debug = True
        else:
            debug = False

        FLASK.run(
            GLOBAL_CONFIG["dev_server_interface"],
            GLOBAL_CONFIG["dev_server_port"],
            debug=debug,
            extra_files=os.environ["CONFIG"].split(":"),
            processes=GLOBAL_CONFIG["dev_processes"],
            threaded=False,
        )

But in ../wat_bal/maproom_monit.py they are:

    if __name__ == "__main__":
        APP.run_server(
            host=CONFIG_GLOBAL["dev_server_interface"],
            port=CONFIG_GLOBAL["dev_server_port"],
            debug=CONFIG_GLOBAL["mode"] != "prod",
            processes=CONFIG_GLOBAL["dev_processes"],
            threaded=False,
        )

So is it only important to you which config options are defined? Or do you also want me to follow how the rest of this code is written, for example how ../app.py has a warning message while the others do not?

enacts/crop_suitability/layout_crop_suit.py
```diff
@@ -0,0 +1,374 @@
...
+                        ),
+                        "Choose a target season:",
+                        dbc.Select(
+                            id="target_season",
+                            value= "MAM",
+                            options=[
+                                {"label":"Dec-Feb", "value":"DJF"},
+                                {"label":"Mar-May", "value":"MAM"},
+                                {"label":"Jun-Aug", "value":"JJA"},
+                                {"label":"Sep-Nov", "value":"SON"},
```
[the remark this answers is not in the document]
drewmresnick: Yes this sounds like a good solution. See my other comment regarding this about one of the parameters being 'max season length' with a proposal for an alternative method of choosing seasons. Although that case and this one seem fairly similar.

drewmresnick:
> As a general comment, can we name the app similarly to the CSMT [Maproom](http://iridl.ldeo.columbia.edu/maproom/Health/Regional/Africa/Malaria/CSMT/index.html)? I think the inspiration comes from there. So something like Climate Suitability fro Crop ... something-something.

As simple as: "Seasonal Climate Suitability for Crops in Ethiopia" ?

enacts/crop_suitability/layout_crop_suit.py
```diff
@@ -0,0 +1,358 @@
...
+                    ),
+                    Block(
+                        "Dry spells",
+                        #Sentence( #Originally was going to be able to define peri
+                                    #this but I simplified for now
+                        #    "Maximum number of dry spells within the first",
+                        #    Number("number_months", 3, min=0, max=12),
+                        #    "months of the season:",
+                        #    Number("number_dry_spells", 4, min=0, max=99),
+                        #),
```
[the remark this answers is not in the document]
drewmresnick: Sure I can do that. Should I also work on the selection of seasons / years in a separate PR or should I include that in here as well?

remicousin:
> @remicousin I just pushed the changes so that you should be able to build the maproom by running `CONFIG=../config.yaml:../config-nmal:../myconfig.yaml python maproom_crop_suit.py`
> 
> Where myconfig.yaml just has your db password. Note it is slow so it will take a minute to load the map and time series plot.
> 
> Going back now to look at the additional comments and will respond to them / push additional commits as it appropriate.

It builds without failing but I am not sure how you set this up but it is so that I can't change the port and hostname so basically devi:8050 so I won't do that because you might be using it for testing purposes as you work.

But I think we are missing something from @aaron-kaplan to build individual pages again.

remicousin:
> > As a general comment, can we name the app similarly to the CSMT [Maproom](http://iridl.ldeo.columbia.edu/maproom/Health/Regional/Africa/Malaria/CSMT/index.html)? I think the inspiration comes from there. So something like Climate Suitability fro Crop ... something-something.
> 
> As simple as: "Seasonal Climate Suitability for Crops in Ethiopia" ?

Maybe just Climate Suitability for Crops. I would remove seasonal because CSMT does make a climatological analysis and shows indeed the seasonal cycle only. Here we are going to show year-to-year variability as well. You can add "in Ethiopia" but that part would have to be dealt with through the national config file. I typically would not add it because the app has a vocation to be part on a website dedicated to Ethiopia, so there is no need to add Ethiopia everywhere in titles or so.

enacts/crop_suitability/layout_crop_suit.py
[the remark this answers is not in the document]
remicousin: My gut feeling is that we should deal with the seasons/years controls as I suggested here, for those for xarray seasons, and leave it for later to offer more specific choices for seasons. With the hope that the year-dependent won't be affected, or barely, even though I can't guarantee that as we are all learning the best ways to do the same analysis over multiple years.

enacts/crop_suitability/layout_crop_suit.py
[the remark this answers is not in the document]
remicousin: I am not seeing the comment about "max season length" but I sometimes still get lost on the github PR pages :)
But as I mentioned ealrier, I would go about setting the season/year controls as I suggested for the 4 xarray seasons. Then if a parameter like max season length is not compatible with such ways, I would remove it from this version and make an Issue for a next version, like dry spells.

Or please point me to where this comment is and maybe I'll see a low-hanging fruit solution that deserves not a new Issue/PR.

remicousin (changes requested):

drewmresnick:
> > @remicousin I just pushed the changes so that you should be able to build the maproom by running `CONFIG=../config.yaml:../config-nmal:../myconfig.yaml python maproom_crop_suit.py`
> > Where myconfig.yaml just has your db password. Note it is slow so it will take a minute to load the map and time series plot.
> > Going back now to look at the additional comments and will respond to them / push additional commits as it appropriate.
> 
> It builds without failing but I am not sure how you set this up but it is so that I can't change the port and hostname so basically devi:8050 so I won't do that because you might be using it for testing purposes as you work.
> 
> But I think we are missing something from @aaron-kaplan to build individual pages again.

I copied the water balance maproom since that is the other standalone maproom. I am not sure about why you would not be able to change the port and hostname.. I had not tried to do so though I have been working on shortfin if that makes any difference to you.

aaron-kaplan:
> It builds without failing but I am not sure how you set this up but it is so that I can't change the port and hostname so basically devi:8050 so I won't do that because you might be using it for testing purposes as you work.

In case it wasn't clear, that was the reason for [this comment](https://github.com/iridl/python-maprooms/pull/280#discussion_r1157624946).

> But I think we are missing something from @aaron-kaplan to build individual pages again.

I started working on that, but a lot of other things have taken priority. I think the way things are set up here is workable for now, just need the right parameters in the `run` command.

enacts/config-nma.yaml
```diff
@@ -167,21 +167,25 @@ crop_suit:
...
             description: Coverage for the region for maximum temperature data.
             units: C
             data_var: temp
+            map_max: 40
         tmin_map:
             menu_label: Minimum temperature
             description: Coverage for the region for minimum temperature data.
             units: C
             data_var: temp
+            map_max: 30
```
drewmresnick, line 188: @remicousin For these values I am not sure how to decide what the max should be for temp and precip. I tried looking at the mean and max values of the temp and precip data and I did not feel like the values I was getting were giving very good visuals for the map. I am wondering if you have any suggestions or insights on choosing the max values?

drewmresnick:
The remaining tasks to do are:

1. Improve speed (I have a meeting with Aaron today to discuss this).
2. Add tests for suitability function (move this function to a different file?)
3. Fix / improve season def
4. Add / fix parameters in suitability function:
    - Maximum season length
    - Minimum number of wet days in a season
    - maximum number of dry spells in first x months of season
I recieved a clarification email from Tufa regarding the seasons def, "the length of the season is the duration between onset and cessation, which are defined by the onset and cessation parameters"
So actually now I think that the season is not a start date and length defined by the user, but actually parameters for onset and cessation dates which are used as the start and end dates of the rainy season. I replied to Tufa to confirm this though.

As Rémi and I discussed, most of this will be done in separate PR's. I will try to fix the loading speed of the current state before merging this branch, and other smaller fixes but then to deal with the additional parameters in the suitability function and the fixing of the season def, I will open new PR's. Sound good?

enacts/crop_suitability/maproom_crop_suit.py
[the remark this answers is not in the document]
aaron-kaplan: @drewmresnick please address this. It's making it harder for me to help you with the branch, because I can't run it on shortfin01 if anyone else is using port 8050.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,545 @@
...
+        mymap_max,
+        [i for i in range(0, mymap_max + 1) if i % tick_freq == 0],
+        #[0,1,2,3,4,5],
+    )
+
+if __name__ == "__main__":
+    APP.run_server(
+        GLOBAL_CONFIG["dev_server_interface"],
+        GLOBAL_CONFIG["dev_server_port"],
+        debug=GLOBAL_CONFIG["mode"],
```
aaron-kaplan, line 541: Should be
```
debug = GLOBAL_CONFIG["mode"] != "prod"
```
or copy the more complex logic from app.py, which also turns warnings into errors when in debug mode so we don't miss them.

drewmresnick: Is it preferred that I use the more complex logic? I did not add it in when I just pushed the correction but I can add it if it is preferred.

drewmresnick:
@aaron-kaplan @remicousin I have finished the final improvements to the first iteration of this maproom. The other conversations we had regarding selection of season and the calculation for dry spells will be addressed in separate PR's that I will start once this is merged. 

So I am looking for any final comments on this and hoping to get this merged asap to begin the improvements on it.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,548 @@
...
+    tmax_data,
+    tmin_data,
+    lower_wet_threshold,
+    upper_wet_threshold,
+    max_temp,
+    min_temp,
+    temp_range,
+    target_season,
+):
+    if isinstance(rainfall_data,xr.DataArray):
```
aaron-kaplan, line 261: I don't understand this if/else block. I think one case is where it's called for the timeseries plot and the other is for the map, but I don't see any reason it should be a Dataset in one case and a DataArray in the other--that just seems like a coincidence that happens to hold in the current version but might not in the future. Also, I don't understand why you need to select out data from the target season in one case but not in the other. Wouldn't you need to do that in both cases?

aaron-kaplan: Yes, I think it would be good to do the season selection in one place that's on the path for both map and time series. Then in the map case we select a single year, and in the time series we select a single point.

> The difference in the data structure is a result of the timeseries data being one lat/lng for all timesteps versus the tile being all lat/lng for one point in time. 

There's nothing about those two array shapes that dictates that one must be a Dataset and the other a DataArray. Both xarray classes can represent both shapes. There's just some inessential detail of the current implementation that makes one of them come out a Dataset and the other a DataArray. If you needed to know the array's shape, you should look directly at its shape (check how many dimensions it has, check if it has a dimension called `T`, ...), not at whether it's stored in a Dataset or a DataArray. But if you move the selection of the season out of this function as suggested above, then I don't think you even need to know the shape anymore. The `if` case and the `else` case can be collapsed into one.

aaron-kaplan: That sounds like a bug in `sel_snap`--if given a Dataset, it should return a Dataset. I'll file an issue for that. But in any case, I don't think it makes any difference for this application.  Most operations that can be applied to a Datset can also be applied to a DataArray and vice versa.

aaron-kaplan: > If that is not good practice I can make a new empty dataset to append the outputs to.

Yeah, that seems more robust to future changes.

drewmresnick: Okay, I changed it so that now I create an empty dataset like this:

    crop_suitability = xr.Dataset(
        data_vars = dict(
        ),
        coords = dict(
            X = avg_tmax["X"],
            Y = avg_tmax["Y"],
            year = avg_tmax["year"],
        ),
    )

As of now I am still having a hard time figuring out a clean way to append data to an empty dataset in both cases of appending from a dataArray and dataset, since the way you can select variables and data from within the two is different.

I will keep looking but right now the easiest thing I think to do is to convert the dataarrays to datasets before passing them into the crop_suitability() function.

For example:

    crop_suitability = crop_suitability.assign(
        max_temp = tmax[temp_var], min_temp = tmin[temp_var],
        temp_range = avg_temp_range[temp_var], precip_range = total_precip_range[precip_var],
        wet_days = wet_days[precip_var])

Where tmax, tmin, etc are dataset/datarrays with outputs from calculations done on the original data

Trying to assign data to the variables in the new dataset using this type of indexing works for a dataset, but since the dataarray does not use variable names in the same way, it does not let you select for the data in the same way.. `tmax[temp_var]` only works for the dataset; but simply using `tmax` to call the data would not work for the dataset.

Maybe there is an easy workaround I am not seeing for this?

drewmresnick: update: I pushed the edited code for you to look at

aaron-kaplan: I looked at the new version and frankly I can't make heads or tails of it. At [one point](https://github.com/iridl/python-maprooms/pull/280/files#diff-329e8bc1434c84455197fee53abce5eb2b89e5be9cdc5b63434cc6c4879ec297R308-R311) you seem to be taking the sum of max temp, min temp, temp range, precip range, and wet days?! I think you're better off discarding your most recent changes and going back to the version this comment was on.

I gave bad advice before. Datasets and DataArrays are interchangeable when you're doing calculations on them, but they're not interchangeable when you need to get the results out. I think the root of the problem here is actually that `rr_mrg`, `tmin_mrg`, and `tmax_mrg` are Datasets when they ought to be DataArrays. In all the other maprooms they're DataArrays, see e.g. (monthly)[https://github.com/iridl/python-maprooms/blob/master/enacts/monthly/maproom.py#L59], where the `[GLOBAL_CONFIG['dekadal']['vars'][name][2]]` is the part that extracts the DataArray from the Dataset. That part is missing from e.g. [line 41](https://github.com/iridl/python-maprooms/pull/280/files#diff-329e8bc1434c84455197fee53abce5eb2b89e5be9cdc5b63434cc6c4879ec297R41) of maproom_crop_suit.py. Once everything is a DataArray, you should have an easier time combining the two cases.

BTW I don't think there's a bug in `sel_snap` after all. Every time you call `sel_snap` you're giving it a DataArray (e.g. `rr_mrg.precip`), not a Dataset (`rr_mrg`), so returning a DataArray is the expected behavior.

drewmresnick: The part of the crop_suitability() function you are referring to where I sum all of those variables is creating the index - perhaps what I name each of those variables in the dataset is midleading. What they are are just '0' or '1' if the condition was met in a prior calculating in the function. So it is adding this up to get a number 1-5 where 5 has all of the conditions met and 0 has none of the conditions met. Not actually just adding up temperature, precip, etc values. This was in the previous commit. 

Yes I did not notice how I did not select the DataArray at the beginning - this fixes my issue! I will push a commit with this improvement.

aaron-kaplan:
Makes sense.

I'm hoping @remicousin can do the detailed review. Tag me if you need input on something specific.

remicousin:
I will but you two had exchanges at a pace that was higher that my ability to lean on a long PR :)

aaron-kaplan:
I don't think it's important to read our past exchange. There was a performance problem that has been fixed, and there was another problem that I misunderstood at first, gave some bad advice and sent Drew down the wrong path, then eventually we figured it out. You can focus on the scientific/mathematical appropriateness of the calculations.

remicousin (changes requested):
I finished my review.

Overall, I am wondering if you saved some lines and readability by putting the seasonal precip/tmin/tmax inside the crop_suitability Dataset... This way you might be able to do X/Y/year selections for the whole dataset at once... and maybe synthetize other pieces of code.

A personal comment: there is a big inflation of named objects, it helps the reading only if the chosen naming is good which is very subjective. I personally try to name as least objects as possible. Python facilitates that by chaining function. Or, by using same name if an intermediary result is not going to be used again, e.g.:

toto = precip.isel()...
toto = toto.mean()...

Basically, I don't want to name things if I don't have to call them later. But as mentioned starting this comment: it's a personal style.

enacts/crop_suitability/layout_crop_suit.py
```diff
@@ -0,0 +1,337 @@
...
+                ]+[
+                    html.H5("Dataset Documentation"),
+                ],
+                style={"position":"relative","height":"25%", "overflow":"scroll"},#box holding text
+            ),
+            html.H3("Controls Panel",style={"padding":".5rem"}),
+            html.Div(
+                [
+                    Block("Timeseries plot",
+                        "Choose a data layer to view as a timeseries plot:",
```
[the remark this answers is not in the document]
drewmresnick: I fixed the description - however for me it updates the map and ts correctly so I am not sure why it fails for you..

enacts/crop_suitability/layout_crop_suit.py
```diff
@@ -0,0 +1,337 @@
...
+                        ),
+                        "Choose a target year:",
+                        dbc.Input(
+                            id = "target_year",
+                            type = "number",
+                            min = 1961,
+                            max = 2018,
+                            value = CONFIG["param_defaults"]["target_year"],
+                        ),
+                        "Choose a target season:",
```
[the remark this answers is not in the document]
drewmresnick: Thank you for noticing that! I updated it so that they will change outside of the submit button for the season and year(where applicable).

enacts/crop_suitability/layout_crop_suit.py
```diff
@@ -0,0 +1,337 @@
...
+                            options=[
+                                {"label":"Dec-Feb", "value":"DJF"},
+                                {"label":"Mar-May", "value":"MAM"},
+                                {"label":"Jun-Aug", "value":"JJA"},
+                                {"label":"Sep-Nov", "value":"SON"},
+                            ],
+                        ),
+                    ),
+                    Block(
+                        "Rainfall tolerance",
```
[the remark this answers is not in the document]
drewmresnick: I think I put this because Tufa uses "maximum/minimum temperature tolerance" when he emailed me parameters. But looking at the emails I see he does not use that wording for rainfall, he said instead "optimum seasonal total rainfall." So I can put instead "rain suitability" if that is better, or adopt his wording?

enacts/crop_suitability/layout_crop_suit.py
```diff
@@ -0,0 +1,337 @@
...
+                            Number("minimum_temp", CONFIG["param_defaults"]["min_temp"], min=-99, max=999),
+                            "C and",
+                            Number("maximum_temp", CONFIG["param_defaults"]["max_temp"], min=-99, max=99999),
+                            "C",
+                        ),
+                    ),
+                    Block(
+                        "Optimal daily temperature",
+                        Sentence(
+                            "Average daily temperature range of",
```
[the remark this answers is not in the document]
drewmresnick: This is the range of max and min temperature, for example you would not want the difference between the min and max to be more than 15C. Is there a clearer way to describe this you can think of?

enacts/config-nma.yaml
```diff
@@ -127,4 +127,73 @@ onset:
...
+        target_year: 2018
+        target_season: MAM
+        lower_wet_thresh: 500
+        upper_wet_thresh: 700
+        min_temp: 10
+        max_temp: 25
+        temp_range: 15
+        season_length: 75
+        min_wet_days: 60
+        wet_day_def: 10
```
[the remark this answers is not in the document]
drewmresnick: Sure, I was just going as much off of the example Tufa gave me; in his example he put 
Minimum number of wet days within the season: 60

But I can change this to something more realistic.. what do you think?

enacts/crop_suitability/layout_crop_suit.py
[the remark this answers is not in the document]
drewmresnick: I just your your later comment on this.. I will change it to say amplitude.

enacts/crop_suitability/layout_crop_suit.py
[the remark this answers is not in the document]
remicousin: The following is what I get when asking for precip or temp (No map and -- bolding what's most useful:)

Traceback (most recent call last):
  File "/home/remic/miniconda3/envs/enactsmulti/lib/python3.9/site-packages/xarray/core/dataarray.py", line 803, in _getitem_coord
    var = self._coords[key]
**KeyError: 'precip'**

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/remic/miniconda3/envs/enactsmulti/lib/python3.9/site-packages/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/remic/miniconda3/envs/enactsmulti/lib/python3.9/site-packages/flask/app.py", line 1519, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/remic/miniconda3/envs/enactsmulti/lib/python3.9/site-packages/flask/app.py", line 1517, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/remic/miniconda3/envs/enactsmulti/lib/python3.9/site-packages/flask/app.py", line 1503, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  **File "/data/remic/Dash/python-maprooms/enacts/crop_suitability/maproom_crop_suit.py", line 478, in cropSuit_layers
    data_tile = rr_mrg_season[data_var]**
  File "/home/remic/miniconda3/envs/enactsmulti/lib/python3.9/site-packages/xarray/core/dataarray.py", line 812, in __getitem__
    return self._getitem_coord(key)
  File "/home/remic/miniconda3/envs/enactsmulti/lib/python3.9/site-packages/xarray/core/dataarray.py", line 806, in _getitem_coord
    _, key, var = _get_virtual_variable(self._coords, key, dim_sizes)
  File "/home/remic/miniconda3/envs/enactsmulti/lib/python3.9/site-packages/xarray/core/dataset.py", line 177, in _get_virtual_variable
    raise KeyError(key)
KeyError: 'precip'
127.0.0.1 - - [11/May/2023 10:02:07] "GET /tile/5/19/15?data_choice=precip_layer&target_season=MAM&target_year=2018&min_wet_days=60&wet_day_def=10&lower_wet_threshold=500&upper_wet_threshold=700&maximum_temp=25&minimum_temp=10&temp_range=15 HTTP/1.1" 500 -
[2023-05-11 10:02:07,302] ERROR in app: Exception on /tile/5/19/14 [GET]

enacts/crop_suitability/layout_crop_suit.py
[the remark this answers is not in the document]
remicousin: Good idea to use the terms from the person who's asking for the product then. Unless they're blatantly wrong.

enacts/crop_suitability/layout_crop_suit.py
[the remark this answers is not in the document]
drewmresnick: oh yes that is something I quickly fixed before getting to this comment i think. Should be fixed once I push my new commits.

enacts/crop_suitability/layout_crop_suit.py
```diff
@@ -0,0 +1,337 @@
...
+                        "Wet days",
+                        Sentence(
+                            "Minimum number of wet days within a season:",
+                            Number("min_wet_days", CONFIG["param_defaults"]["min_wet_days"], min=0, max=99999),
+                            "days",
+                        ),
+                        Sentence(
+                            "where a wet day is a day with rainfall more than",
+                            Number("wet_day_def", CONFIG["param_defaults"]["wet_day_def"], min=0, max=9999),
+                            "mm",
```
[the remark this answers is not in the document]
drewmresnick: Oh that is a great option to have added thank you for pointing me to it!

enacts/config-nma.yaml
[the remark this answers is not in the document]
drewmresnick: Okay that makes more sense. I thought a minimum of 1 wet days sounded low... I will change the wet_day_def to be 1mm

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,553 @@
...
+    #calculate total precip
+    total_precip_range = xr.where(
+        np.logical_and(
+            sum_precip <= float(upper_wet_threshold), 
+            sum_precip >= float(lower_wet_threshold)
+        ),1, 0)
+    
+    tmax = xr.where(avg_tmax <= float(max_temp), 1, 0)
+    tmin = xr.where(avg_tmin >= float(min_temp), 1, 0)
+    avg_temp_range = xr.where(avg_daily_temp_range <= float(temp_range), 1, 0)
```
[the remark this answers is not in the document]
drewmresnick: @remicousin Can you clarify this?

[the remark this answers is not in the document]
remicousin: layout currently says:

Average daily temperature range of

it should say something like:

optimal average daily temperature amplitude below

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,553 @@
...
+            return error_fig
+    except KeyError:
+        error_fig = pingrid.error_fig(error_msg="Grid box out of data domain")
+        return error_fig
+
+    if data_choice == "suitability_layer":
+        seasonal_suit = data_var
+        timeseries_plot = pgo.Figure()
+        timeseries_plot.add_trace(
+            pgo.Bar(
```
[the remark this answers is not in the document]
drewmresnick: I created a PR for this to be done.

remicousin:
@aaron-kaplan I think I would like to merge this as it is. My recent additions makes it work with the rest of the maprooms when comes to build, and I applied some fixes that were obviously missing. I listed Issues labeled CSC that I'd like to take care of in other smaller PRs, including polishing of the code if need be but basically would like to merge that giant PR now.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,550 @@
...
+    if (
+            # When we generalize this to other datasets, remember to
+            # account for the possibility that longitudes wrap around,
+            # so a < b doesn't always mean that a is west of b.
+            x_min > data_tile['X'].max() or
+            x_max < data_tile['X'].min() or
+            y_min > data_tile['Y'].max() or
+            y_max < data_tile['Y'].min()
+    ):
+        return pingrid.image_resp(pingrid.empty_tile())
```
aaron-kaplan, line 498: This short circuit should happen as early as possible, before we waste time trying to do the calculations.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,550 @@
...
+    #calculate average daily temperature range
+    avg_daily_temp_range = (
+        seasonal_tmax - seasonal_tmin
+    ).groupby("T.year").mean("T")
+    
+    total_wet_days = xr.where(
+        seasonal_precip >= float(wet_day_def),1,0
+    ).groupby("T.year").sum("T")
+    
+    #calculate total precip
```
aaron-kaplan, line 279: What seems to be happening here is that we're constructing a mask that identifies the places and years where the total precip was within the acceptable range. If I've understood that correctly, then both this comment (total precip was calculated in line 266) and the variable name `total_precip_range` are wrong/inappropriate.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,550 @@
...
+    total_precip_range = xr.where(
+        np.logical_and(
+            sum_precip <= float(upper_wet_threshold), 
+            sum_precip >= float(lower_wet_threshold)
+        ),1, 0)
```
aaron-kaplan, line 284: I think line 280-284 is equivalent to
```
total_precip_range = (
    (sum_precip <= float(upper_wet_threshold)) &&
    (sum_precip >= float(lower_wet_threshold))
)
```
which is a little easier to read, and yields a DataArray instead of a numpy array, which might or might not make it more convenient to use later. Similiarly below.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,550 @@
...
+            x_max < data_tile['X'].min() or
+            y_min > data_tile['Y'].max() or
+            y_max < data_tile['Y'].min()
+    ):
+        return pingrid.image_resp(pingrid.empty_tile())
+
+    data_tile = data_tile.sel(
+        X=slice(x_min - x_min % RESOLUTION, x_max + RESOLUTION - x_max % RESOLUTION),
+        Y=slice(y_min - y_min % RESOLUTION, y_max + RESOLUTION - y_max % RESOLUTION),
+    ).compute()
```
aaron-kaplan, line 503: This `sel` is happening too late. The call to `crop_suitability` already calculated suitability over the entire dataset, when all we needed was one tile's worth.

aaron-kaplan:
Do you have unmerged/unpushed changes to python_maproom_emi/config.yaml? I'm trying to test it with the latest version from master and I'm getting an error about the `wat_bal` configuration being incomplete.

remicousin:
> Do you have unmerged/unpushed changes to python_maproom_emi/config.yaml? I'm trying to test it with the latest version from master and I'm getting an error about the `wat_bal` configuration being incomplete.

Have you pulled it? the [master](https://bitbucket.org/iridl/python_maproom_emi/src/master/config.yaml) branch has wat_bal: null

aaron-kaplan:
I had pulled it, but I also had something for wat_bal in my dev config file which made that maproom come back from the dead.

Now that that's fixed, it's failing because the file paths are set for running on the EMI server. Do you think it makes sense to maintain a file of IRI path overrides in either python_maproom_emi or a separate IRI-only repo to make it easier for different people at IRI to test?

enacts/crop_suitability/maproom_crop_suit.py
remicousin: Check that what I did is appropriate

enacts/crop_suitability/maproom_crop_suit.py
remicousin: I overall simplified the crop_suitability function for this and the previous comment. Have a look. I did have to multiply the variables by 1 to make sure they convert in numbers and don't stay boolean even though I don't really understand why.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,550 @@
...
+        ), 
+    )
+    crop_suitability = crop_suitability.assign(
+        max_temp = tmax, min_temp = tmin, 
+        temp_range = avg_temp_range, precip_range = total_precip_range, 
+        wet_days = wet_days)
+    crop_suitability['crop_suit'] = (
+        crop_suitability['max_temp'] + crop_suitability['min_temp'] + 
+        crop_suitability['temp_range'] + crop_suitability['precip_range'] + 
+        crop_suitability['wet_days'])
```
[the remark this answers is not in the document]
remicousin: Yes. We want to know how many of the criteria are met. I suspect that in future versions we'll have other visuals to see which ones (as maps or as ts).

enacts/crop_suitability/maproom_crop_suit.py
remicousin: See if better now

remicousin:
> I had pulled it, but I also had something for wat_bal in my dev config file which made that maproom come back from the dead.
> 
> Now that that's fixed, it's failing because the file paths are set for running on the EMI server. Do you think it makes sense to maintain a file of IRI path overrides in either python_maproom_emi or a separate IRI-only repo to make it easier for different people at IRI to test?

I don't have a strong opinion on this.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,559 @@
...
+        ), 
+    )
+    crop_suitability = crop_suitability.assign(
+        max_temp = seasonal_avg_tmax_suitability,
+        min_temp = seasonal_avg_tmin_suitability, 
+        temp_range = seasonal_avg_temp_amplitude_suitability,
+        precip_range = seasonal_total_precip_suitability,
+        wet_days = seasonal_wet_days_suitability,
+    )
+    crop_suitability['crop_suit'] = (
```
aaron-kaplan, line 300: I think here would be the appropriate place to convert the booleans to integers, rather than doing it earlier. `max_temp`, `min_temp`, etc. are naturally boolean values.

enacts/crop_suitability/maproom_crop_suit.py
aaron-kaplan: Yes, I see why you have to multiply by 1. If x and y are boolean arrays then the + in  `x + y` performs Boolean OR, even though `x.sum()` is addition, not Boolean OR. Arguably inconsistent behavior in numpy.

aaron-kaplan:
Let's keep the IRI config in python_maproom_emi. Easiest way to maintain it is probably to have the IRI one override the main one (we list both of them in CONFIG) rather than duplicating everything.

enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -0,0 +1,559 @@
...
+        crop_suitability['max_temp'] + crop_suitability['min_temp'] + 
+        crop_suitability['temp_range'] + crop_suitability['precip_range'] + 
+        crop_suitability['wet_days'])
+
+    crop_suitability = crop_suitability.dropna(
+        dim="year", how="any"
```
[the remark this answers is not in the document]
remicousin: I am reading as you do. Then it occured to me that all should be dropped since it should NA out of Ethiopia. But it's not: it's 0, probably because the booleans turn NAs to False and then to integer 0. Then I am not sure what the intend is. There is probably some careful thinking to have about how to deal with NAs for this function.

However: 1/ there is no NAs in ENACTS data so it's not applicable in this very case and 2/ I don't think there are any tests done for this function and I am not sure we can test functions in maproom.py files (or can we?)

So may we make this an issue: Comprehensive management of missing values for this function, along with proper testing for that function (that may or may not require something else to do about testing stuff inside maproom.py files)?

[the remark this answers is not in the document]
aaron-kaplan: I don't see any reason you couldn't write tests for functions in maproom.py. Do you have a specific impediment in mind? Writing tests for the UI requires infrastructure we don't have in place yet, but a function that's pure computation like this one should be testable using the same paradigm you've been using elsewhere.

I'm ok deferring proper handling of NaNs if you want.

enacts/config-nma.yaml
aaron-kaplan: @remicousin resolve this if you're satisfied with the current min and max.

aaron-kaplan (approved):
The absence of any indication that something is happening while the calculation runs is quite confusing, but that's outside the scope of this PR. Other than that, looks good.

PR #281, opened 2023-01-31
adding national shapes so that raster data will print for whole country

drewmresnick:
This is adding to the config files for countries to fix a bug in printing the raster data. When the top level geometry is a subdivision of a country then the raster data will only print for a single subdivision. So adding a national (country level) geometry allows for the whole raster layer to be printed.

### DIFF
--- a/enacts/config-gmet.yaml
+++ b/enacts/config-gmet.yaml
@@ -3,6 +3,11 @@ logo: GMet_Logo.jpg
 institution: GhanaMet
 zoom: 6
 shapes_adm:
+    - name: Country
+      color: black
+      sql: select adm0_code as key, adm0_name as label, ST_AsBinary(the_geom) as the_geom
+              from g2015_2012_0 where adm0_code = 94
+      is_checked: True
     - name: National
       color: black
       sql: select 0 as key, 'Ghana' as label,
--- a/enacts/config-gmet.yaml
+++ b/enacts/config-gmet.yaml
@@ -3,6 +3,12 @@ logo: GMet_Logo.jpg
 institution: GhanaMet
 zoom: 6
 shapes_adm:
+    - name: National
+      color: black
+      sql: select 0 as key, 'Ghana' as label,
+          ST_AsBinary(ST_Simplify(ST_Union(the_geom), .01)) as the_geom
+          from g2015_2014_1 where gid between 688 and 697
+      is_checked: True
     - name: Regions
       color: black
       sql: select gid as key, adm1_name as label, ST_AsBinary(the_geom) as the_geom
--- a/enacts/config-gmet.yaml
+++ b/enacts/config-gmet.yaml
@@ -8,12 +8,6 @@ shapes_adm:
       sql: select adm0_code as key, adm0_name as label, ST_AsBinary(the_geom) as the_geom
               from g2015_2012_0 where adm0_code = 94
       is_checked: True
-    - name: National
-      color: black
-      sql: select 0 as key, 'Ghana' as label,
-          ST_AsBinary(ST_Simplify(ST_Union(the_geom), .01)) as the_geom
-          from g2015_2014_1 where gid between 688 and 697
-      is_checked: True
     - name: Regions
       color: black
       sql: select gid as key, adm1_name as label, ST_AsBinary(the_geom) as the_geom
--- a/enacts/config-bd.yaml
+++ b/enacts/config-bd.yaml
@@ -1,6 +1,12 @@
 # Bangladesh
 zoom: 7
 shapes_adm:
+    - name: National
+      color: black
+      sql: select 0 as key, 'Bangladesh' as label,
+          ST_AsBinary(ST_Simplify(ST_Union(the_geom), .01)) as the_geom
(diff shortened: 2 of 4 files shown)


enacts/config-gmet.yaml
```diff
@@ -3,6 +3,12 @@ logo: GMet_Logo.jpg
 institution: GhanaMet
 zoom: 6
 shapes_adm:
+    - name: National
+      color: black
+      sql: select 0 as key, 'Ghana' as label,
+          ST_AsBinary(ST_Simplify(ST_Union(the_geom), .01)) as the_geom
+          from g2015_2014_1 where gid between 688 and 697
```
remicousin, line 10: it looks like we don't have specific shapefiles sources for Ghana. These are extracted from a Global dataset, that, in turn, does have country-wise shapes, so it might be just as simple in that case to pick that one geometry from the Countries shapes:
http://iridl.ldeo.columbia.edu/SOURCES/.Features/.Political/.World/.Countries_GAUL/

remicousin (changes requested):

enacts/config-gmet.yaml
drewmresnick: Sure, I am having a bit of trouble getting this to work. I found the geometry id [here](http://iridl.ldeo.columbia.edu/SOURCES/.Features/.Political/.World/.Countries_GAUL/.adm0_name/adm0_code+exch+table-+text+text+skipanyNaN+-table+.html) to be 94 so I wrote the entry to be:

    - name: Country
      color: black
      sql: select gid as key, adm0_name as label, ST_AsBinary(the_geom) as the_geom
              from g2015_2014_1 where gid = 94
      is_checked: True

Does this look right to you by a quick glance?

remicousin: I am not an sql expert but I would think so. e.g. found in catalog entries:
./Features/Hydrological/River_Basins/index.tex:31:  (select the_geom from bn_basin_clip_dd **where gid=72**) openquery

drewmresnick: I still was not able to figure this line of SQL out - and don't recall how to access the DB to explore the dataset more directly so I was just looking at the DL entry. @aaron-kaplan would you by any chance have a moment to look at this SQL query?

remicousin: Do you have more to push? if so, do so; if not, confirm all is pushed; then I'll pull and play around

drewmresnick: I do not have anything else to push.

remicousin: got it: the grid for these features is not gid but adm0_code

http://iridl.ldeo.columbia.edu/expert/IRIDB/(select%20adm0_code%20as%20key%2C%20adm0_name%20as%20label%2C%20ST_AsBinary(the_geom)%20as%20the_geom%0A%20%20%20%20%20%20%20%20%20%20from%20g2015_2014_1%20where%20adm0_code%20%3D%2094)/openquery/a-/.SOURCES/.WORLDBATH/.bath/-1/min/-a-/.the_geom/-a/X/Y/fig%3A/white/mask/blue/fill/black/countries/%3Afig/X/336.6667/371.6667/plotrange/Y/-0.8337163/26.67901/plotrange//XOVY/null/psdef//plotaxislength/432/psdef//plotborder/72/psdef/#expert

drewmresnick: Okay so it seems when I use this, the geometry it loads is still divided by region and not just a top level country geometry so still only one region's data is loaded. Not sure if this was the case when you tested it?
Screen Shot 2023-02-14 at 1 32 34 PM

If it is the same case either way maybe it is just worth joining the geometries like is done with the other config files.

remicousin: I hadn't tried actually when I saw the grid was wrong. Now I see the table is wrong. It's

g2015_2012_0

So the query should be:

http://iridl.ldeo.columbia.edu/expert/IRIDB/(select%20adm0_code%20as%20key%2C%20adm0_name%20as%20label%2C%20ST_AsBinary(the_geom)%20as%20the_geom%0A%20%20%20%20%20%20%20%20%20%20from%20g2015_2012_0%20where%20adm0_code%20%3D%2094)/openquery/a-/.SOURCES/.WORLDBATH/.bath/-1/min/-a-/.the_geom/-a/X/Y/fig%3A/white/mask/blue/fill/black/countries/%3Afig/X/336.6667/371.6667/plotrange/Y/-0.8337163/26.67901/plotrange//XOVY/null/psdef//plotaxislength/432/psdef//plotborder/72/psdef/#expert

drewmresnick: Great that solved it. thank you for your help with that!

remicousin: So there should be just the Country set here and the National should be removed, no?

remicousin (changes requested):

enacts/config-gmet.yaml
drewmresnick: Yes, thank you for catching that. Fixed it

remicousin (approved):

PR #282, opened 2023-02-03
Finish renaming config field

aaron-kaplan:
In one branch we renamed this field, and in another branch we added some more datasets, resulting in a conflict that git couldn't detect automatically.


aaron-kaplan:
@kgraaf please review, trivial 2-line change.

PR #285, opened 2023-02-11
Permalinks: URL query string reflects values of controls.

aaron-kaplan:


aaron-kaplan:
Oh, I forgot to do the marker position and the "include upcoming" checkbox. Will add those next week.

The season dropdown doesn't need to be done, because none of the maprooms have more than once choice in that menu. It will go away soon.

aaron-kaplan:
@kgraaf this is ready for review.

PR #292, opened 2023-03-08
Remic/zarrdkd

remicousin:
This was meant to "simply" reinstate the option to convert dekadal files in enactstozarr. Turned out we were not writing the same way zarr files in the daily and dekadal case (dekadal with .zarr) and thus not reading the dekadal .zarr the same way in monthly maproom than we were reading daily in onset.

This attempts to homogenize to the daily case.

As always, never too satisfied about it as the fact that the config files acts both as a sort of catalog and the reference to the source of the data for the apps makes me struggle to pick implementations choices. I also almost lost my Latin with the different occurrences and ways to name the different variables for monthly app.

This works now except for the quantile calculation for the local graph that raises an exception. It's a dask-related problem and I paste it below. I noted that printing the data in mater or in this branch shows that the data is no more an array but is a dask.array.

I'll squeeze a couple of comments in my the PR as I've gone after several problems in this one PR that I initially was hoping to do with one commit...

Exception:

  File "/home/remic/miniconda3/envs/enactsmulti/lib/python3.9/site-packages/xarray/core/computation.py", line 716, in apply_variable_ufunc
    raise ValueError(
ValueError: dimension T on 0th function argument to apply_ufunc with dask='parallelized' consists of multiple chunks, but is also a core dimension. To fix, either rechunk into a single dask array chunk along this dimension, i.e., ``.chunk(dict(T=-1))``, or pass ``allow_rechunk=True`` in ``dask_gufunc_kwargs`` but beware that this may significantly increase memory usage.

data (was all the same but just "array" and its values instead of dask.array...)

<xarray.DataArray 'precip' (month: 12)>
dask.array<stack, shape=(12,), dtype=float32, chunksize=(1,), chunktype=numpy.ndarray>
Coordinates:
    X        float64 28.01
    Y        float64 -13.01
  * month    (month) int64 1 2 3 4 5 6 7 8 9 10 11 12
Attributes:
    long_name:  Merged Station-Satellite Rainfall
    units:      mm

[3.2217151e+02 2.3528911e+02 1.9450565e+02 4.1657860e+01 2.5522552e+00
 2.7783267e-04 8.3468221e-02 1.7447278e-02 2.9066268e-01 2.5372627e+01
 1.2647671e+02 2.7487766e+02]


enacts/enactstozarr.py
```diff
@@ -44,28 +48,29 @@ def regridding(data, resolution):
...
-    data = data.chunk(chunks=CONFIG['chunks'])
+    data = data.chunk(chunks=CONFIG[time_res]['chunks'])
     
-    if output_path[1] == None:
-        zarr = f"{CONFIG['zarr_path']}{input_path[0]}{var_name[2]}"
+    if output_path == None:
+        zarr = f"{CONFIG[time_res]['zarr_path']}{input_path}"
     else:
-        zarr = f"{CONFIG['zarr_path']}{output_path[1]}{var_name[2]}" 
+        zarr = f"{CONFIG[time_res]['zarr_path']}{output_path}"
```
remicousin, line 73: I removed the var_name layer here that seemed to be redundant.

remicousin:
Also note that I adapted the ZMD config file accordingly to those changes. And possibly partially the NMA one. I'll port to all configs once confirmed that these changes are on track.

enacts/config-zmd.yaml
```diff
@@ -21,39 +21,59 @@ shapes_adm:
...
-    tmax: /Data/data23/Zambia/ALL_20200917/MERGED_tmax_dekad/
+# Daily ENACTS Zarr conversion
+daily:
+    nc_path: /Data/data23/Zambia/ #top level directory to nc data
+    zarr_resolution: null #set to null for no regridding
+    chunks: #consider changing chunk sizes if regridding
+        T: 360
+        Y: 27
+        X: 25
+    zarr_path: /data/remic/mydatafiles/ZMD/ #top level directory to nc data
```
aaron-kaplan, line 32: to zarr data / output files, not nc data.

The same mistake has been copy-pasted in other places.

enacts/enactstozarr.py
```diff
@@ -80,5 +85,6 @@ def convert(variable):
     print(f"conversion for {variable} complete.")
     return zarr
 
-for i in CONFIG['vars']:
-    convert(i)
+time_res="dekadal"
```
aaron-kaplan, line 88: This shouldn't be hard-coded, should it? Did you mean to iterate over all the resolutions?

enacts/monthly/maproom.py
```diff
@@ -211,21 +215,21 @@ def clip(x):
     tile = clip(data)
 
     groups = tile.groupby('T.year')
-    if varobj['id'] == "rfe":
+    if var == "Rainfall":
```
aaron-kaplan, line 218: I haven't traced through the code to see if this is actually the case, but I have a feeling this might break when we translate menu items into another language, or rename menu items (e.g. "ENACTS rainfall"). Menu items have display names that are distinct from the id so that you can change the display name without changing the id. Make sure program logic uses the id, not the display name.

enacts/config-nma.yaml
```diff
@@ -26,30 +26,58 @@ shapes_adm:
...
-    tmax: 
-        - NMA_Ethiopia_v7/ALL_NEW/Temperature/daily/tmax/ 
-        - NMA_Ethiopia_v7/ALL_NEW/Temperature/daily/tmax/tmax_mrg_ALL/
-        - temp
-    tmin: 
-        - NMA_Ethiopia_v7/ALL_NEW/Temperature/daily/tmin/
-        - NMA_Ethiopia_v7/ALL_NEW/Temperature/daily/tmin/tmin_mrg_ALL/
-        - temp
+daily:
+    nc_path: /Data/data23/ #top level directory path to find input data
```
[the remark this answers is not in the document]
remicousin: In my mind, the top level of this ENACTS dataset (NMA_Ethiopia_v7/ALL_NEW) doesn't belong to the top level for nc and zarr paths. nc and zarr paths have to do with the file systems that are involved whereas NMA_etc. has to do with the how the data files are curated. Besides, the default case is that the output file path is "null" in which case is equal to the input's. In other words we have no reason to change the way the data files are organized in subfolders whether they are in nc format or zarr format, on that disk system or that other one. On top of that, those paths are what best curate our zarr files so we would really want to be their own object.

enacts/enactstozarr.py
```diff
@@ -7,21 +7,25 @@
...
-      int(datestr[0:4]),
-      int(datestr[4:6]),
-      int(datestr[6:8])
-    )])
+    year = int(datestr[0:4])
+    month = int(datestr[4:6])
+    if time_res == "daily":
+        day = int(datestr[6:8])
+    elif time_res == "dekadal":
+        day = (int(datestr[6:7]) - 1) * 10 + 1
```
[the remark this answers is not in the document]
remicousin: I've only picked this up from what either Kevin or you (I think it's either Kevin or you) wrote to convert dekadal files. I haven't given it more thought for this PR.

enacts/enactstozarr.py
remicousin: I have written this keeping [this](https://github.com/iridl/python-maprooms/pull/277/files#r1091048771) in mind. But maybe that doesn't apply here?

enacts/monthly/maproom.py
remicousin: Very valid concern, however, I have not touched the menu in layout that has only a value entry thus meaning that display name = id. So that "id" here defined in the config file has not been implemented to address that issue. Or it was the intention and didn't happen.
Maybe it would be fair to leave this alone in that PR and instead put up an Issue to address all display-vs-id concern throughout our current tools?

enacts/enactstozarr.py
aaron-kaplan: Drew's comment on the conversation you linked to was

> someone could just comment out variables they wouldn't want for now.

That approach could be generalized to multiple temporal resolutions like this:
```
temporal_resolutions = [
    "daily",
    "dekadal",
]
for t in temporal_resolutions:
    for v in CONFIG['vars']:
        convert(v, t)
```

enacts/monthly/maproom.py
aaron-kaplan: ok.

enacts/enactstozarr.py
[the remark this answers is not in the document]
aaron-kaplan: OK. And I guess it's consistent with how we're handling the daily resolution, since y, m, d implicitly means 0:00 (start of the say), not 12:00 (midpoint). Making an issue to think about this more later.

aaron-kaplan (approved):

remicousin:
@aaron-kaplan can you have a look at the issue mentioned in the initial comment of the PR though? I don't think any of the fixes we've done addresses it.

remicousin:
Rebasing and adapting other countries config files, or leaving place-holders depending on where the country is at with the different products available.
Will merge likely tomorrow unless someone interferes before then.

PR #293, opened 2023-03-09
local time series for Season Length

remicousin:
Now adding the tabbed local graphs for the length of the season, with custom hover data. Can be run for ZMD. Might be a bit slow on devi.

Turned out I had to make corrections to the work just previously merged on the global config set up. I thought I had test that but maybe not... Or I had rebased poorly... (?)

### DIFF
--- a/enacts/onset/layout.py
+++ b/enacts/onset/layout.py
@@ -21,7 +21,7 @@ if not CONFIG["ison_cess_date_hist"]:
     for key in CESS_KEYS:
         CONFIG["map_text"].pop(key, None)
 
-DR_PATH = GLOBAL_CONFIG["rr_mrg_zarr_path"]
+DR_PATH = f"{GLOBAL_CONFIG['zarr_path']}{GLOBAL_CONFIG['vars']['precip'][0]}"
 RR_MRG_ZARR = Path(DR_PATH)
 
 IRI_BLUE = "rgb(25,57,138)"
@@ -413,8 +413,8 @@ def results_layout():
                     dbc.Tab(
                         [
                             html.H6(id="germination_sentence"),
-                            dbc.Spinner(dcc.Graph(id="onsetDate_plot")),
-                            dbc.Spinner(dcc.Graph(id="probExceed_onset")),
+                            dbc.Spinner(dcc.Graph(id="onset_date_plot")),
+                            dbc.Spinner(dcc.Graph(id="onset_prob_exc")),
                         ],
                         label="Onset Date",
                     ),
@@ -425,6 +425,14 @@ def results_layout():
                         ],
                         id="cess_dbct",
                         label="Cessation Date",
+                    ),
+                    dbc.Tab(
+                        [
+                            dbc.Spinner(dcc.Graph(id="length_plot")),
+                            dbc.Spinner(dcc.Graph(id="length_prob_exc")),
+                        ],
+                        id="length_dbct",
+                        label="Season Length",
                     ),
                 ],
                 className="mt-4",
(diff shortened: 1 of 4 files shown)


enacts/onset/layout.py
```diff
@@ -426,6 +426,14 @@ def results_layout():
                         id="cess_dbct",
                         label="Cessation Date",
                     ),
+                    dbc.Tab(
+                        [
+                            dbc.Spinner(dcc.Graph(id="length_plot")),
+                            dbc.Spinner(dcc.Graph(id="length_prob_exc")),
+                        ],
+                        id="length_dbct",
```
aaron-kaplan, line 434: What does dbct stand for?

enacts/onset/maproom.py
```diff
@@ -599,6 +599,143 @@ def cess_plots(
...
+    if not CONFIG["ison_cess_date_hist"]:
+        tab_style = {"display": "none"}
+        return {}, {}, tab_style
+    else:
+        tab_style = {}
+        lat = marker_pos[0]
+        lng = marker_pos[1]
+        try:
+            precip = pingrid.sel_snap(rr_mrg.precip, lat, lng)
+            isnan = np.isnan(precip).sum().sum()
```
aaron-kaplan, line 648: Why sum twice?

aaron-kaplan (approved):

enacts/onset/maproom.py
remicousin: I feel that question has been asked before. I feel like the answer could be something about doing a squeeze but for some reason it's easier/better this way; and so that it helps doing more straightforwarly the isnan > 0 test that follows. But I could be totally wrong.
This is also legacy code from cess_ case and I had the feeling it originally came from Kevin but I actually found it in this [PR](https://github.com/iridl/python-maprooms/pull/9) (see Oct 19th comment) from @drewmresnick even though this discussion doesn't argue specifically about using sum twice.

aaron-kaplan: Yes, I remember asking, but I don't remember getting a satisfactory answer. Could you just try deleting one of them and see what happens?

remicousin: It doesn't seem to make a difference and the last commit removes it. And we didn't keep the double sum in the flex forecast maproom for same purpose.

However, the unchanged behavior reveals unsatisfactory: if I make choices that make it impossible to find an onset date ever, I get the previous graph (ie graph remains unchanged) instead of an error graph... We are only testing whether there is precip data in the first place.

It looks like we need a test where data to plot is all missing, unless it's some dash option to trigger so that it doesn't keep the old graph.

aaron-kaplan: Indeed, the KeyError in line 653 and the TypeError in line 672 and 688 should all be generalized to Exception. We should always display an error message rather than continuing to display whatever happened to be there previously.

I think the reason we used specific exception types is because the error messages are specific. The easiest fix is to make the error messages less specific ("Something went wrong" or "Error" instead of "Grid box out of data domain"). A more helpful version could look something like
```
except Exception as e:
    if isinstance(e, KeyError):
        msg = "Grid box out of data domain"
    else:
        msg = "Internal error"
    error_fig = pingrid.error_fig(error_msg=msg)
    ....
```

remicousin: Alright, I left the checks on location clicks alone and added layers of checks to return an error fig if what feeds the graph is all nan.

aaron-kaplan:
Remember to run the tests. Last week I tried them on one of the open branches and they were broken. I don't remember if it was this branch.

PR #298, opened 2023-03-21
total rain between onset and cessation, multiplying size of problem b…

remicousin:
…y munber of years

### DIFF
--- a/enacts/onset/maproom.py
+++ b/enacts/onset/maproom.py
@@ -872,15 +872,13 @@ def tot_plots(
             cess_delta = cess_delta.isel({"T": slice(1, None)})
         if cess_delta["T"].size != onset_delta["T"].size:
             onset_delta = onset_delta.isel({"T": slice(None, -1)})
-        toto = np.stack([
-            (onset_delta["T"] + onset_delta["onset_delta"]).data,
-            (cess_delta["T"] + cess_delta["cess_delta"]).data,
-        ])
-        toto = toto[~np.isnan(toto).any(axis=1)]
-        seasonal_edges = np.stack([
-            (onset_delta["T"] + onset_delta["onset_delta"]).data,
-            (cess_delta["T"] + cess_delta["cess_delta"]).data,
-        ]).flatten("F")
+        seasonal_edges = xr.concat(
+            [
+                (onset_delta["T"] + onset_delta["onset_delta"]).drop_vars("T"),
+                (cess_delta["T"] + cess_delta["cess_delta"]).drop_vars("T"),
+            ],
+            dim="edges"
+        ).where(lambda x: ~np.isnan(x).any(axis=0), drop=True).data.flatten("F")
         try:
             # Total seasonal rainfall between onset and cessation dates is
             # every other sum over groups defined by onset and cessation dates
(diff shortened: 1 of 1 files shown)


remicousin:
Please hold: I clicked you as reviewer by mistake this morning. This is work in progress... sorry.

enacts/onset/maproom.py
```diff
@@ -768,6 +769,174 @@ def length_plots(
...
+            dim="edges"
+        ).where(lambda x: ~np.isnan(x).any(axis=0), drop=True).data.flatten("F")
+        try:
+            # Total seasonal rainfall between onset and cessation dates is
+            # every other sum over groups defined by onset and cessation dates
+            # as bins edges
+            seasonal_total = precip.groupby_bins(
+                "T",
+                seasonal_edges,
+                include_lowest=True
```
remicousin, line 889: One thing to double check is whether the inclusion is indeed inclusive for both edges. That's what I would want the way it is currently written.

aaron-kaplan:
Memory usage is excessive. I'm working on bringing it down.

remicousin:
> Memory usage is excessive. I'm working on bringing it down.

3 remarks:

1/ This soil moisture, calculated by water_balance, is the same always, at least in this maproom set up. So we could calculate it once and for all, store it in zarr format and read that to feed into the cessation date calculation

2/ This water_balance function was written quickly to satisfy the need to deliver this maproom. It's a special case of agronomy.soil_plant_water_balance but I don't know if soil_plant_water_balance has a more cost effective implementation.

3/ when/if the soil moisture, one day, is required to be computed live for some reason, we could probably be a bit more subtle and compute it only for a period of time allowing to do the seasonal search for the cessation date, as opposed to computing soil_moisture for all times regardless as it's done now.

aaron-kaplan:
I think I see the way forward for computing functions where the value at each time step depends on the previous time steps. We should use [DataArray.reduce](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.reduce.html). Its first argument should be a function `func` that takes two arguments: a DataArray that represents the result of the cumulative calculation for steps 0..i, and a DataArray that represents the input data for step i+1.  Writing it this way will have two advantages: (1) it avoids storing intermediate results longer than necessary, which is currently causing memory problems, and (2) although it won't immediately make the calculation faster, it opens the possibility of someday rewriting `func` in C, which would make it faster with relatively low effort.

This entails rewriting some of the existing code from previous PRs. I think we should rewrite everything in this paradigm: water balance, onset date, cessation date, and seasonal total. Write each one as a function that takes the result of the calculation from previous steps and the rainfall for the current step, and returns 
the result of the calculation for the current step. The seasonal total function will call the onset and cessation date functions, which in turn will each call the water balance function. (I guess that duplicates the water balance calculation; we can probably figure out a way to avoid that.)

aaron-kaplan:
BTW precalculating water balance is also a good idea, but then the zarrification/precalculation script should use `reduce`.

aaron-kaplan:
> This water_balance function was written quickly to satisfy the need to deliver this maproom. It's a special case of agronomy.soil_plant_water_balance but I don't know if soil_plant_water_balance has a more cost effective implementation.

`soil_plant_water_balance` uses a function called `soil_plant_water_step`, which is the type of `func` that `reduce` expects. But then it stores the result of the soil moisture calculation for all times (variable `sm`), whereas I'm suggesting that we should store the soil moisture for time t only long enough to calculate it for t+1.

aaron-kaplan:
@remicousin where do we stand on this? Is the ball in your court or mine?

remicousin:
That one is on hold until we finish rewriting the water balance / cessation to save or not to save all time steps, which is in progress in that PR #351 which is on you

remicousin:
But if anything I'd rather you start with PR #280 because Tufa would like to demo it in 2 weeks, whereas this updated onset or the water balance work, it's less clear when they want them finished.

PR #301, opened 2023-03-28
removed one too many tailing points

remicousin:
This is another error that doesn't have to do with the bug where a previous line plot persists after a callback. Let's merge this to clear the path.

### DIFF
--- a/enacts/onset/maproom.py
+++ b/enacts/onset/maproom.py
@@ -711,7 +711,7 @@ def length_plots(
         if cess_delta["T"][0] < onset_delta["T"][0]:
             cess_delta = cess_delta.isel({"T": slice(1, None)})
             if cess_delta["T"].size != onset_delta["T"].size:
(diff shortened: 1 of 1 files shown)


enacts/onset/maproom.py
```diff
@@ -708,7 +708,7 @@ def length_plots(
         if cess_delta["T"][0] < onset_delta["T"][0]:
             cess_delta = cess_delta.isel({"T": slice(1, None)})
             if cess_delta["T"].size != onset_delta["T"].size:
-                onset_delta = onset_delta.isel({"T": slice(None, -2)})
+                onset_delta = onset_delta.isel({"T": slice(None, -1)})
```
aaron-kaplan, line 711: Why would they not be the same length?

aaron-kaplan: Resolved in person. Ok for this change, but line 710-711 need to be outdented.

PR #302, opened 2023-03-28
applying dropna resolves the bug but connects years that are not neig…

remicousin:
…hbors

The best way to see the bug and its "solution" is to run the app and change the default onset search month from November to September. Then:

dropna is applied to onset date graph and the persisting graph is not happening anymore. However we don't want to connect years that are not neighbor.

dropna is not applied to cessation but cessation doesn't have na so it behaves.

dropna is not applied to length and we can see the persisting graph from before the callback.

I found this [conversation](https://community.plotly.com/t/plotly-dash-graph-showing-some-elements-of-previous-figure-after-replacing-the-figure-through-dash-callback/58653) about it that led me to try getting rid of na.

I will have another PR use another kind of graphics so that we circumvent the bug for now.

### DIFF
--- a/enacts/onset/maproom.py
+++ b/enacts/onset/maproom.py
@@ -447,6 +447,7 @@ def onset_plots(
             error_fig,
             germ_sentence
         )  # dash.no_update to leave the plat as-is and not show no data display
+    onset_delta = onset_delta.dropna(dim="T")
     onset_date_graph = pgo.Figure()
     onset_date_graph.add_trace(
         pgo.Scatter(


aaron-kaplan:
As we've discussed before, a line graph really isn't appropriate here, because drawing a line graph from discrete points involves interpolation between the points, and it's nonsensical to say that the onset in year 1991.5 is halfway between the 1991 onset and the 1992 onset.

Plotly's current behavior in the presence of NaNs is clearly incorrect, but it's not clear to me that Ingrid's (or was it gnuplot's?) behavior (leaving blanks where there's no value) is the right way to handle it either. If interpolation is allowed between regularly spaced points on a continuum, why shouldn't it be allowed across irregularly spaced points too?

Switching to a bar or scatter plot is not just a workaround, it's also the right way to plot this variable. Given that, I'm not sure it's worth my time to figure out the plotly bug.

remicousin:
It's fine with me to move away from a line plot.

But it also means we can't make line plots unless we are certain there is no missing value in the series to plot.

remicousin:
The problem with scatter only is that it's harder to see the time evolution... That's why it makes sense to connect with lines even though the values between markers are meaningless... But I'll see if I find another visual that would do better.

Screenshot 2023-03-30 at 1 40 17 PM

remicousin:
A color gradient helps to see the time sequence better but one might wonder what the color is for while is just for x-axis and to help the reading....

Screenshot 2023-03-30 at 1 49 28 PM

PR #315, opened 2023-04-05
dynamic prob of non-/exc color scales

remicousin:
I might not use the ColorScale class as I should...


aaron-kaplan (commented):
It would be nice to write the logic just once, in a function that's called from both draw_colorbar and fcst_tiles, to avoid the risk of them getting out of sync in the future.

aaron-kaplan (approved):

PR #316, opened 2023-04-10
Remic/nmafcst

remicousin:
A working case of CPT forecast from NMA. Had to adjust a couple of things in ZMD.

### DIFF
--- a/enacts/config-nma.yaml
+++ b/enacts/config-nma.yaml
@@ -1,6 +1,7 @@
 # Ethiopia
 logo: Ethiopia_IRI_98x48.png
 institution: NMA
+map_center: [9., 40.]
 zoom: 5
 shapes_adm:
     - name: National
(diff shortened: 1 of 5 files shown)


remicousin:
Also had to set up the Ethiopia monthly Maproom so I can build the multipage (hence build the forecast -- that's why all this prior work to cleaning up zarrification) and found a couple unrelated bugs.

remicousin:
> > Had to adjust a couple of things in ZMD.
> 
> I don't see changes to the ZMD config in this PR. Did you forget to commit them?

sorry I wasn't very clear, the adjustment needed so that this works for both ZMD and NMA is the series of changes from:
CONFIG["targets"][1]
to
CONFIG["targets"][-1]

It was 1 because (so 2nd in the list) because of the unexpected case that ZMD didn't have files for the first lead of the last start (would have wanted to use 0) which would typically be the default forecast we'd want to show landing on the page. Bur NMA has only 1 lead for now so so can't be 1... So I opted for -1... I was considering making it a config entry but also was thinking that we are likely to overhaul or at least clean up how we scroll through the CPT files, get start, lead, target and read files... So I went for the simplest.

Open to other choice.

aaron-kaplan (approved):

PR #319, opened 2023-04-17
adjusting to new config and colorscale settings

remicousin:
I had not done adjusted this app to the new config system at the time (because I was in a hurry likely) and now there are also the new colorscales to adjust.

For Aaron to review and Drew to be aware as this may bare on how they set up their Crop Maproom.

### DIFF
--- a/enacts/wat_bal/config-defaults.yaml
+++ b/enacts/wat_bal/config-defaults.yaml
@@ -1,0 +1,1 @@
+../config-defaults.yaml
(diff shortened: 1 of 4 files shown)


enacts/wat_bal/config-defaults.yaml
```diff
@@ -0,0 +1 @@
+../config-defaults.yaml
```
aaron-kaplan, line 1: Do we really need these symlinks? Can't you just use `CONFIG=../config-defaults.yaml:...`?

aaron-kaplan (approved):

PR #321, opened 2023-04-19
taw read from file depends on X/Y instead of constant

remicousin:
I ran into a couple of issues in order to this and what I propose here is probably the start of the discussion rather than the final solution. I am also going to add some comments to point out some specific problems that led me to do some changed.

rr_mrg's and taw's X/Y are the same by design but they respectively are double and single precision. That caused the tiling to create tiles of different sizes (by 1 point), resulting in water balance algo to fail (see more details below -- but water balance algo expects coordinates of same name to be identical througout its inputs).

converting rr_mrg's X/Y to single fixed the tiling problem, but water balance was still failing. Turns out that one of the two sets of X/Ys had many values with tailing ...00001. Yet, there are a series of broadcast used in water balance algo to set the stage for the computation, and so as a result, say latitudes y and y+0.000001 were both broadcasted, leading to wrong objects, of differing sizes not allowing the water balance algo to go through.

Inspired by this [conversation](https://github.com/pydata/xarray/issues/2217), I resorted to override-align taw against rr_mrg. This made the initial precision conversion useless and so I commented out to keep it just for the story-telling here for now.

### DIFF
--- a/enacts/wat_bal/maproom_monit.py
+++ b/enacts/wat_bal/maproom_monit.py
@@ -44,6 +44,16 @@ if DATA_PATH is None:
 DR_PATH = f"{GLOBAL_CONFIG['daily']['zarr_path']}{DATA_PATH}"
 RR_MRG_ZARR = Path(DR_PATH)
 rr_mrg = calc.read_zarr_data(RR_MRG_ZARR)
+#rr_mrg["X"] = rr_mrg["X"].astype(np.single)
+#rr_mrg["Y"] = rr_mrg["Y"].astype(np.single)
+
+kaka, taw = xr.align(
+    rr_mrg,
+    xr.open_dataarray(Path(CONFIG["taw_file"])),
+    join="override",
+    exclude="T",
+)
+taw_max = taw.max()
 
 # Assumes that grid spacing is regular and cells are square. When we
 # generalize this, don't make those assumptions.
@@ -364,12 +374,13 @@ def wat_bal_plots(
         error_fig = pingrid.error_fig(error_msg="Grid box out of data domain")
         return error_fig
     precip.load()
+    taw = pingrid.sel_snap(xr.open_dataarray(Path(CONFIG["taw_file"])), lat, lng)
     try:
         sm, drainage, et_crop, et_crop_red, planting_date = ag.soil_plant_water_balance(
             precip,
             et=5,
-            taw=60,
-            sminit=60./3.,
+            taw=taw,
+            sminit=taw/3.,
             kc_params=kc_params,
             planting_date=p_d,
         )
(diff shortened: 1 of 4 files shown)


enacts/wat_bal/maproom_monit.py
```diff
@@ -364,12 +374,13 @@ def wat_bal_plots(
         error_fig = pingrid.error_fig(error_msg="Grid box out of data domain")
         return error_fig
     precip.load()
+    taw = pingrid.sel_snap(xr.open_dataarray(Path(CONFIG["taw_file"])), lat, lng)
```
remicousin, line 377: I was not able to use taw here however defined earlier so had to read the file again. I am told that I reference taw before defining it.
The alignment is not necessary because it's only to fix precisions problems that are irrelevant to sel_snap here.
But it doesn't feel great to read file again.... does it?

enacts/wat_bal/maproom_monit.py
```diff
@@ -459,14 +470,19 @@ def wat_bal_tile(tz, tx, ty):
     )
 
     mymap_min = 0
-    mymap_max = 60 #taw.max()
+    mymap_max = 8* (int(taw_max.values) // 8)
```
remicousin, line 473: I had to compute taw_max at the top of the file otherwise I would be returned an error that a resource is already in use. That's probably all fine but I'd figure I'd mention that here as that goes along the lines of the things I don't quite understand in this PR.

enacts/wat_bal/maproom_monit.py
aaron-kaplan: On the contrary, as a rule we *should* read the data from disk each time we need it.  Once there are more than a few maprooms, we won't be able to keep all our datasets in memory at all times.

It's ok to hold data in memory for the duration of a callback, but once the callback finishes we should release the memory to free it up for handling other callbacks.

enacts/wat_bal/maproom_monit.py
```diff
@@ -44,6 +44,16 @@
...
 rr_mrg = calc.read_zarr_data(RR_MRG_ZARR)
+#rr_mrg["X"] = rr_mrg["X"].astype(np.single)
+#rr_mrg["Y"] = rr_mrg["Y"].astype(np.single)
+
+kaka, taw = xr.align(
+    rr_mrg,
+    xr.open_dataarray(Path(CONFIG["taw_file"])),
+    join="override",
+    exclude="T",
+)
```
aaron-kaplan, line 55: No point calling `align` here anymore, now that the only thing we use this copy of the data for is to calculate the max.

aaron-kaplan (approved):
This is fine. I don't have a precise explanation for the errors you described, but they both have to do with concurrent access to a global resource by multiple concurrent threads or processes. It might be between python threads (though I don't think so because `maproom_monit.py` sets `threaded=False`) or between dask workers. In any case, eliminating the global variable `taw` is the right solution.

enacts/wat_bal/maproom_monit.py
remicousin: ok. Let's talk about this some day in the office. I don't think I get how the data requests flow.

enacts/wat_bal/maproom_monit.py
remicousin: Per my comment on that discussion you pointed to, then it would make sense that I put this taw_max hard-coded in config file.

enacts/wat_bal/maproom_monit.py
remicousin: I use it in the tile function see line 476. But then, if I put taw_max hard-coded in config file, and if file reading should rather be within callbacks, then I could put that reading and alignment in the tile function.

remicousin: one-to-last commit reorganizes according to our discussion.

enacts/wat_bal/maproom_monit.py
```diff
@@ -470,10 +460,16 @@ def wat_bal_tile(tz, tx, ty):
     )
 
     mymap_min = 0
-    mymap_max = 8* (int(taw_max.values) // 8)
+    mymap_max = CONFIG["taw_max"]
     mycolormap = CMAPS["precip"]
 
-    taw_tile = taw.sel(
+    garbage, taw_tile = xr.align(
```
aaron-kaplan, line 462: It's conventional to use the name `_` (underscore) for a return value you don't intend to use.

aaron-kaplan (approved):

PR #325, opened 2023-04-21
FBF: Fix vuln colorscale--opencv expects bgr not rgb.

aaron-kaplan:
I guess I forgot to test the vulnerability layer before approving the colormap change. It's currently displaying shades of blue rather than shades of yellow/orange.

### DIFF
--- a/fbfmaproom/__about__.py
+++ b/fbfmaproom/__about__.py
@@ -4,4 +4,4 @@
 name = "fbfmaproom"
 author = "IRI, Columbia University"
 email = "help@iri.columbia.edu"
-version = "2.45.0"
+version = "2.45.1"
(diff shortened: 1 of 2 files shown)


remicousin (approved):
I haven't interfered with this apply_mask function so I assume you know what it is up to.

aaron-kaplan:
You did change it here: https://github.com/iridl/python-maprooms/pull/246/files#diff-7e36d4e5a9d8a7b6255214ee91840aa2ece17c138a346dc1656535d142675c1fR563

That change replaces a BGRA with a Color (which is RGB). That value gets converted to a numpy array and passed to opencv, which expects BGRA.

aaron-kaplan:
This may not be correct, or complete. Now I get yellows instead of blues, but the map still looks quite different from the original. I have a feeling the scale may be inverted. I'll have to look into it more on Monday.

remicousin:
I will try to as well. Unlikely today.

remicousin:
The thing is that I never knew how to make the Vulnerability map appear so on that front, my "testing" was only on the colorbar...

aaron-kaplan:
I tried the same confirmation again today and what I see from this branch looks identical to what's in production. I suspect I accidentally selected different years when I attempted to compare last week.

PR #328, opened 2023-04-26
CI

aaron-kaplan:
Runs (some of) the fbfmaproom tests in CI. ~Ready to review, but not to merge until after the other PRs that fix the broken tests.~

This version runs the tests on the CI VM, uncontainerized. I want to switch to running everything with `docker run` because that will let us share a build/test workflow between local and GitHub, and reduce the lockin to GitHub, but it's going to take some work to get that to be fast so I'm going with this version for now.

### DIFF
--- a/.github/workflows/fbf-test.yaml
+++ b/.github/workflows/fbf-test.yaml
@@ -1,0 +1,70 @@
+on:
+  pull_request:
+  workflow_dispatch:
+env:
+  CONDA_TYPE: Mambaforge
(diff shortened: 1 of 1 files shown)


aaron-kaplan:
@kgraaf I've rebased after merging the PRs that fixed the tests, and the tests passed in CI. Please review.

kgraaf:
I know very little about this sort of thing, but, this looks like a cogent initial set of tests to me

kgraaf (approved):

PR #332, opened 2023-05-03
Speed up table generation by avoiding redundant shape queries

aaron-kaplan:
test_table_cb goes from 4.5s to 1.5s.


kgraaf:
Looks great

kgraaf (approved):

kgraaf:
Looks great

Issue #251, opened 2022-12-16
Better controls of elements sizes in Blocks


remicousin:
> Make the input boxes narrower so we can fit more than one of them per line?
> 
> ![image](https://user-images.githubusercontent.com/766406/207972491-743fa68d-926a-4925-ad3e-a05e3d0dbb89.png)

They are much smaller on my laptop. I am not entirely sure how that is controled at present. Can we make this an issue at least until widget/component/controls are one?

<img width="413" alt="Screen Shot 2022-12-16 at 3 05 03 PM" src="https://user-images.githubusercontent.com/42619415/208179879-242780f1-b53e-481e-8861-db1f7b384214.png">

_Originally posted by @remicousin in https://github.com/iridl/python-maprooms/issues/236#issuecomment-1355527711_

remicousin:
Generally speaking it would be nice, in my opinion, to have more compact controls that take less (white) space

aaron-kaplan:
This is a browser difference. I usually use Chrome; when I try firefox I see the same thing as you.

I looked up the [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text#size) on the default width of HTML inputs. Apparently the default is "20 characters wide", but that's underspecified when using a variable-width font, so different browser authors make different choices.

I tried fiddling with CSS styles but didn't manage to make them smaller in chrome. Will keep working on it.

remicousin:
It's not using the default since I introduced the html-size argument for controls' Number. But I see that the doc says it's not reliable...

aaron-kaplan:
Hmm. `Number` is being used with `html_size` but it's not taking effect, while `DateNoYear` and `Text` don't accept `html_size` so they're using the default size.

remicousin:
All the above might become irrelevant based on findings in Issue #472

PR #333, opened 2023-05-03
Adding the Compare to... time series options

remicousin:
This adds a 2nd time series to the local graph that puts the monitored one in context. Controls allow to compare different things (another planting date, another year, another crop).

There are a couple of problems.

I commented out the assertion in controls.py because it complained even though I really can't tell why it behaves differently from all the other Sentence objects. I printed that it believes the Number object for year_ago is a list instead of what it is asserting. There are Numbers in other Sentences... I don't get it...

The other problem is that the improper behavior of printing additional stuff first mentioned in #302  (and not addressed because we moved then to bar plots for which it's irrelevant) has returned. I get it each time I change the planting date month of the Compare to... ts from default June to May.


enacts/wat_bal/layout_monit.py
```diff
@@ -311,6 +328,50 @@ def controls_layout(lat_min, lat_max, lon_min, lon_max, lat_label, lon_label):
...
                         dbc.Button(id="submit_kc", children='Submit'),
                     ),
+                    Block(
+                        "Compare to...",
+                        Sentence(
+                            "Planting Date",
+                            DateNoYear("planting2_", 1, CONFIG["planting_month"]),
+                        ),
+                        Sentence(
+                            Number("year_ago", 1, min=0, max=99, html_size=3),
```
aaron-kaplan, line 338: Wouldn't it be less confusing to name the years (1998, 1999, ...) instead of counting how far back they are?

aaron-kaplan (commented):
I would prefer to see the code that's the same for the two plots factored out into a function or a loop rather than copy-pasted. Copy-paste mistakes are easy to make and often hard to notice.

I wonder if the missing controls I commented on below will fix the assertion. If not, let me know and I will investigate.

This plotly [issue](https://github.com/plotly/plotly.js/issues/6407) sounds similar to the problem you saw. It was [fixed](https://github.com/plotly/plotly.js/blob/master/CHANGELOG.md#2200----2023-03-15) in plotly.js 2.17.1, which [made it](https://github.com/plotly/plotly.py/blob/master/CHANGELOG.md) into plotly.py in version 5.12. We currently have plotly 5.11. Please try upgrading plotly (and everything else) by following the section "Adding or removing dependencies" in the README. (While you're at it, you could change the title to "Adding, removing, or upgrading dependencies.") Then create a new environment as described in the section "Creating a conda environment with this project's dependencies" (add that pointer to the "Adding..." section too), and test.

enacts/wat_bal/layout_monit.py
remicousin: you mean confusing for the programmer or for the user? I agree for the user it would be likely easier to enter day, month and year.

Either way, the difficulty with having the year as input is the realignment on the same year for plotting purposes...

Let's say the planting dates are 15 Apr 2023 and 3 Apr 2012, then it's easy to rewrite the ts starting in 3 Apr 2012 to start in 3 Apr 2023. But now let's say we have 3 Jan 2023 and 28 Dec 2012. How do I systematically make sure that I re-align on 28 Dec 2022?  I could take all the 28 Dec, take the difference with 3 Jan 2023, and pick the smallest.

Other idea?

enacts/wat_bal/layout_monit.py
```diff
@@ -311,6 +328,50 @@ def controls_layout(lat_min, lat_max, lon_min, lon_max, lat_label, lon_label):
                         ),
                         dbc.Button(id="submit_kc", children='Submit'),
                     ),
+                    Block(
+                        "Compare to...",
+                        Sentence(
+                            "Planting Date",
+                            DateNoYear("planting2_", 1, CONFIG["planting_month"]),
```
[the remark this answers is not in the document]
remicousin: nope (you already asked that one :) )

DateNoYear creates 2 controls with each its id which are going to be yadiyada_day and yadiyada_month.

remicousin:
> I would prefer to see the code that's the same for the two plots factored out into a function or a loop rather than copy-pasted. Copy-paste mistakes are easy to make and often hard to notice.
> 
> I wonder if the missing controls I commented on below will fix the assertion. If not, let me know and I will investigate.
> 
> This plotly [issue](https://github.com/plotly/plotly.js/issues/6407) sounds similar to the problem you saw. It was [fixed](https://github.com/plotly/plotly.js/blob/master/CHANGELOG.md#2200----2023-03-15) in plotly.js 2.17.1, which [made it](https://github.com/plotly/plotly.py/blob/master/CHANGELOG.md) into plotly.py in version 5.12. We currently have plotly 5.11. Please try upgrading plotly (and everything else) by following the section "Adding or removing dependencies" in the README. (While you're at it, you could change the title to "Adding, removing, or upgrading dependencies.") Then create a new environment as described in the section "Creating a conda environment with this project's dependencies" (add that pointer to the "Adding..." section too), and test.

While a new conda-lock is building... I have a couple questions.

So practically, code could potentially be factored in through the pgo.Scatter objects. Whether I go that far, or not, the question I have is how to deal with the trys/excepts that potentially return in the middle of the code an errorfig? Should the error checkings be inside the factored in function and the function returns an object or an errorfig (it would be weird that the function doesn't return objects of same nature always); or the function returns an object or an error and then follow up code returns the errorfig or keeps going.
I am sure there is best practice for this but I just don't know what it is.

See per my comment that there is no missing control/id. And the code fails with the Number object that comes after the DataNoYear object anyways...

enacts/wat_bal/layout_monit.py
aaron-kaplan: I meant for the user. Leave the value as integer number of years, but display it as "1 year ago (28 Dec 2022)"?

remicousin:
Also... how to have italic or bold or what-have-you in the text in our f""" """ strings? When mentioning the name of something in the page like the control Compare to..., I would like to write it _Compare to..._
I read a bit about it yesterday but got lost through many pages with examples using f" " or markdown with """ """ (and never f""" """) and I figured I would ask in case it's trivial to you.

enacts/wat_bal/layout_monit.py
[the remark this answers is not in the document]
aaron-kaplan: Sounds vaguely familiar. I don't like this. It seems important to be able to grep for an input name.

aaron-kaplan:
> Should the error checkings be inside the factored in function and the function returns an object or an errorfig (it would be weird that the function doesn't return objects of same nature always); or the function returns an object or an error and then follow up code returns the errorfig or keeps going.

I would keep display concerns (anything related to plotly) out of the calcuation function. Have it return a DataArray (?) on success; on failure, it could either return a sentinel value (e.g. `None`), or raise an exception that you catch in the calling function.

I don't know what difficulty you're having with italics. dcc.Markdown looks like an easy way to get it. " ", """ """, f" ", and f""" """ are all just ways of building a string. You can put markdown in any of them. Want to show me a particular example that didn't work?

aaron-kaplan:
After changing the month selector from June to May, I don't think I'm seeing the strange behavior with either version of the conda environment. Could you attach a screenshot? Here's mine.

image

aaron-kaplan:
BTW I'm confused by the multiple submit buttons that all seem to do the same thing.

aaron-kaplan:
I also uncommented the assertion you had commented out, and that assertion isn't failing either. Are you sure you don't have any uncommitted changes?

enacts/wat_bal/layout_monit.py
remicousin: The user now inputs a full planting date (d-m-y) to compare to.

remicousin:
> BTW I'm confused by the multiple submit buttons that all seem to do the same thing.

what is the confusion about more specifically?

remicousin:
> I also uncommented the assertion you had commented out, and that assertion isn't failing either. Are you sure you don't have any uncommitted changes?

I do still get assertion errors... Actually had to comment one more...

aaron-kaplan:
Got it. That one does fail for me. The previous one, in line 151, which you
also commented out, isn't failing (possibly not even being hit--I haven't
tried to understand the flow.).

controls.py
```diff
@@ -148,12 +148,12 @@ def Sentence(*elems):
...
 
     for i in range(start, len(elems) - (1 if tail else 0), 2):
-        assert (isinstance(elems[i], str) or isinstance(elems[i], html.Span))
+        #assert (isinstance(elems[i], str) or isinstance(elems[i], html.Span))
         groups.append(dbc.Label(elems[i], size="sm", className="m-1 d-inline-block", width="auto"))
         groups.extend(elems[i + 1])
 
     if tail:
-        assert (isinstance(elems[-1], str) or isinstance(elems[-1], html.Span))
+        #assert (isinstance(elems[-1], str) or isinstance(elems[-1], html.Span))
```
aaron-kaplan, line 156: This assertion says that we're not allowed to have two controls in a row, controls have to alternate with text; but the `DateNoYear` function generates two controls in a row: day of month and month. I'm not sure why this constraint was imposed, but it seems to have something to do with how groups are formed, which I'm not entirely following. @kgraaf please comment.

remicousin: More specifically, the problem was that DateNoYear was followed by Number. I interjected a "" inbetween the 2 and re-enabled the assertions and going through.

aaron-kaplan:
I think the only thing still unresolved here is the issue of the multiple submit buttons, which I find problematic.
- The buttons are in a box that scrolls, and they're never all visible at once, so unless you're already familiar with the interface you won't know that some of them are there. Sometimes the only button I can see is one above the control I'm changing, which doesn't submit the value of that control, but that's not obvious. The one that does submit the value of that control is below, scrolled out of sight.
- All three buttons cause the graph to be regenerated, so it's not clear why we need three of them. I guess one changes the map and the chart while two only change the chart, so maybe there's a case to be made for having two, but even that might be more confusion than it's worth.

remicousin:
A few comments of different nature about this:

1/ You clearly haven't read the text describing the Maproom. It's also in a scrolling box and so it's also hard to see that there is more to read. Maybe we should change scrolling boxes with collapsable content?

2/ Regardless, a typical user will jump to the controls and not bother read text/instructions until they realize they don't understand what's going on, so the criticism remains valid.

3/ I personally feel those controls take too much space:
3a/ I would put the smaller ones (Pick a point, variable) in the top Control Bar (along with date) in a smaller shape
3b/ the bigger ones that make sense to put all together (Current Season and Compare to) are probably better where they are rather than in the top Control Bar, but I would work on their looks to be more compact and have a lot less white space. Hopefully that would require less of scrolling and possibly even give more room to text above.

4/ I believe those controls/buttons are in individual "Cards" objects that could have bolder or color coded frames so that they can be seen better.

Opinions?

aaron-kaplan:
@kgraaf should get involved in this discussion since it has implications for what UI layout options we want in maprooms in general. (See Rémi's comment preceding this one.)

@remicousin do you want to move ahead with this version and improve the UI in subsequent PRs?

aaron-kaplan:
My problem with the submit buttons is related to something I've mentioned before: for some reason the input boxes are much wider for me than they are for you, and as a result the form takes much more vertical space. Fixing that would probably make this a non-issue. See screenshot below.

Move ahead and address that later?

image

enacts/wat_bal/layout_monit.py
```diff
@@ -309,7 +358,57 @@ def controls_layout(lat_min, lat_max, lon_min, lon_max, lat_label, lon_label):
...
+                        Sentence(
+                            Number("kc2_late", CONFIG["kc_v"][3], min=0, max=2, html_size=4),
+                            "through",
+                            Number("kc2_late_length", CONFIG["kc_l"][3], min=0, max=99, html_size=2),
+                            "days of late-season to",
+                        ),
+                        Sentence(
+                            Number("kc2_end", CONFIG["kc_v"][4], min=0, max=2, html_size=4),
+                        ),
+                        dbc.Button(id="submit_kc2", children='Submit', color="primary"),
```
aaron-kaplan, line 410: Using a color in the form that corresponds to the color of the graph line affected by that form is a nice touch, but if you use "blue" and "green" for the  lines and "primary" and "secondary" for the buttons then they're likely to get out of sync in the future. Different bootstrap themes use different button colors.

The dash-bootstrap docs don't say that we can use a color name like "blue" here, but give it a try, it might work anyway. If it doesn't, I would just let it use the default color.

remicousin:
> My problem with the submit buttons is related to something I've mentioned before: for some reason the input boxes are much wider for me than they are for you, and as a result the form takes much more vertical space. Fixing that would probably make this a non-issue. See screenshot below.
> 
> Move ahead and address that later?
> 
> image

It's funny that the behavior differ on your system vs mine... But anyway indeed we already have an issue #251 for this.

enacts/wat_bal/layout_monit.py
remicousin: and in case you're asking, the Card's outline=True option doesn't do s..t (that would not have fixed the fact that button doesn't know colors, but that would have made it obsolete because the whole content of the Card would have colored borders...)

enacts/wat_bal/layout_monit.py
```diff
@@ -309,7 +358,67 @@ def controls_layout(lat_min, lat_max, lon_min, lon_max, lat_label, lon_label):
...
+                                other_year_default,
+                                min=other_year_min,
+                                max=other_year_max,
+                                html_size=5
+                            ),
+                        ),
+                        Sentence(
+                            "for",
+                            Text("crop2_name", CONFIG["crop_name"]),
+                            "crop cultivars: initiated at",
```
aaron-kaplan, line 386: This and the corresponding input in the previous Block should be a Select (choose from a predefined set of options) rather than an unrestricted Text input?

remicousin: Not at this point. The crop name text is mostly for cosmetics. There may be another version alter when the user would actually pick a crop name in a list, and that would set Kc values (and inform the user what those values are). But in this version, the user picks the "crop" by setting the Kc values themselves, and they can give it a name for convenience, cosmetics.

aaron-kaplan: Ah. I tried typing in "ffgg" as the crop name, and when it didn't result in an error message, I concluded that something was wrong with the error handling. It never occurred to me that was meant to be an unconstrained free-text field.

I'll allow that I'm not typical of the users of this application, but I find this confusing/surprising.

remicousin: I'll come up with something else in the next PR.

PR #335, opened 2023-05-09
Remic/steps

remicousin:
Rewriting "dummy" water balance to allow outputing a time series or the last point.

If it eases your review, the history of commits make sense.

The Onset Maproom is still working fine with this without touching it. There is nothing to add to it until the cess date is rewritten the same way so that it can take advantage of a reduced water balance run.

As mentioned, I didn't feel the need to use xr.reduce (but I missed the reason it was mentioned).

After this, either continue on with cess date function; or apply the methodology to the "real" water balance then move on to cess date.


enacts/calc.py
```diff
@@ -25,13 +25,37 @@ def read_zarr_data(zarr_path):
...
+    ).clip(min=0, max=taw).assign_coords(
+        {time_dim: [(sm_history[time_dim][-1] + np.timedelta64(1, "D")).values]}
+    )
+    if not reduce:
+        #at this point, this assumes that
+        #peffective, et and taw don't add new dimensions
+        #through broadcasting to sm_history
+        #so that the concat can happen
+        sm = xr.concat([sm_history, sm], time_dim)
+    return sm
```
aaron-kaplan, line 49: I would prefer to have the following simple function that just describes the calculation itself
```
def water_balance_step(sm_prev, peffective, et, taw):
    return (sm_prev + peffective - et).clip(min=0, max=taw)
```
and put the burden on the caller to manage the bookkeeping about what to do with the result (store it in a big array vs. just pass it to the next iteration).

I'm not sure that preference will be compatible with an eventual reimplementation in C/Cython, though. So if you don't agree with my aesthetic opinion, feel free to ignore it, at least for now.

Unrelated: in the case where we need to keep the entire history, pre-allocating the whole array and writing slices into it will probably be significantly faster than incrementally growing the array by calling `concat` on each iteration.

enacts/onset/tests/test_calc.py
```diff
@@ -169,13 +177,28 @@ def test_water_balance2():
...
+    precip = xr.DataArray(values, dims=["X", "T"], coords={"T": t})
+    wb = calc.water_balance(precip, 5, 60, 0, reduce=True)
     expected = [
-        [0.0, 1.0, 0.0, 60.0],
-        [5.0, 12.0, 21.0, 32.0],
+        [60.0],
+        [32.0],
     ]
-    assert np.array_equal(wb.soil_moisture, expected)
+    assert np.array_equal(wb.soil_moisture.transpose(), expected)
```
aaron-kaplan, line 201: What's going on with this `transpose`? Why not just write `expected` to be what you're actually expecting?

enacts/calc.py
remicousin: So rather something like what I've just pushed?

enacts/onset/tests/test_calc.py
remicousin: just being lazy

aaron-kaplan: If it's more convenient to write the transposed version, then you could do
```
expected = [ ... ].transpose()
```
Less confusing without being less lazy.

enacts/calc.py
remicousin: I'll do the pre-allocated array after we're good on where to put the burden of reduction decision.

aaron-kaplan: Yes, for my first comment. The comment about preallocating the array still stands.

remicousin: Before I do that, a comment that preallocating the array means I will have to manually broadcast everything to have the right sm array from the start (which resulted in most of the lines of code in the "good" water balance function).

whereas here, even though [xr.concat](https://docs.xarray.dev/en/stable/generated/xarray.concat.html) claims that

"Each object is expected to consist of variables and coordinates with matching shapes except for along the concatenated dimension."

sm_i is concatenated to soil_moisture in the first iteration in test_water_balance2() even though sminit has no X dimension but precip does... so apparently broadcasting things smoothly.

remicousin:
Or maybe I run the first iteration of water_balance_step so that it does the broadcasting, then I allocate with the proper T dim, then I loop from 1 to T.size?

enacts/calc.py
remicousin: Or maybe I run the first iteration of water_balance_step so that it does the broadcasting, then I allocate with the proper T dim, then I loop from 1 to T.size?

aaron-kaplan: Whatever seems easiest. If it's going to make the code harder to read, we should verify that it actually helps performance significantly. But you can't verify that without making the change first.

remicousin: how about that.

I don't know why but all the initialization and broadcasting problems are fading away...

enacts/calc.py
```diff
@@ -65,28 +74,30 @@ def water_balance(
...
+        soil_moisture, _ = xr.align(soil_moisture, daily_rain[time_dim], join="outer")
+    # Looping on time_dim
+    for t in daily_rain[time_dim][1:]:
+        sm_t = water_balance_step(
+            soil_moisture.sel({time_dim: t - np.timedelta64(1, "D")}, drop=True),
+            daily_rain.sel({time_dim: t}).expand_dims(dim=time_dim),
+            et,
+            taw,
+        )
+        soil_moisture = sm_t if reduce else xr.where(
```
aaron-kaplan, line 94: You're still allocating a new `soil_moisture` array on each iteration. `where` creates a new array and fills it with values copied from its arguments.

In contrast, the assignment in the old  (red) lines 80-82 modifies the existing `soil_moisture` array.

enacts/calc.py
```diff
@@ -65,28 +74,30 @@ def water_balance(
...
-    soil_moisture = xr.DataArray(
-        data=np.empty(daily_rain.shape),
-        dims=daily_rain.dims,
-        coords=daily_rain.coords,
-        name="soil_moisture",
-        attrs=dict(description="Soil Moisture", units="mm"),
+    # Initializing soil_moisture
+    soil_moisture = water_balance_step(
+        sminit,
+        daily_rain.isel({time_dim: 0}).expand_dims(dim=time_dim),
```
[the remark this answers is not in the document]
remicousin: yes because it conveniently assigns the right 1-point time dimension to first iteration of soil_moisture, which I either want to use with align if not reduce, or have the time dim information if reduce.

enacts/calc.py
remicousin: or make a more involved if here with:

if reduce:
    soil_moisture = sm_t
else:
    soil_moisture[t] = sm_t

?

aaron-kaplan: I would do the latter.

enacts/calc.py
```diff
@@ -65,28 +69,29 @@ def water_balance(
...
-    ).clip(0, taw)
-    # Looping on time_coord
-    for i in range(1, time_coord_size):
-        soil_moisture[{time_coord: i}] = (
-            soil_moisture.isel({time_coord: i - 1})
-            + delta_rain_et.isel({time_coord: i})
-        ).clip(0, taw)
-    water_balance = xr.Dataset().merge(soil_moisture)
+    if not reduce:
+        soil_moisture, _ = xr.align(soil_moisture, daily_rain[time_dim], join="outer")
```
aaron-kaplan, line 80: I guess the thing being assigned to `_` is `daily_rain` broadcast to the dimensions of `soil_moisture`. Seems like a waste to allocate that big array only to throw it away. Could this be done with `broadcast_like` instead?

aaron-kaplan: (Since this happens only once, it's not a big deal. )

aaron-kaplan (approved):

enacts/calc.py
remicousin: broadcast_like seems even more straightforward.

remicousin:
@aaron-kaplan , in terms of next steps... I could rewrite cess_date in loop form now so that I can nest in this water_balance reduce case, and we'll see what that does to the resource needs.

OR

I convert the complete water_balance to the new paradigme, then have it work for multiple years/seasons, then rewrite cess_date in loop form taking that version of water balance with reduction.

AND

regardless of the options above, shall we keep the vectorial version of cess_date? If so, shall we have the vectorial case applied when input is already sm, and the looping one when the input is "raw" rainfall? and if so, should the two cases be 2 different functions? or one function of which inputs deals with the possible variety of inputs and methods?

Any suggestions?

aaron-kaplan:
I think I would go for the first option because it's a smaller step. But are you asking because it's ultimately wasted work when we replace the water balance calculation with a more sophisticated one? (I don't have the big picture loaded in my head right now.) I think if the smaller step is useful in the interim then take it, because something is better than nothing and we never know what will come up and prevent us from making the larger change. If it's completely useless then skip it.

All else being equal I would rather maintain one version of cess_date instead of two. But if the performance difference between the vectorized one and the one with the loop is large, then we can consider keeping both as a performance optimization.

PR #336, opened 2023-05-10
Replace dataset config dictionary with object

aaron-kaplan:
This is a small step along the way to being able to display any variable on the map, and to stop faking the  dates of variables whose true time is different from the target season. It replaces the config dictionary representing a dataset with an object. There are two classes, one for obs datasets and one for forecast datasets, that both implement the same interface.

A future step will to be to describe a table column as a Dataset plus slicing criteria and an optional time shift. Trying to

### DIFF
--- a/fbfmaproom/fbfmaproom.py
+++ b/fbfmaproom/fbfmaproom.py
@@ -42,13 +42,6 @@ import dash_bootstrap_components as dbc
 from collections import OrderedDict
 
 
-CONFIG = pingrid.load_config(os.environ["CONFIG"])
-
-
-PFX = CONFIG["core_path"]
-TILE_PFX = CONFIG["tile_path"]
-ADMIN_PFX = CONFIG["admin_path"]
-
 SERVER = flask.Flask(__name__)
 
 SERVER.register_error_handler(ClientSideError, pingrid.client_side_error)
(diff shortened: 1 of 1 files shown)


fbfmaproom/fbfmaproom.py
```diff
@@ -71,82 +64,107 @@ def is_valid_root(path):
...
-    __name__,
-    external_stylesheets=[dbc.themes.BOOTSTRAP],
-    server=SERVER,
-    url_base_pathname=f"{PFX}/",
-    meta_tags=[
-        {"name": "description", "content": "content description 1234"},
-        {"name": "viewport", "content": "width=device-width, initial-scale=1.0"},
-    ],
-)
-APP.title = "FBF--Maproom"
```
aaron-kaplan, line 84: I had to move stuff around in the file so that functions that get called at import time are defined before they get called. This is annoying; I would like to switch to using the [application factory](https://flask.palletsprojects.com/en/2.3.x/patterns/appfactories/) approach, and was tempted to do it now, but decided to wait.

fbfmaproom/fbfmaproom.py
```diff
@@ -184,6 +202,39 @@ def format_enso(x):
...
+format_funcs = {
+    'number0': number_formatter(0),
+    'number1': number_formatter(1),
+    'number2': number_formatter(2),
+    'number3': number_formatter(3),
+    'number4': number_formatter(4),
+    'timedelta_days': format_timedelta_days,
+    'bad': format_bad,
+    'enso': format_enso,
+}
```
aaron-kaplan, line 214: I removed the `year` formatter from this dictionary. It was a closure over a local variable of `table_columns`, which made reorganizing things complicated. I eventually realized that the `year` formatter is only used once, for a column that's already a special case anyway, so there was no benefit to keeping it in the dictionary. Now `format_funcs` and `make_column` can be defined at the top level instead of inside `table_columns()`.

kgraaf:
This looks good! I don't really have anything to add, I did look over all of the diff and I don't see anything I'd really change

kgraaf (approved):

PR #337, opened 2023-05-11
unit step for cessation date search

remicousin:
Soooo... I think this is the unit-step we are looking for for a cessation date search loop. BUT... it's not super intuitive so definitely needs other eyes on it.

Operating in timedelta domain. The loop on time will also have to update the spell length which is the criteria to find onset. And if the input is rain rather than readily available soil moisture, the loop will also have to get an updated sm. Since spells typically involve more than 1 day, that loop will have to save just the bit of history (e.g. 5 days) it needs to evaluate spell length.

Let me know if I (don't) make any sense.

### DIFF
--- a/enacts/calc.py
+++ b/enacts/calc.py
@@ -434,7 +434,7 @@ def cess_date(
     for t in daily_data[time_dim][dry_spell_length_thresh:]:
         dry_day = daily_data.sel({time_dim: t}).expand_dims(dim=time_dim) < dry_thresh
         spell_length = (
-            spell_length.squeeze(time_dim, drop=True)+ dry_day.astype("timedelta64[D]")
+            spell_length.squeeze(time_dim, drop=True) + dry_day.astype("timedelta64[D]")
         ) * dry_day
         cess_delta = cess_date_step(
             cess_delta.squeeze(time_dim, drop=True),
@@ -809,7 +809,6 @@ def seasonal_cess_date(
         soil_moisture, search_start_day, search_start_month, end_day, end_month
     )
     # Apply cess_date
-    print(grouped_daily_data)
     seasonal_data = (
         grouped_daily_data[soil_moisture.name]
         .groupby(grouped_daily_data["seasons_starts"])
--- a/enacts/calc.py
+++ b/enacts/calc.py
@@ -332,6 +332,38 @@ def onset_date(
     return onset_delta
 
 
+def cess_date_step(cess_yesterday, spell_length):
+    """Updates cessation date delta according to today's soil moisture spell length
+
+    A cessation date is found at the first day of the first dry spell.
+    Today's cessation date delta is one day further away than yesterday's.
+
+    Parameters
+    ----------
+    cess_yesterday : DataArray[np.timedelta64]
+        Yesterday's distance in days from cessation date.
+        Can not have a time dimension of size greater than 1.
+    spell_length : DataArray[np.timedelta64]
+        Today's length in days of a dry spell.
+    
+    Returns
+    -------
+        DataArray[np.timedelta64]
+        Updated cessation date delta from today.
+
+    Notes
+    -----
+        It is understood that a spell is defined once long enough.
+        Thus its length is defined as soon as the spell reaches its minimum length.
+        Thus today's spell length if NaT if no spell has been long enough so far.
+        Once a cessation date is found (hence its delta from today),
+        the value of `spell_length` is ignored.
+    """
+    return cess_yesterday.where(
+        ~np.isnat(cess_yesterday), other=(np.timedelta64(2, "D") - spell_length)
+    ) - np.timedelta64(1, "D")
+
+
 def cess_date(
     soil_moisture, 
     dry_thresh, 
--- a/enacts/calc.py
+++ b/enacts/calc.py
@@ -418,24 +418,24 @@ def cess_date(
     """
     # Initializing
     spell_length = (
-        daily_data.isel({time_dim: 0}).expand_dims(dim=time_dim) < dry_thresh
+        daily_data.isel({time_dim: 0}) < dry_thresh
     ).astype("timedelta64[D]")
     cess_delta = xr.DataArray(np.timedelta64("NaT", "D"))
     # Loop
     for t in daily_data[time_dim][1:]:
-        dry_day = daily_data.sel({time_dim: t}).expand_dims(dim=time_dim) < dry_thresh
-        spell_length = (
-            spell_length.squeeze(time_dim, drop=True) + dry_day.astype("timedelta64[D]")
-        ) * dry_day
+        dry_day = daily_data.sel({time_dim: t}) < dry_thresh
+        spell_length = (spell_length + dry_day.astype("timedelta64[D]")) * dry_day
         cess_delta = cess_date_step(
-            cess_delta.squeeze(drop=True),
+            cess_delta,
             spell_length,
             np.timedelta64(dry_spell_length_thresh, "D"),
         )
     # Delta reference (and coordinate) back to first time point of daily_data
     cess_delta = (
-        cess_delta[time_dim] + cess_delta
-    ).squeeze(time_dim, drop=True) - daily_data[time_dim][0].expand_dims(dim=time_dim)
+        daily_data[time_dim][-1]
+        + cess_delta
+        - daily_data[time_dim][0].expand_dims(dim=time_dim)
+    )
     return cess_delta
 
 
--- a/enacts/calc.py
+++ b/enacts/calc.py
@@ -332,7 +332,7 @@ def onset_date(
     return onset_delta
 
 
-def cess_date_step(cess_yesterday, spell_length, spell_length_tresh):
+def cess_date_step(cess_yesterday, dry_spell_length, dry_spell_length_thresh):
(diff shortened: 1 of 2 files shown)


remicousin:
And once the loop has run its course, it can decide what reference it wants to use for the cess_delta it just produces and can adjust it accordingly to keep going with whatever the code might require next.

enacts/calc.py
```diff
@@ -332,6 +332,38 @@ def onset_date(
...
+    -----
+        It is understood that a spell is defined once long enough.
+        Thus its length is defined as soon as the spell reaches its minimum length.
+        Thus today's spell length if NaT if no spell has been long enough so far.
+        Once a cessation date is found (hence its delta from today),
+        the value of `spell_length` is ignored.
+    """
+    return cess_yesterday.where(
+        ~np.isnat(cess_yesterday), other=(np.timedelta64(2, "D") - spell_length)
+    ) - np.timedelta64(1, "D")
```
aaron-kaplan, line 364:
```
(2 - spell_length) - 1 =
1 - spell_length =
- (spell_length - 1)
```
That means the day after the dry spell started, doesn't it? Is that a mistake? Doesn't seem right but maybe I'm not interpreting it the way you intended.

It might be a little easier to reason about if you made cess_delta positive (and then subtract it instead of adding it when it comes time to convert it to a date).

Alternatively... I notice you're not using the time coordinate, unlike in the water balance step function. What if you make T represent the date you're currently processing, either as an absolute date or a timedelta since the start of the search period, and when you find a dry spell, put the T of the start of the dry spell into cess_delta. Then cess_delta represents the difference between the cessation date and the start of the search period, rather than between "now" and the cessation date. Seems like it might be simpler than incrementing every delta on each implementation, and maybe more efficient too.

remicousin: Let's say the spell length that makes a cessation is 5. Somewhere today spell length reaches 5 and -(5-1) is -4. 4 days from today (the 5th day of the 5-day long spell) was the 1st day of the 5-day long spell.

I believe (nearly) all you are saying otherwise will happen in the looping function, not in the step one. water_balance_step actually ignores the time dimension too. It's the loop calling it that involves it and that manipulates it the way that makes best sense for it.

So in that regards, I do expect the loop to eventually reset the reference time for those delta to the start of the search (if we keep doing in that way -- and btw, that will "flip" back the deltas to be positive values). The loop will get a new "now" time dim at each iteration, also needed by the water_balance_step (for whom the start of search date doesn't have meaning btw). Having both a now T and another reference T I feel will expose to complications as I will have to make sure operations won't try to broadcast one with the other, and fail because they are not the same value, so I will have to constantly remove one and put it back or something like that.

All the above is more gut feeling than careful assessment though. I would assume you are correct about the efficiency since there won't be an addition done at each iteration but only assignments when appropriate.

What do you think?

aaron-kaplan: OK for the meaning of `spell_length - 1`. Any thoughts about using a positive number, i.e. making the negative implicit? (It's not possible for a dry spell to start in the future and continue through today.)

I don't entirely follow your argument about the different ways of handling time, but it does seem natural for the step function not to know anything about the time dimension.

remicousin: We can make the negative implicit and operate in the positive numbers by saying that the output of cess_date_step is a distance (delta) such as cess_date = today - delta (currently it is assuming to be today + delta). That means that in the loop function, when the loop is over, we'll have a positive delta with respect to last T of the loop (T_last) so that cess_date = T_last - delta. When we give it the coordinate T as the start of the search of the cessation date, the new_delta will become T_last - delta - T_start_of_search, making a positive new_delta again, but then, the cess_date will be
T_start_of_search + new_delta, being a bit incoherent with the behavior of cess_date_step... but cess_date_step is of different nature of cess_date so... either way is fine with me.

remicousin: I could write the loop function in this PR if it helps see the whole thing together.

aaron-kaplan: The sign of the delta is just an aesthetic preference, and I don't have the full picture in my head yet, so it's fine to leave it if you don't share the preference.

Seeing the whole loop would probably help.

remicousin:
Changes cess_date to loop algo. This currently covers only the case where daily_data is already soil moisture. I added one test and it otherwise passes all the older tests. The spell_length initialization will change as it will need to be integrated inside the loop, along with the reduced water balance algo.

I did change cess_date_step as it appeared more natural to me to have spell_length be just that instead of NaT when the minimum number of days to define a spell has not been reached yet. So if the treshold is 3, the spell_length where it is 0, 1, 2, remains so and the new argument tells what to do about it inside the step. Initially I thought the loop would take care of that but as I was going, it seems unnecessarily more complicated.

Let me know what you think.

enacts/calc.py
```diff
@@ -345,39 +382,72 @@ def cess_date(
 
     Parameters
     ----------
-    soil_moisture : DataArray
-        Array of daily soil moisture.
+    daily_data : DataArray
+        Array of daily soil moisture or daily rainfall.
```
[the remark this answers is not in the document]
remicousin: My thinking was to test that either et, taw, sminit are all None, either are all non None, and to raise an exception that it should be one or the other and that accordingly, data should be sm or rain. But I am not sure what I could do to check that they do provide the right daily_data when all are None or not None...

enacts/calc.py
```diff
@@ -345,39 +382,72 @@ def cess_date(
...
+    spell_length = ((daily_data < dry_thresh)
+                    .rolling(**{time_dim: dry_spell_length})
+                    .sum()
+                    .dropna(time_dim)
+                    .isel({time_dim: 0})
+                    .expand_dims(dim=time_dim)
+                    .astype("timedelta64[D]")
+    )
+    cess_delta = cess_date_step(
+        (spell_length * np.nan).squeeze(time_dim, drop=True).astype("timedelta64[D]"),
```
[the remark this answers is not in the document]
remicousin: It's even simpler now.

enacts/calc.py
```diff
@@ -345,39 +382,72 @@ def cess_date(
...
-    dry_spell_roll = dry_spell.rolling(**{time_coord: min_dry_days}).sum() == min_dry_days
-    cess_mask = dry_spell_roll * 1
-    cess_mask = cess_mask.where((cess_mask == 1))
-    cess_delta = cess_mask.idxmax(dim=time_coord)
+    # Initializing
+    spell_length = ((daily_data < dry_thresh)
+                    .rolling(**{time_dim: dry_spell_length})
+                    .sum()
+                    .dropna(time_dim)
+                    .isel({time_dim: 0})
```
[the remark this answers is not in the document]
remicousin: It was laziness to get the right T associated with the result. It turns out it was wrong because just doing .sum() ignores whether there was contiguous dry days, and whether there was an interruption with a wet day.

So now it just initializes with day 1. Per definition, no cess date will be found over the first few iterations since there is a minimum length threshold to reach.

And all that is likely to change when covering the case where input is rain and the water balance must be done first before assessing spells, and thus further integrated in the loop.

enacts/calc.py
```diff
@@ -345,39 +382,63 @@ def cess_date(
...
+        dry_day = daily_data.sel({time_dim: t}).expand_dims(dim=time_dim) < dry_thresh
+        spell_length = (
+            spell_length.squeeze(time_dim, drop=True) + dry_day.astype("timedelta64[D]")
+        ) * dry_day
+        cess_delta = cess_date_step(
+            cess_delta.squeeze(drop=True),
+            spell_length,
+            np.timedelta64(dry_spell_length_thresh, "D"),
+        )
+    # Delta reference (and coordinate) back to first time point of daily_data
```
aaron-kaplan, line 435: I don't follow why all the squeezing and expanding is necessary, but I'll take your word for it. If it's just leftover from some previous version, try removing it.

enacts/calc.py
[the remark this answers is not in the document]
aaron-kaplan: I think what I had in mind when I made this comment is that maybe users would make fewer mistakes if this function had two separate and mutually exclusive arguments `soil_moisture` and `rainfall`, instead of having a single argument `daily_data` that does double duty.

Minor comment, take it or leave it.

aaron-kaplan (approved):
I'm fine with merging this version if you confirm that nothing is broken in the maproom.

enacts/calc.py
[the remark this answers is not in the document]
remicousin: How do you do mutually exclusive arguments? Both optional and you check and raise Exception if both (or none) are given? and in that very case, adding a layer of checks with the other optional arguments that belong to the rainfall case.

enacts/calc.py
remicousin: It was not a leftover but can indeed do without and add the T we want at the end of the function only. It might not fly when the water balance option kicks in but might as well start from something simpler

remicousin:
All the maprooms using cessation are working fine, and so after rebasing. When you tell me how-to, I'll include the mutually inclusive argument set up.

On a side note, I was thinking that we could also stop the loop once (if) all the cess_dates have been found... I would create a mask that shows where no data are missing, then testing in each loop instance that none of these are missing anymore (that would mean that all findable cess dates have been found) and then break the loop. Do you think that would save on efficiency? can one break a loop?

PR #339, opened 2023-05-15
Dataset config part 2

aaron-kaplan:
Continuing what I started in #336. Still no change in functionality, just contining to move towards having a single dataset abstraction that forecasts and obs both implement.

### DIFF
--- a/fbfmaproom/fbfmaproom.py
+++ b/fbfmaproom/fbfmaproom.py
@@ -163,9 +163,8 @@ def table_columns(dataset_config, predictor_keys, predictand_key,
         tooltip=None,
         type=ColType.SPECIAL,
     )
-    for key in predictor_keys:
+    for key in predictor_keys + [predictand_key]:
         tcs[key] = make_column(key, dataset_config)
-    tcs[predictand_key] = make_column(predictand_key, dataset_config)
 
     return tcs
 
--- a/fbfmaproom/fbfmaproom.py
+++ b/fbfmaproom/fbfmaproom.py
@@ -467,23 +467,14 @@ def select_forecast(country_key, forecast_key, issue_month0, target_month0,
     return da
 
 
-
-def select_obs(country_key, obs_keys, target_month0, target_year=None):
-    ds = xr.Dataset(
-        data_vars={
-            obs_key: CONFIG["countries"][country_key]["datasets"]["observations"][obs_key].open()
-            for obs_key in obs_keys
-        }
-    )
+def select_obs(country_key, obs_key, target_month0, target_year=None):
+    da = CONFIG["countries"][country_key]["datasets"]["observations"][obs_key].open()
     if target_year is not None:
         target_date = (
             cftime.Datetime360Day(target_year, 1, 1) +
             pd.Timedelta(target_month0 * 30, unit='days')
         )
-        try:
-            ds = ds.sel(time=target_date)
-        except KeyError:
-            raise NotFoundError(f'No value for {" ".join(obs_keys)} on {target_date}') from None
+        da = da.sel(time=target_date)
 
     with warnings.catch_warnings():
         # ds.where in xarray 2022.3.0 uses deprecated numpy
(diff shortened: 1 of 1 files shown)


fbfmaproom/fbfmaproom.py
```diff
@@ -106,14 +105,60 @@ def units(self):
...
+        path = data_path(self.path)
+        try:
+            ds = xr.open_zarr(path, consolidated=False)
+        except Exception as e:
+            raise Exception(f"Couldn't open {path}") from e
+        ds = ds.rename({
+            v: k
+            for k, v in self.var_names.items()
+            if v is not None and v != k
+        })
```
kgraaf, line 118: This is just a comment but every day Python's syntactic special cases never cease to amaze me

fbfmaproom/fbfmaproom.py
```diff
@@ -106,14 +105,60 @@ def units(self):
...
+
+
+        if val_min is None:
+            val_min = self.range[0]
+        if val_max is None:
+            val_max = self.range[1]
+        da.attrs["colormap"] = self.colormap
+        da.attrs["scale_min"] = val_min
+        da.attrs["scale_max"] = val_max
+        return da
```
kgraaf, line 129: I'm kind of confused what's going on here, I thought we were using this class you designed to store attributes, or is the class more a descriptor for a zarr file (set) (as opposed to the xarray data set)?

fbfmaproom/fbfmaproom.py
```diff
@@ -106,14 +105,60 @@ def units(self):
...
+
 
 class ObsDataset(Dataset):
```
kgraaf, line 132: Wasn't sure where to put this but do you think these classes or at least the base class belong in Pingrid? I feel like some of this functionality could be used in our other maprooms. Some is conceptually similar to stuff I've been doing. Perhaps this just needs to be revisited once my library is done (soon)

fbfmaproom/fbfmaproom.py
```diff
@@ -106,14 +105,60 @@ def units(self):
             self._units = self.open().attrs.get('units')
         return self._units
 
+    def open(self, val_min=None, val_max=None):
+        path = data_path(self.path)
+        try:
+            ds = xr.open_zarr(path, consolidated=False)
+        except Exception as e:
```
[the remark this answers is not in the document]
aaron-kaplan: I don't remember what past conversations we've had on the subject, but I often complain about people catching exceptions too broadly when there's a risk of handling an unforseen error as if it were the one the author foresaw. In the current case, there's no error-specific handling happening, it's just whatever goes wrong, I want to see the path we were trying to open in the traceback (that information had been missing, which makes it more work to debug the problem).

fbfmaproom/fbfmaproom.py
aaron-kaplan: We had two different functions for opening datasets depending on whether they were forecasts or obs. Then we had if/then/else scattered all over the place to deal with the fact that we never know which we'll be dealing with. What I'm doing here is trying to centralize the differences in one place so that most of the code doesn't need to care if a given set of numbers comes from a forecast or an obs.

The new class doesn't contain any data, we still use xarray for that. It's just an abstraction that helps with opening a zarr dataset and getting xarray out of it.

I'm not actually finding myself entirely convincing here. Have to think more explicitly about the role of each of the classes.

fbfmaproom/fbfmaproom.py
aaron-kaplan: Yeah maybe someday. Fleshing out the ideas here first.

fbfmaproom/fbfmaproom.py
aaron-kaplan: The question I was asking myself is whether the differences between ObsDataset and ForecastDataset are essential differences, or if they're just historical accidents and could be merged into one. After a closer look, I think some of them are incidental but the differences in the `select` method, and the differences between `ForecastDataset.select` and `xr.DataArray.sel`, are significant.

There might be simplifications to be made, but I'd like to go ahead with this for now if you have no further comments @kgraaf .

PR #341, opened 2023-05-19
Configure choice of maprooms in config file

aaron-kaplan:
This is time-sensitive because we need it in order to set up the Mali maprooms. If review is going to take a while for some reason, I'll just do Mali the way we did Zambia, as a temporary measure, but I'd rather not.

This makes it so that the list of maprooms to load comes from the config file, rather than being hard-coded in python. This gives multiple benefits:
- No longer need to copy the long `if __name__ == '__main__'` block into each maproom. All you need is app.py, even if you only want to configure and run a single maproom.
- We can (in the future, not set up yet) distribute all of the ENACTS maprooms as a library, and users will be able to choose which maprooms from the library they want to configure and use.
- This also provides a convenient mechanism for adding custom maprooms to the ready-made setup. You just have to have the custom maproom package on your python path and then name it in the config file.

The change to the config file looks huge, but it's mostly just indentation changes. I recommend viewing with whitespace differences suppressed (will add a link for that once the PR is created).

My original thought was that if you don't configure a given maproom then it won't be created. Unfortunately that's not how it currently works--because of the way we handle config-defaults.yaml, there ends up being some configuration for all maprooms. You have to explicitly set a maproom's configuration to `null` to turn it off.

READMEs still need updating. I'll work on that while you're reviewing the code.

- [x] Update READMEs
- [x] Test all countries' configs
- [x] fix assets


remicousin:
Looks great. Was just able to create the ZMD and ANACIM maprooms suite. I need to run to PT but I noticed something must be done to assets. Currently used by onset and wat_bal to put an insituttional icon in the top left corner of the page. That icon is broken. (assets may be used by other things but I just saw that one).

aaron-kaplan:
Thanks for catching the assets problem.

"Hide whitespace" is a toggle under the gear menu in the "Files changed" tab.

aaron-kaplan:
zmd and anacim are the only two I've tested. The other still need testing.

remicousin:
I ran all the config. None work but _zmd and _sng. They all come with different errors that seem indicative enough to fix. Let me know if you want me to do that. Note that _bd and _gt are not really enacts maprooms. They are just flexible forecasts (and they were created because that's what I had handily accessible to create such Maproom for seasonal and sub-seasonal). Nonetheless, they could probably be working in that framework, the point is rather that they are not destined to become an online product any time soon.

For _zmd and _sng, I was able to unbuild one or more maprooms very easily.

aaron-kaplan:
I think I have fixed the remaining errors that were caused by my changes. A few maprooms still don't work but I think they were already broken, so we can address them in subsequent PRs according to their urgency.

aaron-kaplan:
@remicousin please review.

remicousin (approved):
I was able to build all the maprooms but -kmd and -mali. -kmd seems to be looking after a strangely named shapefile and -mali doesn't know map_center. Both those should be trivial for me to fix.
-gt works but loads no data. Probably the files that it used to read are not there anymore. It doesn't matter anyway this was for development, not for a deliverable.

PR #342, opened 2023-05-23
converting daily et0 to zarr

remicousin:
It turns out I also have ENACTS daily precip in /data/remic/mydatafiles/ but maybe we should just stop writing in our locations and rather write to some common place at some point?

### DIFF
--- a/enacts/config-sng.yaml
+++ b/enacts/config-sng.yaml
@@ -39,7 +39,7 @@ datasets:
             T: 592
             Y: 31
             X: 29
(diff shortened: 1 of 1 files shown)


aaron-kaplan (commented):
I have been thinking (on and off) about how to organize zarr datasets under /Data, but We do need to decide on a good organization for zarr data under /Data, but I'm not sure proprietary ENACTS data belongs to that

aaron-kaplan (approved):

PR #344, opened 2023-05-25
Multi site take 2

aaron-kaplan:
Follow-up to https://github.com/iridl/python-maprooms/pull/341 that solves some problems I ran into when setting up the Mali server.


remicousin (approved):

aaron-kaplan:
Mali and Ethiopia configs are here:
https://bitbucket.org/iridl/python_maproom_malimeteo/
https://bitbucket.org/iridl/python_maproom_emi/

PR #345, opened 2023-05-25
Fix onset graph by correcting misnamed kwarg

aaron-kaplan:


remicousin (approved):
my bad... Could even get rid of it...

PR #351, opened 2023-06-01
Remic/raincess

remicousin:
This adds the option to have cess_date driven by rain rather than sm.

It made me re-arrange the test and agronomy files to be more coherent with the new set up for multi-page apps.

It turns out I needed not the water_balance with reduction but rather used directly water_balance_step.

The test is creating daily_rain data that would yield same sm (with et=5 and sminit=0) than the test with sm directly.

To be coherent with previous work, I will introduce stuff like et=xr.DataArray(et) so that the user can input numbers. I've just realized I've forgotten doing that.

This will break the maprooms since there is an additional (boolean) argument in cess_date, but I will correct the maprooms in this PR once we are good with the function. And since I'll have to do that, I'll rewrite the maprooms to use the rain as input as opposed to computing the full array of sm first. (at least for the case where it doesn't do multiple years, the multiple years case is covered by the changes brought up to the seasonal functions in this PR). Then in another PR, we can either work on the multiple years here, or pause on that and apply all those stepping changes to the "real" water balance function and use that one for good. Either way.

R


enacts/calc.py
```diff
@@ -417,13 +420,25 @@ def cess_date(
...
+        sm = daily_data.isel({time_dim: 0})
+    else:
+        if sminit is None or et is None or taw is None:
+            raise Exception("sminit, et and taw can not be None")
+        else:
+            sminit = xr.DataArray(sminit)
+            et = xr.DataArray(et)
+            taw = xr.DataArray(taw)
+            sm = water_balance_step(sminit, daily_data.isel({time_dim: 0}), et, taw)
+    spell_length = (sm < dry_thresh).astype("timedelta64[D]")
```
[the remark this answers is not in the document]
remicousin: we discussed having the first step as initializing step outside the loop in the previous PR #335 (see May 10 discussion), related to broadcasting.

enacts/tests/test_calc.py
```diff
@@ -555,9 +555,41 @@ def test_cess_date():
...
+         [5, 5, 7, 3, 5]],
+        dims=["X", "T"], coords={"T": t}
+    )
+    cess_delta = calc.cess_date(
+        daily_rain,
+        False,
+        1,
+        3,
+        et=xr.DataArray(5),
+        taw=xr.DataArray(10),
```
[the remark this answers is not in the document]
remicousin: I might have left theses because I wanted to see whether DataArray of DataArray is DataArray

aaron-kaplan:
How do you run the tests? When I run `pytest` in the `enacts` directory I get "No module named 'calc'".

aaron-kaplan:
Got it, `python -m pytest`. I forgot that they're not equivalent.

aaron-kaplan:
As you mentioned in the PR description, this is incomplete--the maprooms have to be adapted to the API change. I have some comments that need to be addressed before that happens, and some that can be addressed after or in parallel.

In https://github.com/iridl/python-maprooms/pull/371 I have suggested a different API. We should reach agreement on that before changing the maprooms (although it wouldn't be the end of the world if we had to change the maprooms twice--I don't think the change will be very big).

Comments that can be addressed separately because they don't affect the API:
- `test_seasonal_cess_date_*` are excessively slow (9s each), despite working on only a small amount of test data, both on your branch and my refactored one. I will look into this.
- Um, I had one more, maybe it will come back to me...

aaron-kaplan:
Oh, the other thing is that seasonal_cess_date is currently hard-coded to use pre-calculated soil moisture. Presumably you plan to give it the option to calculate from rain? Maybe that's part of adapting the maprooms to the API change.

remicousin:
ok... so first thing first, I am cool with the folding of step 0 into the loop. I don't know what I struggle so much with broadcasting...

Then, functions defined in functions and functions as arguments of functions is new paradigm to me... so it opens doors (once I clear up the fog). Then I have comments as I am processing this.

One is that I am not clear what's at stake in having 2 functions of one function if "if" checks in it. To me, cess_date is really a function of sm but one could not have sm but could derive sm from rain. But we don't want to do that because both cess_date(sm) and sm(rain) involve loops and we don't want to have to run one full loop for sm first, saving many things we will eventually discard, then run the loop again for cess_date (and not mentioning that now we are really talking about the one year (or season) case and that we want to run multiples years at once at some point soon).

But, now that I perceive the concept of functions as arguments of functions a bit, we could have once cess_date function that takes daily_sm as mandatory input. We however allow daily_sm to be stated as None, in which case the user would specify a function as optional argument to compute sm from. But we still need to pass as argument a whole (in time) array of rain. Can we pass it water_balance (not water_balance_step that ignores time) with reduce=True, sminit that updates every round of the loop with previous sm, the full daily_rain, et, taw...?

That would make things probably look complicated to the user, compared to 2 distinct functions. It would allow modularity for the user to plug something else for the sm_func. Knowing that at present we are working with the "simple" water_balance but at some point we'll have to replace it with the more complete one (currently in agronomy.py).

I guess what I come down to is whether it's worth to have one master function where the user is able to choose what input function to use to produce sm... the difficulty being that those input functions don't have the same inputs themselves than the simple case of the master function (they should also all produce same outputs but that seems less difficult to comply with). Instead of spawning functions as new ways of computing sm come through (But maybe I am extrapolating too much: we do have 2 ways currently but one is a special simpler case of the other)...

I'll leave at that for now to see what that triggers in your brains.

remicousin:
ah... yes... It seems I only thought through part of it...

remicousin:
Pushing so I can pull on shortfin01 and adapt Maprooms

remicousin:
it still took some time for cess date dependent images to show up (cess date and lengths ts and length map), on shortfin01...

It's not going to be the exact same results because sminit now is initialized every year, whereas before we had a full time stream of sm so we had a computed sm for the date the search starts. Something to discuss with hydrologists when we will actually have users for these. For now, I think it's ok.

aaron-kaplan:
> It's not going to be the exact same results because sminit now is initialized every year, whereas before we had a full time stream of sm so we had a computed sm for the date the search starts.

Hopefully the results aren't very sensitive to starting with sminit vs. starting with the computed value from the previous year. If they are sensitive to it then it seems like a problem, because errors will accumulate.

remicousin:
> > It's not going to be the exact same results because sminit now is initialized every year, whereas before we had a full time stream of sm so we had a computed sm for the date the search starts.
> 
> Hopefully the results aren't very sensitive to starting with sminit vs. starting with the computed value from the previous year. If they are sensitive to it then it seems like a problem, because errors will accumulate.

From my experience with past simulations, it's not very sensitive with intialization, except for the 1st few days so it would impact finding very early cessation dates from the search date. If the search date is like 1 month from the end of the rainy season, it would not matter, if someone plays with starting their search later, then at what value they initialize may matter. I think it's a matter of communicating that in the App, and maybe give the initialization as input choice.

Note that the difference is not with starting with value computed from previous year. The way it was set up before, I was computing the entire 40 years of daily soil moisture. Then if you start searching on April 1st, it was starting with the April 1st value of that year (or March 31st of that year).

enacts/calc.py
```diff
@@ -736,7 +792,7 @@ def seasonal_onset_date(
     # + seasonal_onset_date.onset_delta
     return seasonal_onset_date
 
-def seasonal_cess_date(
+def seasonal_cess_date_from_sm(
```
aaron-kaplan, line 795: Ugh. If we have to make two different versions of lots of different functions, that takes the air out of my argument for splitting cess_date_from_sm and cess_date_from_rain. Is this as far as it goes or is it likely to propagate further?

enacts/calc.py
```diff
@@ -804,9 +860,65 @@ def seasonal_cess_date(
...
+    seasons_ends = grouped_daily_data["seasons_ends"].rename({"group": time_dim})
+    seasonal_cess_date = xr.merge([seasonal_data, seasons_ends])
+
+    # Tip to get dates from timedelta search_start_day
+    # seasonal_onset_date = seasonal_onset_date[time_dim]
+    # + seasonal_onset_date.onset_delta
+    return seasonal_cess_date
+
+
+def seasonal_cess_date_from_rain(
```
aaron-kaplan, line 878: This function is mostly identical to seasonal_cess_date_from_sm. If we decide to keep the functions separate, I would like to factor out the common parts. I see how to do it, but I won't bother writing it down until we reach a decision on splitting or merging the functions.

aaron-kaplan (commented):
What is performance like now?

enacts/calc.py
remicousin: This was mostly to have the Maprooms work. If our work in PR #372 turn out successfully, those named seasonal_ xxx functions will become obsolete and we will apply our seasonal operators to whatever functions.

enacts/calc.py
remicousin: This was mostly to have the Maprooms work. If our work in PR #372 turn out successfully, those named seasonal_ xxx functions will become obsolete and we will apply our seasonal operators to whatever functions.

aaron-kaplan: OK. Do you want to merge this now and revisit later, or should we keep it open until PR #372 is finalized?

remicousin: I'd rather merge because we aren't sure how longer PR #372 will take; because from here, I also have to see what are the next steps in making the "main" water balance function step-wise; and because this PR also rearranges the tests and the functions locations and I'd rather not have that type of change hanging any longer.

remicousin:
Shall we merge this? Porting to water balance can come later, as well as a applying multiple-year seasonal operators when they are ready.

remicousin:
@aaron-kaplan , I don't think the changes here are needed to progress on the water balance work, but since I also moved files around, it may be better to merge sooner than later or it will be painful to rebase...

aaron-kaplan:
I scanned through the comments and I didn't find anything that hasn't been addressed. Do you know what this was waiting for?

Github says there are conflicts. Please rebase so I can review what will actually happen when we merge.

remicousin:
rebased, removed a print from the tests. Passed all the tests (arccos warning is there but being fixed in other PR). Maprooms (tested Senegal "onset" and "wat_bal") seem to work fine. Bug to click ts to update map still there but being fixed in other PR. Cessation dates and length of season are not timing out on shortfin01 and calculated.

I think there was nothing else to wait for but your last scan and then your clicking the approve button.

aaron-kaplan (approved):

aaron-kaplan:
Whoops, after approving I noticed that there are still conflicts.

aaron-kaplan (changes requested):
resolve conflicts

remicousin:
rebased

remicousin:
@aaron-kaplan we probably want to merge this to facilitate the likely new or follow-up work coming up

remicousin:
I am going to rebase

remicousin:
Rebased... there was no conflict to fix manually

aaron-kaplan:
Did you test after rebasing? This was forked a long time ago, so I wouldn't be surprised if something incompatible happened in the meantime.

remicousin:
I built successfully the Water Balance and Growing season Maprooms for Senegal, after having changed in a couple of places inputs of functions such as et=5 back to just 5. I think we wanted that to ease readability but it failed and complained that this type of argument must be last. That bug it also on master though. So could change it here, or first thing first after merging this.

aaron-kaplan:
Are you saying those maprooms are broken on master? How did we break them without noticing?

If you have the changes all ready, might as well add them to this branch.

remicousin:
But I can and will make the change here.

remicousin:
done

aaron-kaplan (approved):

remicousin:
We need not deploy anything for now. There is more to come and Jim said we can deliver end of June. At that time, there might be improvements that make the cessation date calculations (and those derived from) more efficient or usable at all on all partners' we wanted to install to. And there will be some Water Balance products.
So unless some other partner brings about something that requires updates before then, we're good until June.
On a side note, it also means we will probably want the enactstozarr updating facility also at that time.

PR #352, opened 2023-06-06
Remic/watbalvars

remicousin:
Adding other desired variables derived from the water balance. Some of the work revealed some problems and called some fixes. Each commit is about one of those variables and their associated fix so it can be easier to read commit by commit. The last commit adds the definition of the new variables.

The first commit adding PAW revealed that a previous change broke the capacity to click the ts to pick a day to map. This broke because we removed the year from the abscissas of the graph so we have to get that year back to select the right day.

In adding water excess (2nd commit), I dropped a comment that may or may not call for a change in pingrid. Likely not: it's about having a test return NaN instead of False when one of the compared objects is NaN.

To add effective precipitation (3rd commit), I had to introduce the runoff calculation of the water balance. In runoff function, I had dropped all NaNs with the intention to get rid of the first few days made NaN by the rolling of api. It turns out this was a bit brutal because it would alter the shape of tiles that have coordiantes out of the domain of data definition. So instead I changed api to specifically drop the first few days; and I had runoff stop dropping NaNs.

There are some more variables to add but these make only sense (both in space and time) with a variable ET, so need to be able to input such an ET first (which is in the plans once I compute an ET climatology).

### DIFF
--- a/enacts/wat_bal/maproom_monit.py
+++ b/enacts/wat_bal/maproom_monit.py
@@ -381,37 +381,34 @@ def wat_bal_ts(
     planting_year=None,
     time_coord="T",
 ):
-    water_balance_outputs = wat_bal(
-    precip,
-    et,
-    taw,
-    planting_day,
-    planting_month,
-    kc_init_length,
-    kc_veg_length,
-    kc_mid_length,
-    kc_late_length,
-    kc_init,
-    kc_veg,
-    kc_mid,
-    kc_late,
-    kc_end,
-    planting_year=planting_year,
-    time_coord=time_coord,
-)
-    for wbo in water_balance_outputs:
-        if map_choice == "paw" and wbo.name == "sm":
-            ts = 100 * wbo / taw
-        elif map_choice == "water_excess" and wbo.name == "sm":
-            ts = xr.DataArray(
-                np.isclose(wbo, taw).cumsum(),
-                dims="T",
-                coords={"T": wbo["T"]},
-            )
-        elif map_choice == "peff":
-            ts = precip_effective
-        elif (wbo.name == map_choice):
-            ts = wbo
+    try:
+        water_balance_outputs = wat_bal(
+            precip, et, taw,
+            planting_day, planting_month,
+            kc_init_length,
+            kc_veg_length,
+            kc_mid_length,
+            kc_late_length,
+            kc_init, kc_veg, kc_mid, kc_late, kc_end,
+            planting_year=planting_year,
+            time_coord=time_coord,
+        )
+        for wbo in water_balance_outputs:
+            if map_choice == "paw" and wbo.name == "sm":
+                ts = 100 * wbo / taw
+            elif map_choice == "water_excess" and wbo.name == "sm":
+                ts = xr.DataArray(
+                    np.isclose(wbo, taw).cumsum(),
+                    dims="T",
+                    coords={"T": wbo["T"]},
+                )
+            elif map_choice == "peff":
+                ts = precip_effective
+            elif (wbo.name == map_choice):
+                ts = wbo
+    except TypeError:
+        #Later tested to return error image rather than broken one
+        ts = None
     return ts
 
 


remicousin:
@aaron-kaplan maybe it's time to put this one back on your radar.

enacts/wat_bal/maproom_monit.py
```diff
@@ -127,6 +127,9 @@ def update_time_sel(planting_day, planting_month, graph_click, current_options):
     if dash.ctx.triggered_id == "wat_bal_plot":
         time_range = current_options
         the_value = graph_click["points"][0]["x"]
+        for a_date in time_range:
+            if a_date.startswith(the_value):
```
aaron-kaplan, line 131: Looks like you're doing string operations on string representations of dates. That feels kind of icky--I probably would have parsed the dates to datetimes and then done date operations on them. I guess this way is fine if we're sure that the date strings are generated the same way, e.g. we're not going to have January represented as `1` in one place and `01` in another.

enacts/wat_bal/maproom_monit.py
```diff
@@ -352,6 +352,8 @@ def wat_bal_ts(
         for wbo in water_balance_outputs:
             if map_choice == "paw" and wbo.name == "sm":
                 ts = 100 * wbo / taw
+            elif map_choice == "water_excess" and wbo.name == "sm":
+                ts = ((wbo / taw) == 1).cumsum()
```
aaron-kaplan, line 361: Using `==` with floating point numbers is often not the right thing to do.  Is `x / x == 1` guaranteed to be true? When `wbo` gets truncated to `taw`, does it always end up exactly equal to `taw`? Probably, but `isclose` might be safer.

enacts/wat_bal/maproom_monit.py
```diff
@@ -336,13 +336,18 @@ def wat_bal_ts(
         p_d.dt.year == planting_year, drop=True
     )).squeeze(drop=True).rename("p_d")
     precip = precip.where(
-        (precip["T"] >= p_d) & (precip["T"] < (p_d + np.timedelta64(365, "D"))),
+        (precip["T"] >= p_d - np.timedelta64(7 - 1, "D"))
```
aaron-kaplan, line 339: Why did you add this timedelta shift?

enacts/wat_bal/maproom_monit.py
```diff
@@ -549,7 +556,10 @@ def wat_bal_tile(tz, tx, ty):
     p_d = calc.sel_day_and_month(
         precip["T"], int(planting_day), planting_month1
     ).squeeze(drop=True).rename("p_d")
-    precip = precip.sel(T=slice(p_d.dt.strftime("%-d %b %y"), the_date))
+    precip = precip.sel(T=slice(
+        (p_d - np.timedelta64(7 - 1, "D")).dt.strftime("%-d %b %y"),
+        the_date,
+    ))
```
aaron-kaplan, line 562: You had to make this change in two places. Consider factoring out the calculation that is common to wat_bal_ts and wat_bal_tile.

enacts/wat_bal/maproom_monit.py
remicousin: So the difficulty here resides in the fact that we are plotting two time series, for the same days of different years. Because of that, we decided to remove the year from the label of the x-axis and so the_value is of the format d Mmm (e.g. 9 Oct) while time_range has the current year. So that's why I had to do this gymnastic to get the correct value in time_range list by having only the d Mmm from clicking the time series.

This being said, I preserved the year in the time series customdata but only for the current year so that it shows when mousing over the curve. Thus, I could retrieve the year from customdata. However, the user can also click the comparison year to select a date from time_range so I'd have to put the current years in the customdata of the comparison year and make sure it is not shows in the ts graphs...

Then once I have both day, month and year as strings from the click on the graph, I can indeed work on datetimes rather than strings. But all the above sounds tedious so maybe the whole paragdigme of doing this should be revisited. On the other hand, I do like that feature of clicking the time series to see the map at that time location and I think it then deserves to become part of general functionality to be saved somewhere... so maybe leave all that to a dedicated PR?

enacts/wat_bal/maproom_monit.py
remicousin: Because I introduced the "runoff" part of the process which is based on 6-day rolling window applied to the precip so one needs to keep 6 days of precip prior to planting date (p_d) to make sure effective precipitation exists for planting.

enacts/wat_bal/maproom_monit.py
remicousin: yes... but there isn't much more than that in common. wat_bal_ts takes a slice from p_d - 6 to today; whereas wat_bal_tile takes a slice from p_d - 6 to user-chosen date (located somewhere between p_d and today). And the logic to identify p_d, and slice things are different in each

enacts/wat_bal/maproom_monit.py
aaron-kaplan: Use datetimes as values, and hovertemplate and tickformat to define how they should be presented visually.

enacts/wat_bal/maproom_monit.py
aaron-kaplan: use a variable `window = 7` instead of hard-coding the literal 7 in multiple places.

enacts/wat_bal/maproom_monit.py
aaron-kaplan: It's more complicated than I realized because there are three different variants of the calculation: for the map, for the monitoring curve, and for the reference year curve. Nonetheless, Rémi will look into factoring out common code.

enacts/wat_bal/maproom_monit.py
```diff
@@ -372,15 +372,12 @@ def wat_bal_ts(
...
 
 
-def plot_scatter(ts, name, color, dash=None, customdata=None):
-    hovertemplate = "%{y} on %{x}"
-    if customdata is not None:
-        hovertemplate = hovertemplate + " %{customdata}"
+def plot_scatter(ts, name, color, dash=None):
     return pgo.Scatter(
-        x=ts["T"].dt.strftime("%-d %b"),
+        x=ts["T"].data,
```
aaron-kaplan, line 377: How did you know this `data` attribute existed? I can't find it in the documentation. It seems to be the same as `values`, which is documented.

enacts/wat_bal/maproom_monit.py
```diff
@@ -372,15 +372,12 @@ def wat_bal_ts(
...
         y=ts.values,
-        customdata=customdata,
-        hovertemplate=hovertemplate,
+        #xaxis_tickformat="%d %B",
```
aaron-kaplan, line 379: The commit comment says this is broken, possibly because of missing values. I don't understand how there could  be missing values in T. What are the symptoms?

enacts/wat_bal/maproom_monit.py
remicousin: it actually is in the [doc](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.data.html).

aaron-kaplan: Thanks. Not sure why I failed to find it.

enacts/wat_bal/maproom_monit.py
remicousin: you're right, it can't be missing values in T... Below is the error that is raised. All works fine when xaxis_tickformat is defined when using update_layout though.

ValueError: 'NoneType' object is not subscriptable

Invalid value received for the 'xaxis' property of scatter

    The 'xaxis' property is an identifier of a particular
    subplot, of type 'x', that may be specified as the string 'x'
    optionally followed by an integer >= 1
    (e.g. 'x', 'x1', 'x2', 'x3', etc.)
        

Property does not support subscripting:
xaxis_tickformat
^^^^^

enacts/wat_bal/maproom_monit.py
remicousin: I factored out up to a certain point. Passed that point, peffective and application of the spwb function are probably factorable out, but there are things in the way that I am not sure what they are for, and/or where they can/must be. These are a load and a try in the ts section, and there is all the tiling business in the map section.

enacts/wat_bal/maproom_monit.py
remicousin: Done... might be factored out and more graciously handled if factoring out can be expanded.

enacts/wat_bal/maproom_monit.py
remicousin: This should be implemented in full now.

enacts/wat_bal/maproom_monit.py
aaron-kaplan: This is resolved, right?

enacts/wat_bal/maproom_monit.py
```diff
@@ -492,23 +553,45 @@ def wat_bal_plots(
             error_msg="Please ensure all input boxes are filled for the calculation to run."
         )
 
+    #Save actual start of ts2
+    ts2_start = ts2["T"][0]
+
+    #Find corresponding day closer to ts
     p_d2 = calc.sel_day_and_month(
         precip["T"], int(planting2_day), calc.strftimeb2int(planting2_month)
```
aaron-kaplan, line 561: Why are we passing around a string representation of the month, and converting it to an integer each time we need to use it in a calculation? Wouldn't it be more straightforward to use integers for values (not labels) in the menu? (Not a blocker.)

enacts/wat_bal/maproom_monit.py
```diff
@@ -492,23 +553,45 @@ def wat_bal_plots(
...
+    #to feed to customdata to hovertemplate
+    ts2_customdata = calc.sel_day_and_month(
+        precip["T"], ts2["T"][0].dt.day, ts2["T"][0].dt.month
+    )
+    ts2_customdata = ts2_customdata.where(
+        abs(ts2_start - ts2_customdata) == abs(ts2_start - ts2_customdata).min(),
+        drop=True,
+    ).squeeze(drop=True)
+    ts2_customdata = pd.date_range(datetime.datetime(
+        ts2_customdata.dt.year.values,
```
aaron-kaplan, line 582: `datetime.datetime` can't do anything with arrays,  so i guess you're counting on the `squeeze` to turn `ts2_customdata` into a scalar? Is that guaranteed to work? What if there are two entries for the same date? Too unrealistic to be worth checking for?

I don't understand how `pd.date_range` is being used. You specify `start` and `periods`, but neither `freq` nor `end`, so how does it know how wide to make the intervals?

aaron-kaplan:
There are conflicts. Please rebase.

enacts/wat_bal/maproom_monit.py
```diff
@@ -134,9 +136,13 @@ def update_time_sel(planting_day, planting_month, graph_click, current_options):
         ).squeeze()
         time_range = time_range.where(
             time_range >= p_d, drop=True
-        ).dt.strftime("%-d %b %y").values
-        the_value = time_range[-1]
-    return time_range, the_value
+        )
+        time_options = [
+            {"label": tr.dt.strftime("%-d %b %y").values, "value": tr.dt.strftime("%Y-%m-%d").values}
```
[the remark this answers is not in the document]
remicousin: for the label, %-d means the leading 0 is not printed and that is what is shown to the user, e.g. 8 Jan 2009, as a somewhat human friendly way of printing dates. For the values, it creates e.g. 2009-01-08, a string format that is readily understood by Xarray and friends and used as very first examples in Xarray [doc](https://docs.xarray.dev/en/stable/user-guide/time-series.html).

So that when time_selection is an input of a callback, it can be used readily in slice or what have you.

I guess in more recent work for flex. fcst. Maproom, we've been cleaner on what we give to and take from menus that propose dates, so maybe when that piece is merged there (maybe it is already), I can make an Issue to port (and maybe further clean-up / generalize) this handling of dates in menus and callbacks to other Maprooms, including this one here?

enacts/wat_bal/maproom_monit.py
remicousin: This is what made us use %Y-%m-%d as a "standard" value to pass around callbacks, menus, functions and plots, and then reformat if we wanted to. I think part of the problem was discussed in the office so doesn't transpire here. I believe we could not use datetimes values because in the menu, it becomes a string or something. So we reverted to using %Y-%m-%d. I might be wrong... I don't remember well...

enacts/wat_bal/maproom_monit.py
remicousin: This is a legacy of controls.py's Month function. I can make an issue to revisit this as we see fit

enacts/wat_bal/maproom_monit.py
remicousin: correct, if no squeeze, error is returned:

TypeError: only integer scalar arrays can be converted to a scalar index

where could return 2 values if the user picked a second planting day equidistant from 2 first planting dates of consecutive years (e.g. something like 1 Jul for the second p d and 1 Jan for the first -- probably not quite that but you get the idea). So it's possible even though the spirit of the Maproom is to compare with what-if considerations such as what if I had planted a couple weeks earlier/later, but in any case looking at the same season of interest thus not going 6 months away. But as it stands the Maproom allows it to happen.

[pd.date_range](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.date_range.html) default freq is "D"

remicousin:
Addressed latest reviews through commit or comments. Rebased, passed tests, Senegal Wat Bal Monit Maproom looks to work ok.

remicousin:
@aaron-kaplan , in anticipation of apparent wake-up from AICCRA Senegal project, it may be good to merge this sooner than later.

enacts/wat_bal/maproom_monit.py
aaron-kaplan: I think dash uses JSON as its data representation, and JSON has no date type.

enacts/wat_bal/maproom_monit.py
aaron-kaplan: I would like to further reduce the duplication between `wat_bal_tile` and `wat_bal_ts`.

We can do the `load` in both cases (or in neither if you find that it makes no difference to performance).

The try/except in `wat_bal_ts` seems to have been written with a specific case in mind, and I can't guess what that is (a comment would have been useful). Whether it belongs inside or outside the factored-out function depends on whether the error happens in the calculation part or not. I suggest just getting rid of the try/except, doing the refactoring, and then putting back the try/except in the right place when we see the error again.

Another difference is that `wat_bal_ts` gets `taw` as a parameter, while `wat_bal_tile` reads it from a file. That seems like an arbitrary and unnecessary difference.

enacts/wat_bal/maproom_monit.py
aaron-kaplan: Since it's possible for there to be more than one entry, let's explicitly select the zeroth one, rather than crashing if we ever hit that case.

remicousin: done

enacts/wat_bal/maproom_monit.py
remicousin: Done. There is probably a need to fix the data reading as you did for flex. fcst. but I'd rather have that be a separate PR.

_ts and _tile differ a bit because _tile is what is directly called by the Callbacks whereas _ts is called (twice) by _plots that is the Callback one and where  raw is read from file.

enacts/wat_bal/maproom_monit.py
```diff
@@ -381,37 +381,34 @@ def wat_bal_ts(
...
+                    dims="T",
+                    coords={"T": wbo["T"]},
+                )
+            elif map_choice == "peff":
+                ts = precip_effective
+            elif (wbo.name == map_choice):
+                ts = wbo
+    except TypeError:
+        #Later tested to return error image rather than broken one
+        ts = None
```
aaron-kaplan, line 411: Is there any information available to explain why it failed? (If not, this is fine.)

aaron-kaplan (approved):

enacts/wat_bal/maproom_monit.py
remicousin: Before running wat_bal_ts, the location (X/Y) has already been applied and tested. So here, it's mostly to fail graciously if something else goes wrong, in particular poor choices of callback Input. And what comes right after calling each of the 2 wat_bal_ts call is:

if (ts is None):
        return pingrid.error_fig(
            error_msg="Please ensure all input boxes are filled for the calculation to run."
        )

remicousin: I am going to resolve this conversation so I can merge. We can always get back to this in another PR.

PR #367, opened 2023-07-24
Remic/cs cpolish

remicousin:
Some polishing that seemed worth doing sooner than later: better titles and better colorscales.
The set up of the Colors would probably be better with a registry like for the Scales but I don't have time for that. I think there is an Issue listed for this.

### DIFF
--- a/enacts/crop_suitability/maproom_crop_suit.py
+++ b/enacts/crop_suitability/maproom_crop_suit.py
@@ -63,6 +63,13 @@ RESOLUTION = rr_mrg['X'][1].item() - rr_mrg['X'][0].item()
 # The longest possible distance between a point and the center of the
 # grid cell containing that point.
 
+CROP_SUIT_COLORMAP = pingrid.ColorScale(
+    "crop_suit",
+    [BROWN, BROWN, ORANGE, ORANGE, YELLOW, YELLOW,
+        PALEGREEN, PALEGREEN, GREEN, GREEN, DARKGREEN, DARKGREEN],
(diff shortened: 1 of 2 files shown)


enacts/crop_suitability/maproom_crop_suit.py
```diff
@@ -554,16 +561,27 @@ def set_colorbar(
     data_choice,
 ):
     if data_choice == "suitability_layer":
+        colormap = pingrid.ColorScale(
+            "crop_suit",
+            [BROWN, BROWN, ORANGE, ORANGE, YELLOW, YELLOW,
+                PALEGREEN, PALEGREEN, GREEN, GREEN, DARKGREEN, DARKGREEN],
+            [0, 0.5, 0.5, 1.5, 1.5, 2.5, 2.5, 3.5, 3.5, 4.5, 4.5, 5],
+        )
```
aaron-kaplan, line 569: Instead of copy-pasting, define this once and assign to a variable that can be used in multiple places.

aaron-kaplan (approved):

remicousin:
I think that's it. The rest is either more cosmetics or more robust coding that is however hidden to the user.

(what part of the change requires the new container image?)

aaron-kaplan:
Any change to the python code requires a new container image, as that's how the code gets to the servers.

remicousin:
ok. I don't know if python_maproom_malimeteo and _zmd are up to speed. _malimeteo probably just need a crop_suitability: null added to its config file, but _zmd is pretty much older set up...

aaron-kaplan:
That's ok, that's why an explicit action is required to upgrade (they don't just automatically get the latest version). We can upgrade each partner site individually on their own schedule.

PR #368, opened 2023-07-25
max year can be default target year which can be max year

remicousin:
removing hard-coded entry. There is probably more to do but that can be part of the issue about checking that paramters are in the right config files.


aaron-kaplan (approved):
OK for the quick fix, but it will probably  be annoying to have to increment the default year in the config file once a year, so consider using the most recent year in the dataset rather than a hard-coded year as the max and default.

PR #370, opened 2023-07-26
Remic/conf

remicousin:
Rationalizing config files. (python_maproom_ conf files coming soon.


enacts/crop_suitability/layout_crop_suit.py
```diff
@@ -41,6 +41,7 @@ def app_layout():
     lon_max = np.around((rr_mrg["X"][-1]+lon_res/2).values, decimals=10)
     lat_label = str(lat_min)+" to "+str(lat_max)+" by "+str(lat_res)+"˚"
     lon_label = str(lon_min)+" to "+str(lon_max)+" by "+str(lon_res)+"˚"
+    target_year = str(rr_mrg["T"][-1].dt.year.values)
```
aaron-kaplan, line 44: Are you sure you've tested this? `values` returns an array, so I would have thought `str(x.values)` came out something like `[1995]` instead of `1995`. But if you're sure it works then it's fine.

aaron-kaplan: Also (sorry for nitpicking) I think `max_year` would be a better name for this variable. If we decide to use something other than the most recent year as the default target, this value will no longer be the target but it will still be the max.

aaron-kaplan:
For software that runs only on our servers, I like releasing every small change as soon as we make it, because it's easier to deal with errors if you discover them right after you make the mistake rather than weeks later when release a bunch of changes at a time. But this application is now running on several partner servers, and that deployment paradigm probably won't work as well there. We should probably start batching changes up and releasing them from time to time when there's enough new stuff to make people actually want to upgrade. Especially in places like Ethiopia where we have to schedule an AnyDesk session to make any change. 

To start with, we should start a `CHANGELOG.md` file to keep track of (a) new features introduced in each version, and (b) backwards-incompatible changes (like the config file format change in this PR) that require some action on the administrator's part when upgrading.

Ideally we also keep track of who is running which version of the software so we know who to email when there's something high-priority like a security vulnerability. We would probably fail to keep that list up to date by hand; I'll look into doing something automatic.

enacts/crop_suitability/layout_crop_suit.py
remicousin: tested again and also printed it to see it and it's just the 4 characters of the year...

no problem with nitpicking, half of them tend to have crossed my mind and I was too lazy to act on it.

aaron-kaplan:
To reiterate, I think this PR should introduce a CHANGELOG.md containing two lists: a list of improvements, and a list of backwards-incompatible changes. Subsequently, every PR should update the change log.

remicousin:
Do you have a template, or best practices, or something like this, for that?

aaron-kaplan:
No. On further reflection, keeping a list of new functionality is probably overkill at this stage of the project. Just make note of the backwards-incompatible change to the config file format.

remicousin:
How about that?

enacts/CHANGELOG.md
```diff
@@ -0,0 +1,19 @@
+# ENACTS Maprooms
+
+# Change Log for config giles
```
aaron-kaplan, line 3: giles -> files

aaron-kaplan (approved):

aaron-kaplan:
Oops, I started a comment last night but had to run for the bus and then forgot to finish it later.

Most/all of the country config files in this repo are out of date. It's ok for config files in other repos to lag behind this repo because their Dockerfiles specify which version of the application they expect, but the config files in this repo are implicitly always for the current version. In principle, we should always keep them up to date, but we haven't been, and maybe testing all of them each time we change something is too much effort. What do you think?

Hopefully it's just a temporary problem, as we'll eventually be moving them all to separate repos. I guess I'll re-approve and we can figure out how to deal with it later.

aaron-kaplan (approved):

aaron-kaplan (approved):
Remove KMD config.

PR #373, opened 2023-08-03
Remic/controls

remicousin:
First step towards more compact controls. First couple commits actual fixes to a proper set up.
This is not necessarily prettier but it is more compact.
Cards may not be a good choice as the Header+Body takes a lot of space.
Labels in FormFloating are nicer (except it seems I can't reduce the amount of white space they generate) but actually apply only to Input, and we would also want some labeling that encompasses more than one control (that Card's Header do well).
Benno's fieldsets are doing all that well.

So next step may be to look for something other than Cards... Even though there doesn't seem to be an alternative in dbc. Maybe it's ok... I wish I could squeeze the vertical size of "Pick a Point" and put it in the top banner of the page


aaron-kaplan:
Please give  a working config for testing. I tried a few of the config files from this repo as well as the one from python_maproom_emi and none of them worked.

remicousin:
Could I merge this? maybe undoing the application of fieldsets to onset's layout (everything before that are improvement on controls, and one can switch as they like between controls and fieldsets) ? Pepsico are paying a lot of attention to looks so I'd like to use fieldsets in their app. So I guess I could also just take fieldsets here and put it in pepsico...

aaron-kaplan:
There are conflicts.

How do you envisage the future of this change? You don't plan to maintain both controls.py and fieldsets.py in perpetuity, do you? If we merge this now, how soon will we be able to eliminate controls.py?

remicousin:
I rebased, surprisingly without much trouble (if only I could remember to save file before committing/pushing them....)

Chrome looks just as fine.

Screenshot 2024-12-09 at 2 26 21 PM

I did have in mind that we might want to support several UI, or support the capacity to plug in other UI. I didn't go very far into such a standardization. The only thing is that as of now, one just needs to either import controls or fieldsets to switch from one UI to another (the other changes in onset apply to both set up). The 2 remain close: the Block is the really different one; 2 functions are identical; all other are quite close. So there may be a better way to propose the 2 UIs. But maybe that is for later work?
In the long term, I could imagine that if controls/fieldsets were well documented and if constraints on how to make your own were well laid out, then yeah, there could be multiple such .py files... Doesn't mean we'd maintain them all... not even sure if controls.py would be one we'd maintain ultimately...

remicousin:
@aaron-kaplan , I need the ISIMP App to be in a showable stage no later than my end of Tuesday. Either way I am going to have to sweat a bit on css and such because I know the Pepsico folks are going to pay too much attention to that. I made progress everywhere I could on the app itself and now setting up the Controls has become a blocker to go further. So I need to know if I should expect to be able to use what's in this PR soon soon, or not, in which case I'll work from what we currently have in master for this very app.

aaron-kaplan:
I'm not keen on supporting both cards and fieldsets, especially if it means maintaining two similar but different versions of each control. I think we need a framework that's simple and prescriptive, not flexible.

Cards looked terrible in Chrome last I checked (every input stretched across the full width of the box), so if fieldsets look better, maybe we should standardize on that. Alternatively, maybe we could take inspiration from your fieldsets to improve cards.

For the imediate future, I suggest you put fieldset.py in the isimp directory. If Tufa's Ethiopia project goes according to plan, I should be spending significant time on python maprooms in the new year.

remicousin:
ok. I removed fieldsets from this PR. There are still some updates to controls that make the Cards a bit more compact. And also the factoring out of the Pick Location control. I paste below a screenshot from Chrome, to be compared with [this](http://213.55.84.78:8082/python_maproom/onset/). Even though I don't see it worse on Chrome than on Firefox...

Screenshot 2024-12-16 at 2 13 42 PM

controls.py
```diff
@@ -22,10 +22,10 @@ def Text(id, default):
         dbc Input component with text inputs.
     """
     return [ dbc.Input(id=id, type="text",
-                       size="sm", className="m-1 d-inline-block w-auto", debounce=True, value=default) ]
+                       size="sm", class_name="m-0 p-0 d-inline-block w-auto", debounce=True, value=default) ]
```
aaron-kaplan, line 25: I have no idea what this stuff means. Do you?

controls.py
```diff
@@ -22,10 +22,10 @@ def Text(id, default):
...
 
 
-def Number(id, default, min=None, max=None, html_size=None):
+def Number(id, default, min=None, max=None, width="auto"):
```
aaron-kaplan, line 28: I believe this is what fixes the too-wide inputs on Chrome. Chrome and Firefox have different defaults for what happens when `html_size=None`.

aaron-kaplan:
I wasn't a big fan of the original, but I'm not convinced this is an improvement. I have nothing better to suggest, other than someday sitting down and really learning Bootstrap.

aaron-kaplan (approved):
Don't forget to test the other maprooms.

aaron-kaplan:
Does this depend on https://github.com/iridl/python-maprooms/pull/435 or is it an alternate implementation?

remicousin:
> Does this depend on #435 or is it an alternate implementation?

This here improves only controls.py while #435 uses fieldsets.py and doesn't use controls.py. So they don't relate

controls.py
remicousin: class_name allows to trigger css style in a more compact syntax. m and p are for [spacing](https://getbootstrap.com/docs/4.0/utilities/spacing/).

remicousin:
Looks as good as it gets on other Maprooms. Note the PickPoint function works in all Maprooms but flex_fcst but that's because we haven't stopped reading data in layout for these. We've only put all reading in CallBacks in flex_fcst. So it's a change that is bound to go away once we've propagated this behavior to all Maprooms. And maybe then I'll find a way to factor out this feature.

Issue #257, opened 2022-12-23
White div under colorbar so you can see it better


aaron-kaplan:
Especially the tick marks can be hard to see.

remicousin:
Resolved by PR #374

PR #374, opened 2023-08-24
adding white background to cs

remicousin:
This is meant to resolve issue #257 and it turns out to be simpler than that...

Might need fine automatic tuning as as it is, p-1 className doesn't insure the tick labels at the extremities of horizontal colorscales are undercovered by the background.

We might want some blackbox function/object for colorscales at some point...


aaron-kaplan:
For visual changes like this, it would be helpful if you could add a screenshot to the PR so the reviewer doesn´t have to check out and run the application to see what it looks like.

aaron-kaplan (approved):

PR #376, opened 2023-09-07
adapting to a case of reading pycptv2 nc files

remicousin:
This is not pretty but that's what we get for rushing work on top of unfinished work (aka cleanly separating files reading, dates selecting, dates formatting).

I gave up on getting the Target date from data so it's hard-coded. It's ok enough because the case reading pycptv2 nc files assumes one app per target date.

It seems that because CONFIG["y_transform"] is used as default value of a function, it needs to be defined. So it's still there in the config (will push that PR soon soon).

This should do for tomorrow's demo. Will have to clean it all up soon because next time we need to add a layer, it will likely be too ugly.

EMI tsv case still works.

### DIFF
--- a/enacts/flex_fcst/cpt.py
+++ b/enacts/flex_fcst/cpt.py
@@ -4,6 +4,8 @@ from datetime import datetime
 import numpy as np
 import cptio
 import xarray as xr
+from pathlib import Path
+
 
 def read_file(
     data_path,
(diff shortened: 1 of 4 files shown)


aaron-kaplan:
In maproom.py, please change 
```
APP.layout = layout.app_layout
```
to
```
APP.layout = layout.app_layout()
```
and also fix it wherever you copied it from. We had already found that pattern to be problematic in another maproom and changed it.

Making this change enables the maproom to start for me. Since it worked for you without that change, I'm guessing you have different library versions than me.

remicousin:
done. All the other apps have the ()

aaron-kaplan:
We have a discussion going by email about the directory structure.

@remicousin could you be more specific regarding the problem with getting target season from the data? Anything else that should be improved in the data representation?

remicousin:
The target date problem is probably not a problem from the data. It's just that I could only manage to get this "dask object" sort of message return instead of the data itself and I just ran out of time to go through. Now that the emergency has passed, I can probably go through this in this PR.

enacts/flex_fcst/maproom.py
```diff
@@ -330,20 +347,23 @@ def local_plots(marker_pos, start_date, lead_time):
...
-            lead_time,
-            pd.to_datetime(start_date),
-            CONFIG["target_period_length"],
-            CONFIG["time_units"],
-        )
-    elif CONFIG["targets"] is not None:
-        target_range = lead_time
+    if CONFIG["forecast_mu_file_pattern"] is None:
+        #target_range = predictions.target_range_formatting(fcst_mu["Ti"].compute(), fcst_mu["Tf"].compute()[0], "months")
+        target_range = "Oct-Dec"
```
remicousin, line 352: @aaron-kaplan , I have done all the possible combinations of compute, dt, data, values, indexing, I can't get it. the inputs of target_range_formatting need be dt... that's all I can figure out. Help please.

enacts/flex_fcst/cpt.py
```diff
@@ -102,3 +104,38 @@ def starts_list(
...
     start_dates = [i.strftime(format_out) for i in start_dates]
     return start_dates
+
+
+def read_pycptv2dataset(data_path):
+    for mm in (np.arange(11) + 1) :
+        monthly_path = Path(data_path + '/' + str(mm).zfill(2))
+        if monthly_path.exists():
+            try:
+                fcst_mu
```
aaron-kaplan, line 114: What is this for?

aaron-kaplan: never mind, i get it now.

enacts/flex_fcst/cpt.py
```diff
@@ -102,3 +104,38 @@ def starts_list(
     start_dates = sorted(set(start_dates)) #finds unique dates in the case there are files with the same date due to multiple lead times
     start_dates = [i.strftime(format_out) for i in start_dates]
     return start_dates
+
+
+def read_pycptv2dataset(data_path):
+    for mm in (np.arange(11) + 1) :
```
aaron-kaplan, line 110: FYI this ends at November.

Don't bother fixing, because I'm rewriting this whole function. I continue to get hangs that appear to be caused by multiple callbacks calling `open_mfdataset` concurrently, so I'm going to call `open_dataset` and then `concat` instead. (The difference is that `open_mfdataset` uses dask.)

aaron-kaplan:
I think I fixed the hanging problem (still don't understand why you didn't experience it), and I fixed reading the target date from the dataset.

I tested on Ethiopia, and I see that there are two starts and two targets, but only one valid target for each start. That's an unhelpful UI. From what I know of the applications (which is limited), mabye we should first have the user select target season, and we update the start date menu based on that. In the case of Ethiopia there would be only one available start.  (I guess this is outside the scope of this PR.)

remicousin:
Tested successfully on Lesotho, EMI cases before my changes (and after Aaron's)
and Tested successfully on the Lesotho, EMI and KMD after my changes.

Good to go for a merge?

aaron-kaplan (approved):
I was hoping we could get leads and targets from the tsv files so we don't need to configure them in the config file. That would allow us to simplify this code. But if you want to leave that for another day, I'm ok with that.

PR #381, opened 2023-10-10
Remic/takesxr

remicousin:
This intents to complete the 2 cases for reading pycptv2 data files and use them in ff maproom. The two cases being i) having a "rolling" forecast with dense S-L structure and ii) having a "season of interest" forecast where data depends only on S.

I used the kmd and lesotho cases of which config repo have been updated accordingly. The kmd case is not great because I have only one forecast...

I am not sure the reading scheme and time manipulations will work with a subseasonal forecast case still. But I'd think it can be made compatible again later.

Next step would be to rationalize the tsv case using same approach (read all data intended for the maproom in one time/function with limited config information). Then the reading functions should be well split from the maproom, and can see if the dates formatting functionality can be better split from the reading functions.

### DIFF
--- a/enacts/flex_fcst/cpt.py
+++ b/enacts/flex_fcst/cpt.py
@@ -150,7 +150,7 @@ def open_var(path, filepattern, SL_dense=True):
     ds = xr.concat(slices, 'T').swap_dims(T='S')
     if SL_dense:
         L = (ds["Ti"].dt.month - ds["S"].dt.month).squeeze()
-        L = (L + 6 * (L -np.abs(L)) / L).values
+        L = L.where(lambda x: x >=0, lambda x: x + 12).values
         ds = (ds
             .assign(Lead=lambda x: x["T"] - x["S"])
             .assign_coords({"L": L})
--- a/enacts/flex_fcst/cpt.py
+++ b/enacts/flex_fcst/cpt.py
@@ -124,9 +124,6 @@ def read_mpycptv2dataset(data_path):
         obs_slices.append(new_obs)
     fcst_mu = xr.combine_by_coords(mu_mslices)["deterministic"]
     fcst_var = xr.combine_by_coords(var_mslices)["prediction_error_variance"]
-    if target_count == 1:
-        fcst_mu = fcst_mu.squeeze(L, drop=True)
-        fcst_var = fcst_var.squeeze(L, drop=True)
     obs = xr.concat(obs_slices, "T")
     obs = obs.sortby(obs["T"])
     return fcst_mu, fcst_var, obs 
--- a/enacts/flex_fcst/cpt.py
+++ b/enacts/flex_fcst/cpt.py
@@ -106,30 +106,40 @@ def starts_list(
     return start_dates
 
 
-def read_mpycptv2dataset(data_path, SL_dense=True):
+def read_mpycptv2dataset(data_path):
     mu_mslices = []
     var_mslices = []
     obs_slices = []
+    target_count = 0
     for targets in Path(data_path).iterdir() :
-        new_mu, new_var, new_obs = read_pycptv2dataset(targets, SL_dense=SL_dense)
+        target_count = target_count + 1
+    for targets in Path(data_path).iterdir() :
+        new_mu, new_var, new_obs = read_pycptv2dataset(targets)
+        if target_count > 1 :
+            L = (((new_mu["Ti"].dt.month - new_mu["S"].dt.month).squeeze() + 12) % 12).values
+            new_mu = new_mu.assign_coords({"L": L}).expand_dims(dim="L")
+            new_var = new_var.assign_coords({"L": L}).expand_dims(dim="L")
         mu_mslices.append(new_mu)
         var_mslices.append(new_var)
         obs_slices.append(new_obs)
     fcst_mu = xr.combine_by_coords(mu_mslices)["deterministic"]
     fcst_var = xr.combine_by_coords(var_mslices)["prediction_error_variance"]
+    if target_count == 1:
+        fcst_mu = fcst_mu.squeeze(L, drop=True)
+        fcst_var = fcst_var.squeeze(L, drop=True)
     obs = xr.concat(obs_slices, "T")
     obs = obs.sortby(obs["T"])
     return fcst_mu, fcst_var, obs 
 
 
-def read_pycptv2dataset(data_path, SL_dense=True):
+def read_pycptv2dataset(data_path):
     mu_slices = []
     var_slices = []
     for mm in (np.arange(12) + 1) :
         monthly_path = Path(data_path) / f'{mm:02}'
         if monthly_path.exists():
-            mu_slices.append(open_var(monthly_path, 'MME_deterministic_forecast_*.nc', SL_dense=SL_dense))
-            var_slices.append(open_var(monthly_path, 'MME_forecast_prediction_error_variance_*.nc', SL_dense=SL_dense))
+            mu_slices.append(open_var(monthly_path, 'MME_deterministic_forecast_*.nc'))
+            var_slices.append(open_var(monthly_path, 'MME_forecast_prediction_error_variance_*.nc'))
     fcst_mu = xr.concat(mu_slices, "S")["deterministic"]
     fcst_mu = fcst_mu.sortby(fcst_mu["S"])
     fcst_var = xr.concat(var_slices, "S")["prediction_error_variance"]
@@ -144,17 +154,9 @@ def open_mfdataset_nodask(filenames):
     return xr.concat((xr.open_dataset(f) for f in filenames), 'T')
 
 
-def open_var(path, filepattern, SL_dense=True):
+def open_var(path, filepattern):
     filenames = path.glob(filepattern)
     slices = (xr.open_dataset(f) for f in filenames)
     ds = xr.concat(slices, 'T').swap_dims(T='S')
-    if SL_dense:
-        L = (ds["Ti"].dt.month - ds["S"].dt.month).squeeze()
-        L = L.where(lambda x: x >=0, lambda x: x + 12).values
-        ds = (ds
-            .assign(Lead=lambda x: x["T"] - x["S"])
-            .assign_coords({"L": L})
-            .expand_dims(dim="L")
-        )
     return ds
    
--- a/enacts/flex_fcst/cpt.py
+++ b/enacts/flex_fcst/cpt.py
@@ -106,7 +106,7 @@ def starts_list(
     return start_dates
 
 
-def read_mpycptv2dataset(data_path):
+def read_pycptv2dataset(data_path):
     mu_mslices = []
     var_mslices = []
     obs_slices = []
@@ -114,7 +114,7 @@ def read_mpycptv2dataset(data_path):
     for targets in Path(data_path).iterdir() :
         target_count = target_count + 1
     for targets in Path(data_path).iterdir() :
-        new_mu, new_var, new_obs = read_pycptv2dataset(targets)
+        new_mu, new_var, new_obs = read_pycptv2dataset_single_target(targets)
         if target_count > 1 :
             L = (((new_mu["Ti"].dt.month - new_mu["S"].dt.month).squeeze() + 12) % 12).values
             new_mu = new_mu.assign_coords({"L": L}).expand_dims(dim="L")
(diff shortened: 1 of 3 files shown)


enacts/flex_fcst/cpt.py
```diff
@@ -106,17 +106,35 @@ def starts_list(
     return start_dates
 
 
-def read_pycptv2dataset(data_path):
+def read_mpycptv2dataset(data_path, SL_dense=True):
```
aaron-kaplan, line 109: Instead of passing this SL_dense value around, can't we just infer that SL_dense is true when the data has an L dimension, and false when it doesn't?

remicousin: read_mpycptv2dataset creates the L dimension so from there down the road in the maprooms, we could check whether the data has an L and choose what to do. Or if you have some other workflow in mind, let me know.

remicousin: @aaron-kaplan , as things stand, and with changes to the config file for which there is a PR in bitbucket too, The app works well with 2 sets of "SL-dense" forecasts (I copied the 2 forecasts Peterson has in their server). So they could have that for their workshop starting Tuesday. I am not saying we should merge as is, but we have little time before that workshop, and I also haven't started to look at IRI Strategy dcument that I would like to make sure I have time to look at today.

So I could do some important but not time-consuming here before merging, and we could put the rest as Issues.

aaron-kaplan: Replied on the config PR https://bitbucket.org/iridl/python_maproom_kmd/pull-requests/7/new-config-system-to-cover-sl_dense-true?link_source=email#comment-443730056

enacts/flex_fcst/cpt.py
```diff
@@ -126,8 +144,17 @@ def open_mfdataset_nodask(filenames):
...
 
 
-def open_var(path, filepattern):
+def open_var(path, filepattern, SL_dense=True):
     filenames = path.glob(filepattern)
     slices = (xr.open_dataset(f) for f in filenames)
     ds = xr.concat(slices, 'T').swap_dims(T='S')
+    if SL_dense:
+        L = (ds["Ti"].dt.month - ds["S"].dt.month).squeeze()
+        L = (L + 6 * (L -np.abs(L)) / L).values
```
aaron-kaplan, line 153: Wow, creative way of writing `L = L % 12`

enacts/flex_fcst/cpt.py
```diff
@@ -126,8 +144,17 @@ def open_mfdataset_nodask(filenames):
...
+        ds = (ds
+            .assign(Lead=lambda x: x["T"] - x["S"])
```
aaron-kaplan, line 155: Do we really need both `L` and `Lead`?

Don't we need to take modulus 12 here as you did in line 153?

enacts/flex_fcst/maproom.py
```diff
@@ -229,8 +244,8 @@ def target_range_options(start_date):
    Input("lead_time","options"),
 )
 def write_map_title(start_date, lead_time, lead_time_options):
-    if CONFIG["forecast_mu_file_pattern"] is None:
-        fcst_mu, fcst_var, obs = cpt.read_pycptv2dataset(DATA_PATH)
+    if CONFIG["forecast_mu_file_pattern"] is None and ~CONFIG["SL_dense"]:
```
aaron-kaplan, line 247: should say `and not CONFIG["SL_dense"]`

`~` is bitwise OR, not logical OR. See https://stackoverflow.com/questions/3845018/boolean-operators-vs-bitwise-operators

aaron-kaplan: You were probably misled by seeing `~` used for logical negation in xarray code. Xarray (numpy?) has no boolean type, so we use bitwise logic because it's almost the same, but subject to the disadvantages mentioned in the Stack Overflow answers.

enacts/flex_fcst/cpt.py
remicousin: We probably don't need both L and Lead. I made that choice at some point. I think we discussed it at some other point. Nothing decisive emerged I think from those interactions so far. I think my motivation was that L is somewhat arbitrary and follows conventions, and also can not be a time object if we want to express it in months because months distances don't exist since months themselves have inconsistent lengths...

So I guess I felt the need to provide an object (Lead) that has the missing information that is simplified in L. Thus it is the actual distance (in time units) between T and S (not just of their months so there is no more a "modulo" 12 that is not one). But unlike L, it is not independent of the year of S because on leap years, if there is 28 Feb between S and T, then that distance will change... and so it's not a good coordinate system... hence L... and the loop buckles up...

No strong feeling with this... It's touching sensitive matters that need better attention than the constant rush we're in...

enacts/flex_fcst/cpt.py
remicousin: haha! yes indeed... but that's because it's not modulo. -2 % 12 is -2, not 10. The goal here is not to change L when positive, and add 12 when it is negative.

enacts/flex_fcst/cpt.py
remicousin: ok. I fixed the 2ndary problems and now I am going to see what I can do free the config file from the boolean SL_dense.

enacts/flex_fcst/cpt.py
aaron-kaplan: Yeah, also I want to use CF conventions for cell bounds instead of Kyle's homegrown representation. I don't see any problem with Kyle's representation, but there is a standard adopted by the community and this isn't it.

aaron-kaplan: You didn't answer me about taking the modulus, but now I see that it's not necessary because in this case we're subtracting dates, not month numbers.

enacts/flex_fcst/cpt.py
aaron-kaplan: Oh right, `L = (L + 12) % 12`

enacts/flex_fcst/cpt.py
remicousin: I pushed something to that effect. Tested on _kmd and _lesotho for which I also pushed the suppression of SL_dense from the config.

enacts/flex_fcst/cpt.py
remicousin: The revisit of SL_dense made Lead even more superfluous and is gone in the last push.

enacts/flex_fcst/cpt.py
remicousin: Haha! oh boy... I used that instead of where now.

enacts/flex_fcst/maproom.py
```diff
@@ -634,9 +660,13 @@ def fcst_tiles(tz, tx, ty, proba, variable, percentile, threshold, start_date, l
     # Reading
     
     if CONFIG["forecast_mu_file_pattern"] is None:
-        fcst_mu, fcst_var, obs = cpt.read_pycptv2dataset(DATA_PATH)
+        fcst_mu, fcst_var, obs = cpt.read_mpycptv2dataset(DATA_PATH)
         fcst_mu = fcst_mu.sel(S=start_date)
         fcst_var = fcst_var.sel(S=start_date)
+        if "L" in fcst_mu.dims:
+            fcst_mu = fcst_mu.sel(L=int(lead_time))
+            fcst_var = fcst_var.sel(L=int(lead_time))
```
aaron-kaplan, line 668: Instead of converting these here, mark them as int in line 657, as we've already done for tx, ty, tz,

enacts/flex_fcst/maproom.py
```diff
@@ -634,9 +660,13 @@ def fcst_tiles(tz, tx, ty, proba, variable, percentile, threshold, start_date, l
...
+        obs = obs.where(obs["T"].dt.month == fcst_mu["T"].dt.month, drop=True)
         is_y_transform = False
     else:
         fcst_mu, fcst_var, obs, hcst = read_cptdataset(lead_time, start_date, y_transform=CONFIG["y_transform"])
```
aaron-kaplan, line 672: You didn't change this to `int(lead_time)`. Does that mean `read_cptdataset` expects a string, not an int? That doesn't seem right.

enacts/flex_fcst/cpt.py
```diff
@@ -106,6 +106,32 @@ def starts_list(
...
+            new_mu = new_mu.assign_coords({"L": L}).expand_dims(dim="L")
+            new_var = new_var.assign_coords({"L": L}).expand_dims(dim="L")
+        mu_mslices.append(new_mu)
+        var_mslices.append(new_var)
+        obs_slices.append(new_obs)
+    fcst_mu = xr.combine_by_coords(mu_mslices)["deterministic"]
+    fcst_var = xr.combine_by_coords(var_mslices)["prediction_error_variance"]
+    if target_count == 1:
+        fcst_mu = fcst_mu.squeeze(L, drop=True)
+        fcst_var = fcst_var.squeeze(L, drop=True)
```
aaron-kaplan, line 129: I don't understand this. If `target_count == 1` then we didn't add `L` in the first place (line 118),

enacts/flex_fcst/cpt.py
```diff
@@ -106,6 +106,32 @@ def starts_list(
     return start_dates
 
 
+def read_mpycptv2dataset(data_path):
```
aaron-kaplan, line 109: Rename this `read_pycptv2dataset` and rename the current `read_pycptdataset` to `read_pycptdataset_one_season`?

enacts/flex_fcst/cpt.py
```diff
@@ -106,6 +106,32 @@ def starts_list(
...
+    mu_mslices = []
+    var_mslices = []
+    obs_slices = []
+    target_count = 0
+    for targets in Path(data_path).iterdir() :
+        target_count = target_count + 1
+    for targets in Path(data_path).iterdir() :
+        new_mu, new_var, new_obs = read_pycptv2dataset(targets)
+        if target_count > 1 :
```
aaron-kaplan, line 118: In hindsight this is not a great way to distinguish dense from sparse, because there can be sparse datasets with multiple seasons. E.g. Ethiopia has two rainy seasons and we generate three starts for each. Too late to change now, we'll do it in a future version.

When generating dense forecasts we should have PyCPT put all leads for a given S in the same file. Then when reading we don't have to add L, it's already there.

enacts/flex_fcst/cpt.py
remicousin: I used _target rather than _season in case we are able to adapt this to subseasonal, where targets are week(s), days...

enacts/flex_fcst/cpt.py
remicousin: True, but in the Ethiopian case example given, we would make one app per target season and so it will be on the data manager to organize their data directory structure accordingly. I think once we have split up data cataloguing and app building, those tedious petty problems will go away.

Is it one process that generates dense forecasts? or is it multiple runs of different pyCPT configs? Because the obs aren't the same, possibly the predictor is not the same either. If so those files are not written through the same process and that would mean post-processing of the multiple files to get one... which is basically what we're doing here already (only again we're meddling what's needed for an app with what's needed to properly catalogue data). This being said, you could possibly have pyCPT generate anyways a 1-point L dim always, then cataloguing could decide whether to use it (dense case) or not

enacts/flex_fcst/maproom.py
remicousin: That sounds likely but read_cptdataset is for the .tsv case that has many other problems we want to resolve so let's leave that to another PR.

enacts/flex_fcst/maproom.py
remicousin: I think can't do that for now because read_cptdateset accepts lead_time as the lead time as a number, or as the target season as a string of characters, and so if the first case could be cast in a int right away, the second case can't. That was because the tsv case had written in the filename either the numerical lead or the target season (e.g. Oct-Dec)...

So I am afraid we'll have to stick with that until we revisit the tsv reading case.

aaron-kaplan:
Will any existing published maprooms need to be modified to work with this version? If so, please explain what needs to change in CHANGELOG.md.

remicousin:
> Will any existing published maprooms need to be modified to work with this version? If so, please explain what needs to change in CHANGELOG.md.

Supposedly no, because all published one run into the tsv case that has been untouched (or possibly improved but not requiring modification).

remicousin:
and assuming the lesotho version is not yet a published one

aaron-kaplan (approved):

PR #382, opened 2023-10-11
Icon for KMD

remicousin:
Is missing and thus their Python Maprooms return a broken image where it's meant to be.


aaron-kaplan (approved):

aaron-kaplan (changes requested):
Wait, no, shouldn't this be in python_maprooms_kmd?

aaron-kaplan:
I see that indeed we're putting all the logos  in the container. That sucks because it means I have to build and publish a new version of the container each time we set up a new partner. We should instead
- Add the icon file to python_maproom_kmd
- Modify the python_maproom_kmd Dockerfile to COPY the icon file into the container to a directory where it will be picked up by dash/flask. I don't know offhand what location that is. @xchourio can you work on this?

remicousin:
So I think we are discarding this PR right? since you set up so that can be added to the country config repo.

PR #383, opened 2023-10-11
rainfall can't be <0

remicousin:
Somehow this was removed [here](https://github.com/iridl/python-maprooms/commit/b5b0d6cc487906fb23fdf34ed0923fb0fb69bc89). Better do it outside the graph so I can check whether it's Precip or not. This is probably not the best way to check on that though... The variable "name" is in the config file though... Maybe we'll need that to be a dictionary or nested yaml with some variable attributes (like name, min/max values if any...)...

not sure...

### DIFF
--- a/enacts/flex_fcst/maproom.py
+++ b/enacts/flex_fcst/maproom.py
@@ -416,6 +416,9 @@ def local_plots(marker_pos, start_date, lead_time):
             "scale": np.sqrt(fcst_var),
         },
     ).rename("fcst_ppf")
+    if CONFIG["variable"] == "Precipitation":
+        fcst_ppf = fcst_ppf.where(lambda x: x >= 0)
+        obs_ppf = obs_ppf.where(lambda x: x >= 0)
     poe = fcst_ppf["percentile"] * -1 + 1
     # Graph for CDF
     cdf_graph = pgo.Figure()
--- a/enacts/flex_fcst/maproom.py
+++ b/enacts/flex_fcst/maproom.py
@@ -417,8 +417,8 @@ def local_plots(marker_pos, start_date, lead_time):
         },
     ).rename("fcst_ppf")
     if CONFIG["variable"] == "Precipitation":
-        fcst_ppf = fcst_ppf.where(lambda x: x >= 0)
-        obs_ppf = obs_ppf.where(lambda x: x >= 0)
+        fcst_ppf = fcst_ppf.clip(min=0)
+        obs_ppf = obs_ppf.clip(min=0)
     poe = fcst_ppf["percentile"] * -1 + 1
     # Graph for CDF
     cdf_graph = pgo.Figure()
@@ -653,6 +653,8 @@ def fcst_tiles(tz, tx, ty, proba, variable, percentile, threshold, start_date, l
             percentile,
             kwargs={"loc": obs_mu, "scale": obs_stddev},
         )
+        if CONFIG["variable"] == "Precipitation":
+            obs_ppf = obs_ppf.clip(min=0)
     else:
         obs_ppf = threshold
     # Forecast CDF


aaron-kaplan:
Seems like Student's t, being symmetric, is the wrong distribution to use for modeling a quantity that can't be negative. Have you ever had that conversation with Simon?

Why is the variable name a config attribute? Lots of classic maprooms have multiple variables, with a dropdown to select one. Don't we want to support that here?

I definitely don't like using the variable's English name to decide whether it's bounded by zero or not, but the solution depends on the answer to the previous question.

remicousin:
I've had conversations about what to do about bounded variables with probably everybody (Simon, Michael, Nachi, Andy...). It's problematic also for Normal distribution, and ELR... all these extend to +/-infinity. I am not discussing with them the validity of choosing those statistical models... they've been doing that for decades. I believe the general assumption is that the seasonal quantities are no more bounded (or rather that their domain of definition is pushed far away enough from the bound -- here 0). Anyways, for the purpose of showing results in figures, clipping at 0 (or something equivalent) is what they suggested to do.

So far, the variable name was a configurable attribute only for cosmetics reasons (to print titles and labels in the app), and I think only so because the name that can be found with files can be too obscure (PRCP, ERF...etc.) for human eyes looking at an app.

Historically, we haven't put more than one variable in those (flexible) forecast Maprooms... We might want to keep it this way: it's painful enough to give a common structure to put together a "Precipitation" set of forecasts, it sounds daunting to try and add another variable to the set... I'd rather have a set of py files that can be configured to use this or that variable. I don't have a strong opinion about that... there's no guarantee that with a precip forecast will come a temperature forecast, or vice versa... or others.

aaron-kaplan:
I'm ok with making the user-visible name of the variable configurable, but then it doesn't make sense to compare it to the hard-coded value `Precipitation` in the code, because it might be set to `Rainfall` at one site and `تساقط` at another.

Two alternative solutions come to mind:
- In other contexts we've discussed providing min and max values in netcdf/zarr metadata attributes. That might help here, but then again, there's a difference between the observed minimum and the theoretical/physical minimum possible value, so we might confuse ourselves there if we're not careful.
- We should probably ensure that all variables are annotated with [CF standard names](https://cfconventions.org/Data/cf-standard-names/47/build/cf-standard-name-table.html). Then we could have a lookup table of minimum and maximum possible values for each quantity.

I can ensure that future versions of pycpt include CF standard_name metadata in their output, but the current version is already out there in the wild and it's going to be hard/impossible to get everyone who sends us forecasts to upgrade. We need to have the flexibility to add metadata on our end when setting up the application. In other words, we need a data catalog that we can edit. I'm sure you're not surprised by that finding.

As we've discussed before, if we convert everything to zarr, then the zarrification script becomes a place we can add metadata. We have so far avoided converting pycpt forecasts to zarr because (a) there was nothing in particular forcing us to, and (b) it makes the process of adding a new forecast to an existing dataset simpler. But maybe it's time to zarrify it.

remicousin:
Agreed with all that is said there. Except maybe that it's the object we output in our calculations that has a lower bound of 0. mu has a lower bound of 0 for the same reason, but var has a lower bound of 0 for other reasons. The lower bound of the obs is 0 too and is the same 0 as the calculated distribution. But we can calculate distribution just from mu and var and would still want to apply a 0 lower bound.

In other words, I am unsure where that 0 should come from. It seems to be appropriately coming from a combination of the data attributes and the function we apply to it.

aaron-kaplan:
Is this even the right thing to do? Given that we're using a distribution that gives non-zero probability to negative values, shouldn't we display it as such, rather than hiding it and displaying a PDF whose integral is not 1?

aaron-kaplan (approved):

remicousin:
I am pretty sure this is what Mike Tippett suggested we do, when we started producing those analyses/viz'. I think he even use the term probability mass at a point or something like that. I think it's legit for the maths/viz' done in this app, there are likely other maths that would not be ok with it.

remicousin:
Rebasing is going to be absurdly complex for this however simple change given the big recent changed made to flex_fcst so I am just going to close this PR and make a new one from where master is now.

PR #389, opened 2023-10-20
Quick fix for problem of irregular coordinates

aaron-kaplan:
We get different numbers in the same column depending on what other columns are in the table, e.g.

image
vs.
image

This is because when we put DataArrays with non-aligned X and Y coordinates into the same Dataset, the resulting X and Y coordinates are no longer regularly spaced, violating an assumption we made in pingrid. The quick fix I'm doing here is to drop coordinate values where the variable is all NaN, to get back to the original coordinate, but in the longer run we should stop putting non-aligned variables into the same Dataset. If nothing else, it's a waste of space, because we have to store a lot of NaNs.

@remicousin you don't need to review the change in context, but you should be aware of the issue.

I have a test server running on shortfin01:9998, waiting for Jeff to open that port so we can each have our own test server.

### DIFF
--- a/fbfmaproom/tests/test_pingrid.py
+++ b/fbfmaproom/tests/test_pingrid.py
@@ -134,16 +134,16 @@ def test_average_over():
 #     assert v.item() == 5.
 
 def test_average_over_nans():
-    data = [[1, np.nan], [2, np.nan]]
+    data = [[1, np.nan], [np.nan, 2.,]]
     da = xr.DataArray(
         data=data,
         coords={
-            'lon': [0., 1.],
-            'lat': [0., 1.],
+            'lon': [0., 0.1],
+            'lat': [0., 0.1],
         },
     )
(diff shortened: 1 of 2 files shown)


remicousin:
In Ingrid, for variables of a dataset that have dimensions with the same name, those dimensions' coordinates should be the same in order to avoid ambiguity on which dimension is called when called, and allow functions operating on such dimension to be applicable across those variables. I put conditional tense because technically, you can not do that, you just expose yourself to those difficulties.

xchourio (approved):

aaron-kaplan:
I tried putting in an assertion that verifies that the coordinate is regularly spaced, but it failed because of floating point rounding differences (e.g. 49.999999 vs. 50.000001). Then I tried using `np.isclose` with default tolerances, but that failed too. So if we're going to make such an assertion, we need to decide how much difference is allowable, and that requires more thought than I'm capable of right now.

PR #392, opened 2023-11-06
Madagascar OND v2

xchourio:


aaron-kaplan:
Please update the label and description of the v1 forecast dataset too. (Don't change the yaml key `pnep`, that has to stay the same, but you can change the values of the `label` and `description` attributes.)

xchourio:
> Please update the label and description of the v1 forecast dataset too. (Don't change the yaml key `pnep`, that has to stay the same, but you can change the values of the `label` and `description` attributes.)

Done!

aaron-kaplan (approved):

aaron-kaplan:
When it's time to push the change to the public server (after FIST approves), please make the same change in a dlconfig PR.  https://bitbucket.org/iridl/dlconfig/src/master/ansible/configs/fbfmaproom2/fbfmaproom.yaml

aaron-kaplan:
We never got all the data files we asked for.

PR #393, opened 2023-11-09
WSRI+NDVI ond,djf for lesotho

xchourio:


xchourio:
@aaron-kaplan I tested it in a local run and it displayed both variables well, but I haven't updated the 'public test server' yet because the FIST team shared that link with partner institutions while waiting for their approval for other countries' maprooms. So, I avoided making the update.

aaron-kaplan:
Please share it on your internal test server on shortfin01.

The external test server and the shortfin dev setup share the same storage, but not the same config files, so you can add a new dataset to the shortfin01 server without making it visible to the partner.

aaron-kaplan (approved):
Looks good,  pending FIST's apporoval.

xchourio:
The comment from Nitin Magima in Trello leads me to understand that we have FIST approval. I'll merge it.

PR #400, opened 2024-02-01
instructions for running enactstozarr.py at partner's

remicousin:


enacts/README.md
```diff
@@ -100,6 +100,25 @@ For final releases of the image, use the `release_container_image` script (no pa
...
+    sudo docker run \
+      --rm \
+      -u $(id -u) \
+      -v /data/datalib/data:/data/datalib/data:rw \
+      -v /usr/local/datalib/build/python_maproom/config.yaml:/app/config.yaml \
+      -e CONFIG=/app/config.yaml \
+      iridl/enactsmaproom \
+      python enactstozarr.py
+
+Should you need to run a modified version of the enactstozarr script, add the option:
```
aaron-kaplan, line 116: Maybe rephrase "If you need to test a modified version...". I don't want this to be misread as saying that it's ok to use some other script that isn't kept in git. But sometimes we need to make changes to the script in git, and it's nice to be able to test them before committing and building a new container.

aaron-kaplan (approved):

enacts/README.md
remicousin: I just removed it

PR #401, opened 2024-02-03
Get logo from flask static dir

aaron-kaplan:
Just one copy of the logo instead of one per page.

This is lipstick on a pig. We still have one copy of the favicon, style.css, and all the javascript libraries for each page. Need a more general way for pages to share resources. But that's for another day.


aaron-kaplan:
Already deployed to the Malawi server so I guess I'll merge it.

aaron-kaplan:
Strictly speaking, no change to existing configurations is necessary, because I left the country-specific icons in the image, at a location where they will be found by the application. But I do want to encourage us to move those built-in images to the country-specific python_maproom_country repos next time we change them, so I'll add a note.

aaron-kaplan:
@remicousin please review again.

remicousin (approved):
I figured after I sent this that worst case scenario would be a broken image on the app so nothing really dramatic (or different from what is happening now).

PR #402, opened 2024-02-05
Update container build instructions

aaron-kaplan:
When reviewing your recent PR, I noticed that the instructions for publishing a new container were inappropriately copied from fbfmaproom2.


remicousin (approved):
I approve even though I have no clue

PR #403, opened 2024-02-13
Djibouti SPI for MAM, OND

aaron-kaplan:


fbfmaproom/fbf-update-data.py
```diff
@@ -325,6 +325,14 @@
...
+    ),
+    (
+        "djibouti/spi-ond",
+        "http://iridl.ldeo.columbia.edu/SOURCES/.UCSB/.CHIRPS/.v2p0/.daily-improved/.global/.0p25/.prcp/X/41.625/43.375/RANGE/Y/10.875/12.875/RANGE/T/(Oct-Dec)/VALUES/monthlyAverage/30/mul/a%3A/3/gamma3par/pcpn_accum/gmean/gsd/gskew/pzero/3/gammaprobs/3/gammastandardize/T//pointwidth/3/def//defaultvalue/%7Blast%7D/def/-1./shiftGRID/T/first/pointwidth/1/sub/add/last/RANGE//long_name/(Standardized%20Precipitation%20Index)/def/%3Aa%3A/T/3/runningAverage/T/12/splitstreamgrid/0./flaggt/%5BT2%5Daverage/1./3./div/flaggt/1./masklt/%5BT%5D/REORDER/CopyStream/%3Aa/mul/DATA/-3/3/RANGE//name//spi/def/T/(Oct-Dec%201991)/last/12/RANGESTEP/"
+    ),
```
remicousin, line 335: I would scrap the T VALUES at the beginning because  I am really not sure the spi functions would like that, and let the final RANGESTEP do the work of selecting the yearly seasons of interest.
On the other hand, the dry mask could probably be done on the season of interest only and save a bit some computing resources.
There is a more proper way than 30 mul to convert monthly mm/day into monthly mm totals. But if this is some legacy code, that is probably fine enough.
The middle T RANGE seems superfluous.

So all in all it would end up as:

http://iridl.ldeo.columbia.edu/SOURCES/.UCSB/.CHIRPS/.v2p0/.daily-improved/.global/.0p25/.prcp/X/41.625/43.375/RANGE/Y/10.875/12.875/RANGE/monthlyAverage/T/(days%20since%201960-01-01)/streamgridunitconvert/T/differential_mul/T/(months%20since%201960-01-01)/streamgridunitconvert/a%3A/3/gamma3par/pcpn_accum/gmean/gsd/gskew/pzero/3/gammaprobs/3/gammastandardize/T//pointwidth/3/def//defaultvalue/%7Blast%7D/def/-1./shiftGRID/T/(Mar-May%201991)/last/12/RANGESTEP//long_name/(Standardized%20Precipitation%20Index)/def/%3Aa%3A/T/(Mar-May)/seasonalAverage/0./flaggt/%5BT%5Daverage/1./3./div/flaggt/1./masklt/%3Aa/mul/DATA/-3/3/RANGE//name//spi/def/#expert

remicousin (approved):
I proposed an improvement but the existing code is also fine

fbfmaproom/fbf-update-data.py
aaron-kaplan: Which step are you concerned would be affected by T VALUES? I just tried downloading the results with and without that reduction, and the results differ by at most 10e-6 (and of course the early reduction in size makes it much faster).

aaron-kaplan: (I tried that experiment on the original URL, not your suggestion.)

remicousin: spi Fortran functions do their own splitting of seasons/years and I am wondering how it likes it if the data is no more evenly spaced...

aaron-kaplan: Tried again with your suggestion and the results are significantly different when I add the restriction on T. I'll leave it out.

fbfmaproom/fbf-update-data.py
```diff
@@ -327,11 +327,11 @@
...
+        "http://iridl.ldeo.columbia.edu/SOURCES/.UCSB/.CHIRPS/.v2p0/.daily-improved/.global/.0p25/.prcp/X/41.625/43.375/RANGE/Y/10.875/12.875/RANGE/monthlyAverage/T/(days%20since%201960-01-01)/streamgridunitconvert/T/differential_mul/T/(months%20since%201960-01-01)/streamgridunitconvert/a%3A/3/gamma3par/pcpn_accum/gmean/gsd/gskew/pzero/3/gammaprobs/3/gammastandardize/T//pointwidth/3/def//defaultvalue/%7Blast%7D/def/-1./shiftGRID/T/(Oct-Nov%201991)/last/12/RANGESTEP//long_name/(Standardized%20Precipitation%20Index)/def/%3Aa%3A/T/(Oct-Nov)/seasonalAverage/0./flaggt/%5BT%5Daverage/1./3./div/flaggt/1./masklt/%3Aa/mul/DATA/-3/3/RANGE//name//spi/def/"
     ),
```
remicousin, line 335: you used Oct-Nov as opposed to Oct-Dec to calculate spi-ond (2 locations in the code).

remicousin (changes requested):

fbfmaproom/fbf-update-data.py
aaron-kaplan: Fixed, thanks.

Code review is probably the only way to catch errors like this. The results of the incorrect URL are entirely plausible.

remicousin: Right... maybe ChatGPT would have pointed it out? but indeed ON-SPI and OND-SPI are objects of same nature and are both correct entities; plus, they probably aren't that different so yeah, would be hard to catch... but also wouldnot be making a terrible mistake...

remicousin (approved):

fbfmaproom/fbf-update-data.py
aaron-kaplan: On second thought, the midpoint of ON is different from the midpoint of OND, which should have messed up the alignment of table rows. Now I have to investigate why I didn't notice that. I may not have tested what I thought I tested.

remicousin: because the dry mask is timeless

aaron-kaplan:
@remicousin Is there a good reason that we're using CHIRPS daily to calculate seasonal SPI? Wouldn't it be faster to use monthly?

remicousin:
I can't think of a good reason. It would definitely be faster to use monthly.

aaron-kaplan:
I tried it and it was actually slower. Turns out the monthly one has a higher spatial resolution than daily.

remicousin:
ah yes... you do have a regridded version (even 2) at 1-degree.... but maybe that is then two coarse for your purpose?

https://iridl.ldeo.columbia.edu/SOURCES/.UCSB/.CHIRPS/.v2p0/.monthly/.global/

aaron-kaplan:
NMME resolution is exactly what we need, thanks. And I see that the regridded ones are cached, so that should be good.

I just need to change the dataset and remove `monthlyAverage` and the rest can stay the same, right?

remicousin:
monthlyAverage and the following units conversion

PR #404, opened 2024-02-20
Remic/entozud

remicousin:
I did a bit of cleaning by factoring out some functionality, removing all CONFIG entries from being inside functions, adding Documentation, and introducing system arguments rather than a loop on vars and time_res.

All of that to have convert either create a new zarr, or append to an existing one. At present this is assuming that there is no gap between the new set and the old set... Not sure what happens if there is. Maybe that's where the region case come into play? (I haven't looked into that).

I only choose to use to_zarr because it's what we were using in the first place. I am wondering if the 2 cases (create/append) could be even a lot closer if appending to nothing happens to create... or in the case reading the zarr store returns nothing, then make an empty zarr store and that's all the 2 cases differ from...

### DIFF
--- a/enacts/enactstozarr.py
+++ b/enacts/enactstozarr.py
@@ -217,8 +217,8 @@ def convert(
         last_T_nc = filename2datetime(netcdf[-1], time_res=time_res)
         if last_T_nc.strftime("%Y%m%d") < last_T_zarr.dt.strftime("%Y%m%d").values :
             print((
-                f'nc set ({last_T_nc.strftime("%Y%m%d")}) ends before '
-                f'zarrs ({last_T_zarr.dt.strftime("%Y%m%d").values})'
+                f'nc set ({last_T_nc}) ends before '
+                f'zarrs ({last_T_zarr})'
             ))
             print("Not changing existing zarr")
         elif (
@@ -230,8 +230,8 @@ def convert(
         else :
             print((
                 f'appending nc to zarr from '
-                f'{last_T_zarr.dt.strftime("%Y%m%d").values}'
-                f' to {last_T_nc.strftime("%Y%m%d")}'
+                f'{last_T_zarr.values}'
+                f' to {last_T_nc}'
             ))
             nc2xr(
                 netcdf,
--- a/enacts/enactstozarr.py
+++ b/enacts/enactstozarr.py
@@ -216,10 +216,7 @@ def convert(
         last_T_zarr = current_zarr["T"][-1]
         last_T_nc = filename2datetime(netcdf[-1], time_res=time_res)
         if last_T_nc.strftime("%Y%m%d") < last_T_zarr.dt.strftime("%Y%m%d").values :
-            print((
-                f'nc set ({last_T_nc}) ends before '
-                f'zarrs ({last_T_zarr})'
-            ))
+            print(f'nc set ({last_T_nc}) ends before zarrs ({last_T_zarr})')
             print("Not changing existing zarr")
         elif (
             last_T_nc.strftime("%Y%m%d") ==
@@ -228,11 +225,7 @@ def convert(
             print(f'both sets end same date: {last_T_nc}')
             print("Not changing existing zarr")
         else :
-            print((
-                f'appending nc to zarr from '
-                f'{last_T_zarr.values}'
-                f' to {last_T_nc}'
-            ))
+            print(f'appending nc to zarr from {last_T_zarr.values} to {last_T_nc}')
             nc2xr(
                 netcdf,
                 var_name,
--- a/enacts/enactstozarr.py
+++ b/enacts/enactstozarr.py
@@ -6,7 +6,6 @@ import xarray as xr
 import datetime as dt
 from pathlib import Path
 import pingrid
-import pandas as pd
 from functools import partial
 
 
--- a/enacts/enactstozarr.py
+++ b/enacts/enactstozarr.py
@@ -7,16 +7,80 @@ import datetime as dt
 from pathlib import Path
 import pingrid
 from functools import partial
+import calc
 
 
 CONFIG = pingrid.load_config(os.environ["CONFIG"])
-TIME_RES = sys.argv[1] #e.g. daily, or dekadal -- check in your config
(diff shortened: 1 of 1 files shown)


enacts/enactstozarr.py
```diff
@@ -43,49 +120,148 @@ def regridding(data, resolution):
...
+    --------
+    calc.read_zarr_data, filename2datetime, nc2xr, xarray.Dataset.to_zarr
+    """
+    print(f"converting files for: {time_res} {var_name}")
+    netcdf = list(sorted(Path(input_path).glob("*.nc")))
+    if Path(output_path).is_dir() :
+        current_zarr = calc.read_zarr_data(output_path)
+        last_T_zarr = current_zarr["T"][-1]
+        last_T_nc = filename2datetime(netcdf[-1], time_res=time_res)
+        if last_T_nc.strftime("%Y%m%d") < last_T_zarr.dt.strftime("%Y%m%d").values :
```
aaron-kaplan, line 219: Why the strftime? Doesn't this work?

```
if last_T_nc < last_T_zarr.values
```

enacts/enactstozarr.py
```diff
@@ -43,49 +120,148 @@ def regridding(data, resolution):
...
+            print((
+                f'nc set ({last_T_nc.strftime("%Y%m%d")}) ends before '
+                f'zarrs ({last_T_zarr.dt.strftime("%Y%m%d").values})'
+            ))
+            print("Not changing existing zarr")
+        elif (
+            last_T_nc.strftime("%Y%m%d") ==
+            last_T_zarr.dt.strftime("%Y%m%d").values
```
aaron-kaplan, line 227: likewise here

enacts/enactstozarr.py
```diff
@@ -43,49 +120,148 @@ def regridding(data, resolution):
...
+        ) :
+            print(f'both sets end same date: {last_T_nc}')
+            print("Not changing existing zarr")
+        else :
+            print((
+                f'appending nc to zarr from '
+                f'{(last_T_zarr+np.timedelta64(1, "D")).dt.strftime("%Y%m%d").values}'
```
aaron-kaplan, line 234: Thinking ahead to using this function for other datasets besides enacts... 1 day isn't the minimum possible increment, e.g. CFSv2 is 6-hourly. I think you could just make it simple and say "from old-end-time to new-end-time".

enacts/enactstozarr.py
```diff
@@ -41,55 +120,148 @@ def regridding(data, resolution):
...
+                f' to {last_T_nc.strftime("%Y%m%d")}'
```
aaron-kaplan, line 235: Even when creating a string to be printed, I'm not sure this strftime is helpful. You can just print a datetime, and the default string representation is more readable than %Y%d%d with no delimiters.

aaron-kaplan (commented):
In https://bitbucket.org/iridl/subx-cloud/src/master/zarrify.py I took the approach of creating an empty store and then writing into regions of it. (The motivation in that case was to parallelize writes.) It might be more complicated that what's needed in this case.

It would be nice to be able to either (a) skip opening any netcdf files whose data are already present in the zarr, for speed, or (b) compare pre-existing zarr data to what's in the netcdf files, for extra safety when you're not in a hurry. But that can wait for a subsequent version.

enacts/enactstozarr.py
remicousin: What you propose complains that it can't compare int with datetime.datetime
int is last_T_zarr.values
I had tried all sorts of things and I always run into comparing objects of different types...

remicousin:
I don't understand, I pushed with
git push origin remic/entozud
but nothing is changing here... What does that mean?

enacts/enactstozarr.py
aaron-kaplan: Hmm. Before posting that, I actually opened some random zarr dataset and tried `ds['T'][-1].values` and got a np.datetime64. Could you point me to the dataset you're testing this on so I can try to figure out what's different?

remicousin: these are the 2 objects we try to compare:

converting files for: daily precip
1983-05-31T00:00:00.000000000
<class 'numpy.datetime64'>
1983-05-31 00:00:00
<class 'datetime.datetime'>

aaron-kaplan: So `can't compare int with datetime.datetime` is a very unhelpful error message...

Two alternative suggestions:
- Replace `filename2datetime` with `filename2datetime64` which returns a `np.datetime64`.
- Convert the `datetime.datetime` to a `pd.Timestamp`, which is supposedly comparable with `datetime.datetime`

remicousin: the following works

np.datetime64(last_T_nc) < last_T_zarr.values

aaron-kaplan: That's better than my second suggestion. My first suggestion is arguably simpler.

remicousin: pushed...

sorry... I didn't figured that one out a dew days ago...

aaron-kaplan (approved):
Will you be running this in the container

aaron-kaplan:
Thinking about how to build a workflow around this... when I tried the scipt it was extremely slow. Is it still slow when there's nothing to update?

remicousin:
no. It didn't feel like it. I've been updating either nothing or one month at a time (or creating up to 5 months now)

aaron-kaplan:
I'm asking because I'm wondering whether we should add the ENACTS zarr update step to `update_datalib`. Is it fast enough for that not to be annoying?

remicousin:
It doesn't take time, but we already ran in a case where the filename pattern expected by enactstozarr was not respected (Guyana). So at least for that reason, we might want to wait until we parameterize that aspect or something?

aaron-kaplan:
It will be optional, because not everyone has both ENACTS and python maprooms. So we can put it in the update script with a conditional such that it only runs on servers where it's been enabeld. Then we can enable it at one partner at a time, and if/when we need to enable it on one that requires that the script be generalized, we can make those changes then.

aaron-kaplan:
Which country was the impetus for doing this now?

PR #406, opened 2024-02-27
Djibouti OND WRSI

aaron-kaplan:
This is only for OND because FEWS in the DL doesn't have a rangeland WRSI product for MAM, and the agricultural WRSI products rarely have values over Djibouti. See discussion on https://trello.com/c/Y9Y12Kaa/38-djibouti-wrsi


aaron-kaplan:
Closing because this is now part of https://github.com/iridl/python-maprooms/pull/407

PR #414, opened 2024-03-26
Xchourio/oromia bad years

xchourio:
Task: Update southern Oromia OND bad years 
Project:  AA(FbF)

To achieve this, we need to produce zarr files from a shared CSV file. I've used the script from Madagascar (bad-years-v3.py) as a basis, which uses a file with similar data values (ranks). I've generalized it in such a way that we can use it for any region/country that provides us with a CSV file. The CSV (see bad_years.csv example) file should have a descriptive first line, followed by a column with the year and a column with the values (ranks) for each station (the column name should be the station). I've added validations considering that in the future, this will be handed over to the local parthners  who will be executing it.

Does it make sense to ask the FbF group to take care of updating the bad_year file following this format, or is it better to do it ourselves?

zarr file is available in:  /data/aaron/fbf-candidate/oromia/bad-years-v3-ond.zarr


fbfmaproom/data-conversion-scripts/oromia/bad-years-v3-3season.py
```diff
@@ -0,0 +1,43 @@
...
+}
+
+seasons=["DJF","JFM","FMA","MAM","AMJ","MJJ","JJA","JAS","ASO","SON","OND","NDJ"]
+
+if  len(sys.argv) < 3:
+    print("\033[91m"+"Two arguments is expected. (season, region)"+"\033[0m")
+    print("\033[1m"+"Example: python bad-years-v3-3season.py ond oromia"+"\033[0m")
+    sys.exit()
+elif not sys.argv[1].upper() in seasons: 
+    print("\033[91m"+"The station is not allowed."+"\033[0m")
```
aaron-kaplan, line 20: station --> season

fbfmaproom/data-conversion-scripts/oromia/bad-years-v3-3season.py
```diff
@@ -0,0 +1,43 @@
...
+    print("\033[1m"+"Example: "+", ".join(list(map(str.lower,seasons))) +"\033[0m")
+    sys.exit()
+elif not sys.argv[2].lower() in regionsdic: 
+    print("\033[91m"+"The region is not defined."+"\033[0m")
+    print("\033[1m"+"regions avalibale: "+ (", ".join(regionsdic.keys())) +"\033[0m")
```
aaron-kaplan, line 25: avalibale --> available

fbfmaproom/data-conversion-scripts/oromia/bad-years-v3-3season.py
```diff
@@ -0,0 +1,43 @@
...
+import xarray as xr
+import sys
+import os
+
+dirout='/data/aaron/fbf-candidate/'
+regionsdic = {
+        "oromia": {"file": "/data/aaron/fbf-candidate/original-data/oromia/bad_years.csv"}
+}
+
+seasons=["DJF","JFM","FMA","MAM","AMJ","MJJ","JJA","JAS","ASO","SON","OND","NDJ"]
```
aaron-kaplan, line 13: We have used one- and four-month seasons occasionally in the AA project. This is fine for now for Oromia, but if we're going to generalize this, I think we need a different representation of the season, e.g. Dec-Feb like in Ingrid.

aaron-kaplan (commented):
Looks good so far. Please add this dataset to fbfmaproom-sample.yaml and test it in the application.

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -1292,6 +1292,17 @@ countries:
                     colormap: precip
                     lower_is_worse: no
                     format: bad
+                bad-years-v3:
```
aaron-kaplan, line 1295: Why is this called v3? I don't see a v2 anywhere.

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -1292,6 +1292,17 @@ countries:
...
+                    label: Drought humanitarian ranking v3
+                    description: Drought ranking of expert committee humanitarian data review (v3)
+                    path: ethiopia/southern-oromia/bad-years-v3-ond.zarr
```
aaron-kaplan, line 1298: The file for the previous version is in the `ethiopia` subdirectory and has `southern-oromia` in its filename. Here you've made a subdirectory called `southern-oromia` instead of having `southern-oromia` in the filename. Please be consistent.

fbfmaproom/data-conversion-scripts/oromia/bad-years-v3-3season.py
```diff
@@ -0,0 +1,57 @@
...
+import pandas as pd
+import xarray as xr
+import sys
+import os
+
+
+dirout='/data/aaron/fbf-candidate/'
+regionsdic = {
+    "ethiopia/southern-oromia": {"file": "/data/aaron/fbf-candidate/original-data/oromia/bad_years.csv"},
+  "djibouti": {"file": "/data/aaron/fbf-candidate/original-data/djibouti/bad_years.csv"}
```
aaron-kaplan, line 12: Why is there only one file for Djibouti? The Trello card has three, one for each season.

fbfmaproom/data-conversion-scripts/oromia/bad_years.csv
```diff
@@ -0,0 +1,9 @@
+one  column for the years (year); a column for each season (in lowercase); the values in the season represent a rank.
+year,ond
+2000,6
+2003,5
+2008,3
+2011,2
+2015,4
+2021,1
+2022,7
```
aaron-kaplan, line 9: You added the oromia csv file but not the djibouti one(s)?

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -1934,6 +1945,17 @@ countries:
                     colormap: precip
                     lower_is_worse: yes
                     format: number0
+                bad-years-v3:
```
aaron-kaplan, line 1948: Same question about the version number here. In Trello, Dante explicitly said it should be called version 2.

fbfmaproom/data-conversion-scripts/oromia/bad-years-v3-3season.py
xchourio: OK, but this would even be in the file name, right? Meaning the final file would be called "southern-oromia-bad-years-v3-Oct-Dec.zarr"

fbfmaproom/data-conversion-scripts/oromia/bad-years-v3-3season.py
xchourio: This is because the script is designed so that a single file contains all  seasons. Therefore, if there is an update or a new season, it only  is added (or updated) in one file. That's why I asked earlier if it makes sense that FbF group to be responsible for maintaining and sharing a single file with updates instead of N files 

Example: 
File:
year, jas, mam, ond
1991 , 6. , 17,  6
1992, 33, 10, 33
1993, 6,  33, 33

then:
python bad-years-v3-3season.py ond djibouti 
python bad-years-v3-3season.py jas djibouti.  
.......... etc

fbfmaproom/data-conversion-scripts/oromia/bad-years-v3-3season.py
aaron-kaplan: Not necessarily, since the filename is arbitrary, but it wouldn't hurt to be consistent. Currently 

(Let's not delay getting the Oromia data into the tool while we discuss standardization for the longer term.)

fbfmaproom/fbfmaproom-sample.yaml
aaron-kaplan: e.g. see line 1288.

fbfmaproom/data-conversion-scripts/oromia/bad-years-v3-3season.py
aaron-kaplan: I see. So you combined the three files into one by hand, right? I prefer to avoid manual steps, because it's hard to keep track of how they were done, and to identify and fix bugs when we make a mistake.

I agree that for the future, it's a good idea to define a standard file format and asking FIST to adhere to it is a good idea, so we don't have to keep writing one-off scripts to parse different formats.

fbfmaproom/data-conversion-scripts/oromia/bad-years-v3-3season.py
```diff
@@ -0,0 +1,154 @@
+import cftime
```
aaron-kaplan, line 1: This file should be moved from the oromia directory to the ethiopia directory.

Since it also handles data for djibout, let's put a git-managed symlink to it from the djibouti directory. And then after we get this merged, we can look into creating a library of reusable data conversion utility functions.

fbfmaproom/data-conversion-scripts/oromia/bad-years-v3-3season.py
```diff
@@ -0,0 +1,154 @@
...
+        except Exception as e:
+            print(f"An error occurred while opening the Zarr file: {e}")
+    else:
+        print("Zarr directory does not exist.")
+        return None
+
+dirout='/data/aaron/fbf-candidate/'
+regionsdic = {
+    #"ethiopia/southern-oromia": {"file": "bad_years.csv","time-span":'2000-2023'}
+    "ethiopia/southern-oromia": {"file": "/data/aaron/fbf-candidate/original-data/oromia/bad_years.csv",
```
aaron-kaplan, line 41: Move the original file to the ethiopia directory as well.

fbfmaproom/data-conversion-scripts/oromia/bad_years.csv
aaron-kaplan: Oh, actually we usually don't keep the data files in git. This is a tiny one, so there's no technical problem keeping it in git, but this isn't the place I would normally look for it.

fbfmaproom/fbfmaproom.py
```diff
@@ -200,6 +200,17 @@ def format_enso(x):
...
 
+def format_nma_bady(x):
+    if np.isnan(x):
+        return ""
+    if np.isclose(x, 3):
+        return "Moderate Impact"
+    if np.isclose(x, 2):
+        return "Normal"
+    if np.isclose(x, 1):
+        return "High Impact"
```
aaron-kaplan, line 211: This order seems wrong. Normal shouldn't be in between moderate and high, it should be lower than moderate. Is this the order Dante said to use?

fbfmaproom/fbfmaproom.py
```diff
@@ -200,6 +200,17 @@ def format_enso(x):
         return "El Niño"
     assert False, f"Unknown enso state {x}"
 
+def format_nma_bady(x):
```
aaron-kaplan, line 203: NMA is the National Meteorology Agency of Ethiopia. We shouldn't have anything country-specific in the code, all differences between countries are supposed to be in the config file. So either we add a mechanism for defining custom categories in the config file, or we call this "format_bad_ternary" (as opposed to format_bad which is for binary classification) and let it be used for other countries too.

aaron-kaplan: BTW yes, I think you found the right mechanism for adding new categories.

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -1934,6 +1945,27 @@ countries:
                     colormap: precip
                     lower_is_worse: yes
                     format: number0
+                bad-years-rank-v2:
+                    label: Reported bad years v2
+                    description: Adjusted rank from NMA’s Bad Year Classification (AA Workshop 2024)
```
aaron-kaplan, line 1950: The apostrophe in this line is a non-ASCII unicode character, which caused the application to fail in the production container environment for some reason. We ought to be able to handle unicode strings, but while I look into that, please replace it with an ASCII apostrophe.

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -2051,6 +2083,27 @@ countries:
...
```
aaron-kaplan, line 2088: another apostrophe

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -2181,6 +2234,27 @@ countries:
...
```
aaron-kaplan, line 2239: another apostrophe

fbfmaproom/fbfmaproom-sample.yaml
xchourio: replaced, from NMA’s to NMA's

fbfmaproom/fbfmaproom-sample.yaml
xchourio: same above

fbfmaproom/fbfmaproom-sample.yaml
xchourio: replaced, from NMA’s to NMA's

fbfmaproom/data-conversion-scripts/oromia/bad-years-v3-3season.py
xchourio: done, I moved the script to ethiopia , then -> git symbolic-ref bad-years-v3-3season.py ../djibouti/bad-years-v3-3season.py

aaron-kaplan:
@xchourio one last change please: rename `bad-years-v3-3season.py` to `southern-oromia-bad-years-v3-3season.py`.

xchourio:
> @xchourio one last change please: rename `bad-years-v3-3season.py` to `southern-oromia-bad-years-v3-3season.py`.

Done!

xchourio:
Is it ok if I merge this branch to master?  (I'm already completed the
rebase step)

aaron-kaplan:
Please do.

PR #415, opened 2024-03-28
PyCPT forecast file structure

aaron-kaplan:
Allows issue month dirs to be at the top level instead of inside a subdir per season, in the case where there's only one season. The structure with the intermediate dir is still accepted too.

### DIFF
--- a/enacts/flex_fcst/cpt.py
+++ b/enacts/flex_fcst/cpt.py
@@ -110,22 +110,25 @@ def read_pycptv2dataset(data_path):
     mu_mslices = []
     var_mslices = []
     obs_slices = []
-    target_count = 0
-    for targets in Path(data_path).iterdir() :
-        target_count = target_count + 1
-    for targets in Path(data_path).iterdir() :
-        new_mu, new_var, new_obs = read_pycptv2dataset_single_target(targets)
-        if target_count > 1 :
-            L = (((new_mu["Ti"].dt.month - new_mu["S"].dt.month).squeeze() + 12) % 12).values
-            new_mu = new_mu.assign_coords({"L": L}).expand_dims(dim="L")
-            new_var = new_var.assign_coords({"L": L}).expand_dims(dim="L")
-        mu_mslices.append(new_mu)
-        var_mslices.append(new_var)
-        obs_slices.append(new_obs)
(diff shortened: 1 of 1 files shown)


enacts/flex_fcst/cpt.py
```diff
@@ -111,21 +111,22 @@ def read_pycptv2dataset(data_path):
...
+    children = list(data_path.iterdir())
+    if 'obs.nc' in map(lambda x: str(x.name), children):
+        fcst_mu, fcst_var, obs = read_pycptv2dataset_single_target(data_path)
+    else:
+        for target in children:
+            new_mu, new_var, new_obs = read_pycptv2dataset_single_target(target)
+            if len(children) > 1:
+            mu_mslices.append(new_mu)
+            var_mslices.append(new_var)
+            obs_slices.append(new_obs)
```
aaron-kaplan, line 125: This must not be the version I tested. The indentation is wrong.

enacts/flex_fcst/cpt.py
```diff
@@ -111,21 +111,22 @@ def read_pycptv2dataset(data_path):
     var_mslices = []
     obs_slices = []
     target_count = 0
```
remicousin, line 113: target_count obsolete and _slices initializations could be put under else:

enacts/flex_fcst/cpt.py
```diff
@@ -111,21 +111,22 @@ def read_pycptv2dataset(data_path):
...
+
+    data_path = Path(data_path)
+    children = list(data_path.iterdir())
+    if 'obs.nc' in map(lambda x: str(x.name), children):
+        fcst_mu, fcst_var, obs = read_pycptv2dataset_single_target(data_path)
+    else:
+        for target in children:
+            new_mu, new_var, new_obs = read_pycptv2dataset_single_target(target)
+            if len(children) > 1:
+            mu_mslices.append(new_mu)
```
remicousin, line 123: no more L? How is the appending going to work?

enacts/flex_fcst/cpt.py
remicousin: funny... my brain just re-indented in its mine as it was expecting things to be.

aaron-kaplan: Fixed. Please review.

remicousin (approved):

PR #421, opened 2024-05-02
Load boundaries only once

aaron-kaplan:
On a slow connection, complex admin boundary layers take a long time to load. This doesn't fix that, but it reduces the impact by loading them only once, on initial page load, rather reloading them than every time any of the controls change.


aaron-kaplan:
Oops, forgot to add a reviewer when I sent this last week.

remicousin:
where is this branched from? there is no config-dev-sample.yaml in this branch. Is that on purpose? Do you want to rebase on master first?

aaron-kaplan:
Rebased.

remicousin (approved):
Tested and behaves on 2 Senegal Maprooms. I thought I had problems in the past trying to have ids like tile_layer, defined in a CallBack and updated in another. But maybe that depends on subtle set up. And that was a while ago so I probably did it wrong, or was trying something else...

aaron-kaplan:
You're right, there's supposed to be a callback validation step that checks that the declared inputs and outputs of all the callbacks exist in the layout. That check can be suppressed, but as far as I can see we haven't suppressed it. So I don't know why this works. Will investigate.

PR #425, opened 2024-05-30
Remic/pepsico

remicousin:
This is what I plan to use to zarrify ISIMIP. It's a generalization of enactstozarr. The main difference is that for ENACTS, data files and data have the same time step, resulting in no time dimension in nc files. Whereas ISIMIP data is daily in chunks of 10-yearly files.

It seemed to have worked for one variable, for one model, for one scenario. Remains to read that zarr back to see if it seems ok; and to add documenation with respect to the new optional arguments that facilitate the generalization.

Generally speaking, I expect pepsico to use a lot of what is in enacts but I can't have it part of it. So for now, copy-and-pasting things until we push one level up the common parts.

Review can start I think, in spite of what I mentiones remaind to be checked/done (Xandre and Jeff mostly as FYI

### DIFF
--- a/pepsico/isimiptozarr.py
+++ b/pepsico/isimiptozarr.py
@@ -1,36 +1,21 @@
-import os
-import sys
-import numpy as np
+#import os
+#import sys
+#import numpy as np
 import xarray as xr
-import datetime as dt
+#import datetime as dt
 from pathlib import Path
-import pingrid
+#import pingrid
 from functools import partial
-import calc
+#import calc
 
 
-CONFIG = pingrid.load_config(os.environ["CONFIG"])
-VARIABLE = sys.argv[1] #e.g. precip, tmax, tmin -- check your config 
-TIME_RES = sys.argv[2] #e.g. daily, or dekadal -- check in your config
-INPUT_PATH = (
-    f'{CONFIG["datasets"][TIME_RES]["nc_path"]}'
-    f'{CONFIG["datasets"][TIME_RES]["vars"][VARIABLE][0]}'
-)
-OUTPUT_PATH = (
-    (
-        f'{CONFIG["datasets"][TIME_RES]["zarr_path"]}'
-        f'{CONFIG["datasets"][TIME_RES]["vars"][VARIABLE][0]}'
-    ) if CONFIG['datasets'][TIME_RES]['vars'][VARIABLE][1] is None
-    else (
-        f'{CONFIG["datasets"][TIME_RES]["zarr_path"]}'
-        f'{CONFIG["datasets"][TIME_RES]["vars"][VARIABLE][1]}'
-    )
-)
-CHUNKS = CONFIG['datasets'][TIME_RES]['chunks']
-ZARR_RESOLUTION = CONFIG['datasets'][TIME_RES]["zarr_resolution"]
+TOP_PATH = Path("/Data/data24")
+COMMON_PATH = f'ISIMIP3b/InputData/climate/atmosphere/bias-adjusted/global/daily'
+INPUT_PATH = TOP_PATH / COMMON_PATH
+CHUNKS = {"X": 72, "Y": 36, "T": 365*4+1}
 
 
-def set_up_dims(xda, time_res="daily"):
+def set_up_dims(xda, time_res="daily", time_dim=None, lon_dim="Lon", lat_dim="Lat"):
     """Sets up spatial and temporal dimensions from a set of time-dependent netcdf
     ENACTS files.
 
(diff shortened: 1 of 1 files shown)


pepsico/isimiptozarr.py
```diff
@@ -0,0 +1,284 @@
...
+    See Also
+    --------
+    calc.read_zarr_data, filename2datetime64, nc2xr, xarray.Dataset.to_zarr
+    """
+    print(f"converting files for: {time_res} {var_name}")
+    netcdf = list(sorted(Path(input_path).glob(file_var_pattern)))
+    if Path(output_path).is_dir() :
+        current_zarr = calc.read_zarr_data(output_path)
+        last_T_zarr = current_zarr["T"][-1]
+        last_T_nc = filename2datetime64(netcdf[-1], time_res=time_res)
```
aaron-kaplan, line 226: This assumes that the file that comes last in alphabetical order is the latest one chronologically. That might not always be a valid assumption, e.g. I've seen filenames that used month names instead of numbers (Aug comes before Jan alphabetically, but not chronologically) and ones that put the year last (01-2024 comes before 06-2023 alphabetically but not chronologically).

It's ok for a dataset-specific script where we know that the filename pattern supports it, but we need to be careful of this when factoring out a reusable library.

remicousin:
> If/when you're ready to make a shared zarrification library, you can put it in the top level of this repo alongside `pingrid` and `controls.py`, and symlink it into the subdirectories that use it. E.g. `enacts/pingrid` is a symlink to `../pingrid`.
> 
> By the way, let's try to make sure all new zarrification scripts preserve or add appropriate CF `standard_name` metadata from now on.

Will make it a common library outside of this PR, just for the sake of time and to have a zarr version asap for Sheen to work on it. But will try to do it sooner than later so that we don't linker with two versions meant to be the same.

Can you give an example of the CF conventions preservation/additions? or how/what to check exactly? or a workflow to run that check?

pepsico/isimiptozarr.py
remicousin: Right. They definitely are filenames that of which alphanumerical order is not chronological order. Shall I file this in an issue? Neither ENACTS nor ISIMIP are that case.

aaron-kaplan:
OK for extracting a library in another PR.

There might be multiple CF conventions to preserve, but the one I have in mind at the moment is a single attribute called `standard_name`. The full list of possible values is [here](https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html). Both data variables and coordinate variables should have it in most cases. E.g.

https://iridl.ldeo.columbia.edu/SOURCES/.Models/.SubX/.EMC/.GEFSv12_CPC/.hindcast/weekly/.pr/
standard_name: precipitation_flux

https://iridl.ldeo.columbia.edu/SOURCES/.Models/.SubC/.EMC/.GEFSv12/.hindcast/.weekly/.pr/.X/.standard_name/#expert
standard_name: longitude

Set them in xarray like this:
```
ds['pr'].attrs['standard_name'] = 'precipitation_flux'
ds['pr']['X'].attrs['standard_name'] = 'longitude'
```

remicousin:
The original nc files have standard_names and from a quick check, they are ported for all type of vars to the Xarray dataset that is then fed to to_zarr. They are correct for the coordinates (time, lon, lat). I haven't checked if they are coorect for the actual variables.

I am not sure how to automatize this whole process and it might not be possible but maybe a first step I can do now is stop the zarrification process if at least one of the vars involved doesn't have a standard_name? Then for now, the data curator will have to adapt their code to add it, and maybe later we can figure out some more involved automation?

aaron-kaplan:
That sounds fine.

pepsico/isimiptozarr.py
```diff
@@ -29,6 +29,14 @@ def set_up_dims(xda, time_res="daily", time_dim=None, lon_dim="Lon", lat_dim="La
...
         indicates the time resolution of the set of time-dependent files.
         Default is "daily" and other option is "dekadal"
+    time_dim : str, optional
+        indicates name of time dimension in xda, to be renamed "T"
+    lon_dim : str, optional
+        indicates name of longitude dimension in xda, to be renamed "X"
+        Default is "Lon"
+    lat_dim : str, optional
+        indicates name of latitude dimension in xda, to be renamed "Y"
+        Default is "Lat"
```
aaron-kaplan, line 39: If the input data has CF standard_names, we should probably use those to identify the lat, lon, and time dimensions. If it doesn't, then these arguments should become mandatory, and we should add appropriate standard_names to the output.

Just thinking out loud, doesn't have to be done now if you're in a hurry. In any case, there's no point doing this kind of generalization under pepsico, only when it moves to the top level.

aaron-kaplan (approved):

PR #429, opened 2024-06-10
Tried testing opening file, but still says file not found.

sheenzutshi:


remicousin:
Looks like you already committed your comparison.py into master while in this PR comparison.py is empty. So I am going to decline this PR and we can talk about PR later on. I am going to have a look at what's on master now:

https://github.com/iridl/python-maprooms/blob/master/pepsico/comparison.py

remicousin:
Actually you could keep that branch/PR around if you want to exercise on git manipulations. Just make sure not to merge it back into master.

PR #430, opened 2024-06-10
Comparison ready to be tested

sheenzutshi:
Seeing if the comparison of zarr and netcdf files worked.


sheenzutshi:
I finished checking the zarr conversion for the hurs variable. It said that the datasets are identical for all 7 NetCDF hurs files compared with the zarr file.

pepsico/comparison.py
```diff
@@ -1,14 +1,66 @@
...
+    #open datasets
+    zarr_dSet = xr.open_zarr(ds_zarr)
+    nc_dSet = xr.open_dataset(nc_file_path)[['hurs']]
+
+    #mapping from zar to nc dimensions (X,Y,T) vs (lon,lat,time)
+    coord_map = {'X': 'lon', 'Y': 'lat', 'T': 'time'}
+
+    #xararay rename function to reset zarr dimensions to nc labels 
+    zarr_dSet = zarr_dSet.rename({orig_dim: new_dim for orig_dim, new_dim in coord_map.items() if orig_dim in zarr_dSet.dims})
+    zarr_dSet = zarr_dSet.rename({orig_coord: new_coord for orig_coord, new_coord in coord_map.items() if orig_coord in zarr_dSet.coords})
```
remicousin, line 28: it doesn't matter here, but in the future, let's have our libraries use our conventions, that is to say X, Y, T and thus change 3rd party conventions to ours

pepsico/comparison.py
```diff
@@ -1,14 +1,66 @@
...
+
+    #align the zarr chunk with overlapping nc data
+    time_start = max(nc_dSet['time'].min().values, zarr_dSet['time'].min().values) #start time = compare earliest time of both files and choose latest between the 2
+    time_end = min(nc_dSet['time'].max().values, zarr_dSet['time'].max().values) #minimimum of the 2 files max time
```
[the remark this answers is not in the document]
remicousin: my zarrification code is supposed to have done the whole nc dataset, which is 10 years per file. The start and end year of the nc files are in their filename, so indeed Sheen could use that to retrieve the expected start and end of zarr to test for each file and so maybe that's what she should be doing.
Another option would be to test that neither nc_dSet nor zarr_dSet are empty before comparing them.

pepsico/comparison.py
```diff
@@ -1,14 +1,66 @@
 import xarray as xr
-    
-# defining the path to file  
-ds_zarr = '../data/remic/mydatafiles/zarr/test/'
+import os
+#from pathlib import Path
```
remicousin, line 3: you can remove if not using. We use comments only that help code reading. Actual unused code can be deleted rather than commented: that's what git version-control is for (to retrieve potentially past code).

remicousin (changes requested):
Let's add the extra security mentioned in Aaron's comment and we can move ahead

pepsico/comparison.py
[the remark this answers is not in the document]
sheenzutshi: Ok, I fixed this

pepsico/comparison.py
sheenzutshi: Got it, I changed it for future reference.

pepsico/comparison.py
sheenzutshi: Ok makes sense

pepsico/comparison.py
```diff
@@ -1,14 +1,82 @@
...
+
     
-# defining the path to file  
-ds_zarr = '../data/remic/mydatafiles/zarr/test/'
+    #mapping from zar to nc dimensions (X,Y,T) vs (lon,lat,time)
+    coord_map = {'lon': 'X', 'lat': 'Y', 'time': 'T'}
+
+    #xararay rename function to reset nc labels to zarr
+    nc_dSet = nc_dSet.rename({orig_dim: new_dim for orig_dim, new_dim in coord_map.items() if orig_dim in nc_dSet.dims})
+    nc_dSet = nc_dSet.rename({orig_coord: new_coord for orig_coord, new_coord in coord_map.items() if orig_coord in nc_dSet.coords})
```
remicousin, line 50: Would not either one suffice? and then would not the coord case be more general than the dim one?

remicousin: I confirm that either one is enough. In this case all coords are dims and all dims have a coord so the distinction, ast least to rename is meaningless. I would keep coord our of the two.

sheenzutshi: Ok sounds good, I will remove line 49.

pepsico/comparison.py
```diff
@@ -1,14 +1,147 @@
...
+def extract_years(filename):
+    match = re.search(r'(\d{4})_(\d{4})\.nc$', filename)
+    if match:
+        start_year = int(match.group(1))
+        end_year = int(match.group(2))
+        return start_year, end_year
+    else:
+        raise ValueError("Unable to find years")
+
+def find_nearest_indices(dataset, X, Y):
```
remicousin, line 23: sel has a nearest method option

https://docs.xarray.dev/en/stable/generated/xarray.DataArray.sel.html

that I think you could use instead

remicousin (changes requested):

pepsico/comparison.py
sheenzutshi: Thank you for the great suggestion, I have pushed the modified code.

remicousin (approved):
I approve and we can look together at the rebasing/merging process detailed here:
https://bitbucket.org/iridl/setupandtest/wiki/Home
But for anything Pythonic, let's make sure we have @aaron-kaplan approval as well.

pepsico/comparison.py
```diff
@@ -1,14 +1,136 @@
...
+    nc_dSet = xr.open_dataset(nc_file_path)[['hurs']]
+
+
+    #extracting start and end years from nc file
+    start_year, end_year = extract_years(os.path.basename(nc_file_path))
+    time_start = f'{start_year}-01-01'
+    time_end = f'{end_year}-12-31'
+
+    #check if datasets are empty
+    if len(nc_dSet) == 0 or len(zarr_dSet) == 0:
```
aaron-kaplan, line 39: What are you trying to check for here? In practice I think the len of a Dataset is the number of variables it contains, but I can't find that documented anywhere, so you're relying on undocumented behavior here. And I can't imagine what kind of error would result in the Dataset not having any variables.

pepsico/comparison.py
```diff
@@ -1,14 +1,136 @@
...
     
-# defining the path to file  
-ds_zarr = '../data/remic/mydatafiles/zarr/test/'
+    #mapping from zar to nc dimensions (X,Y,T) vs (lon,lat,time)
+    coord_map = {'lon': 'X', 'lat': 'Y', 'time': 'T'}
+
+    #xararay rename function to reset nc labels to zarr
+    nc_dSet = nc_dSet.rename({orig_coord: new_coord for orig_coord, new_coord in coord_map.items() if orig_coord in nc_dSet.coords})
+    
+    nc_dSet_sliced = nc_dSet.sel(T=slice(time_start, time_end))
```
aaron-kaplan, line 50: Why? We got time_start and time_end from the filename. Do you think there might be time points in the file that are outside the range indicated by the filename? I doubt it, but if by some chance that does happen, it would be better to flag that so we can look into it, rather than silently discarding the extra data.

remicousin: Right, that slicing is unnecessary, we can later test the unlikely case where the nc filename doesn't agree with the nc content.

PR #431, opened 2024-06-18
coord is a string

remicousin:
Thus attrs belong to xda[coord] and not at all to coord


remicousin:
@aaron-kaplan please review/approve. This is independent from the dask/zarr environment business

aaron-kaplan (approved):

PR #432, opened 2024-06-25
Monthly averages and maps

sheenzutshi:
It would be great if my pepsico/dailytomonthly_tmin_GFDL-ESM4.py and pepsico/annual_month_maps.py code could be reviewed for any errors. 
The dailytomonthly code intends to print the annual monthly averages for all 12 months. So, all the January's averaged, Februarys, etc. (1951-2014).
The annual_month_maps prints the maps of these monthly averages.


remicousin:
I see that there are the commits from the comp branch/PR. Did you branch monthly from comp? That's probably ok. We'll just have to be careful when merging things back in.

pepsico/annual_month_maps.py
```diff
@@ -0,0 +1,42 @@
...
+    df = month_data['tasmin'].to_dataframe().reset_index()
+
+    # Sort latitudes in ascending order if necessary - for if map is flipped
+    df = df.sort_values(by='Y')
+
+    # Create the map
+    fig = px.imshow(
+        df.pivot(index='Y', columns='X', values='tasmin'), 
+        labels={'color': 'Kelvins'},
+        title=f'Tasmin for Month {month}',
```
remicousin, line 33: can also find Tasmin variable name and likely long_name in metadata  of the zarr

pepsico/annual_month_maps.py
```diff
@@ -0,0 +1,42 @@
...
+        origin='lower'  # Set the origin to lower to flip the y-axis
+    )
+    
+    
+    # Save the map as an HTML file - access through port 8000
+    fig.write_html(f'/home/sz3116/python-maprooms/pepsico/resources/map_month_{month}.html')
```
remicousin, line 39: better not to write figures in the gitHub repo, even if you're careful not to commit it.

remicousin (changes requested):
I dropped some comments that should improve things.

Suggestions for next steps (in another PR: It's better to make small incremental PRs):
-- Make functions and breakout those that are computation vs file writing vs plotting
-- Have scenario, model and variable as parameters of your functions since you are going to apply the same thing over and over again to all of these
-- consider adding a year range selection parameter(s) as well as right now you are averaging across all years but we know that at some point we will select subset of years.

pepsico/dailytomonthly_tmin_GFDL-ESM4.py
```diff
@@ -0,0 +1,33 @@
...
+annual_monthly_avg = monthly_avg_tasmin.groupby("T.month").mean(dim="T")
+
+# Prepare the data for the CSV
+data = {'month': [], 'tasmin': []}
+for month in range(1, 13):
+    month_data = annual_monthly_avg.sel(month=month)
+    tasmin_value = month_data['tasmin'].values  # Extract the tasmin values
+    avg_tasmin_value = np.mean(tasmin_value)  # Calculate the average to get a single value
+    data['month'].append(month)
+    data['tasmin'].append(avg_tasmin_value)
```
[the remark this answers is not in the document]
sheenzutshi: You are right, I fixed this

pepsico/dailytomonthly_tmin_GFDL-ESM4.py
```diff
@@ -0,0 +1,33 @@
...
+
+# Convert to DataFrame
+df = pd.DataFrame(data)
+
+# Save to a single CSV file
+df.to_csv('/home/sz3116/python-maprooms/pepsico/resources/annual_monthly_GFDL-ESM4_tasmin.csv', index=False)
```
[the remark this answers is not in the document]
sheenzutshi: I ddin't change this yet in the current pushed code, but I will. I deleted all the files that were in resources also.

pepsico/annual_month_maps.py
```diff
@@ -0,0 +1,42 @@
...
+# Compute the average tasmin for each month across all years
+annual_monthly_avg = monthly_ds.groupby("T.month").mean(dim="T")
+
+# Prepare the data for the CSV (too big to print in terminal)
+data = {'month': [], 'tasmin': []}
+for month in range(1, 13):
+    month_data = annual_monthly_avg.sel(month=month)
+    tasmin_value = month_data['tasmin'].mean(dim=["X", "Y"]).values  # Get the mean tasmin value
+    data['month'].append(month)
+    data['tasmin'].append(tasmin_value)
```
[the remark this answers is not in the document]
sheenzutshi: Deleted it

pepsico/monthly_seasonal_conv_mapping.py
```diff
@@ -0,0 +1,111 @@
...
+
+def compute_annual_monthly_avg(ds, variable, start_year=None, end_year=None):
+    #compute the avg (variable) for each month across the specified years
+    
+    #if precipitation variable, change from kg to mm per day
+    if variable == 'pr':
+        ds[variable] *= 86400
+    
+    if start_year and end_year:
+        ds = ds.sel(T=slice(f"{start_year}-01-01", f"{end_year}-12-31"))
```
remicousin, line 17: sel's slice accepts None (does nothing) and will select the full year if you give the year only as label, that is to say if input is 2010, will select all of 2010. In other words, you need not the "-01-01" and the "-12-31" and because None is an acceptable input, you need not the if statement and can just apply

ds = ds.sel(T=slice(f"{start_year}", f"{end_year}"))

pepsico/monthly_seasonal_conv_mapping.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+    return fig
+
+    
+
+def main(scenario, model, variable, start_year=None, end_year=None, output_dir='/home/sz3116/python-maprooms/pepsico/resources'):
+    #main to run functions
+
+    # Open zarr file 
+    ds = xr.open_zarr(f'//Data/data24/ISIMIP3b/InputData/climate/atmosphere/bias-adjusted/global/daily/historical/{model}/zarr/{variable}')
```
remicousin, line 72: historical should be {scenario}

pepsico/monthly_seasonal_conv_mapping.py
```diff
@@ -0,0 +1,111 @@
...
+    # Write to CSV file
+    file_path = f'{output_dir}/{scenario}_{model}_{variable}_annual_monthly_avg.csv'
+    df.to_csv(file_path, index=False)
+
+
+def map_averages(df, model, variable):
+
+    df = df.sort_values(by='month')
+    map_fig = px.imshow(
+        df.pivot(index='Y', columns='X', values=variable),
```
remicousin, line 59: both compute_annual_ functions are averaging over X and Y so they don't exist anymore. Are you trying to make maps? or time series? or both?

pepsico/monthly_seasonal_conv_mapping.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # Compute the average for each month across all years
+    monthly_avg = monthly_ds.mean(dim=["X", "Y"])
+    annual_monthly_avg = monthly_avg.groupby("T.month").mean(dim="T")
+    df = annual_monthly_avg.to_dataframe().reset_index()
+    return df
+
+def compute_annual_seasonal_avg(monthly_ds, variable, start_year=None, end_year=None):
+    #compute monthly avgs across all years
+    monthly_avg = monthly_ds.mean(dim=["X", "Y"])
```
remicousin, line 30: this comment is misleading: this is doing a spatial average.

pepsico/monthly_seasonal_conv_mapping.py
```diff
@@ -0,0 +1,111 @@
+import numpy as np
+import pandas as pd
+import xarray as xr
+import plotly.express as px
+from dash import Dash, dcc, html, Input, Output
+from urllib.request import urlopen
+import json
```
remicousin, line 7: I don't think any of those are used? are they? if not, remove them. I am not sure numpy is used either.

pepsico/monthly_seasonal_conv_mapping.py
```diff
@@ -0,0 +1,111 @@
...
+    print("\nVariable attributes:")
+    for var in ds.data_vars:
+        print(f"\nAttributes for variable '{var}':")
+        print(ds[var].attrs)
+    
+    # Compute the average variable for each month across all years
+    annual_monthly_avg = compute_annual_monthly_avg(ds, variable, start_year, end_year)
+    
+    #call seasonal function
+    rolling_seasonal_avg = compute_annual_seasonal_avg(ds, variable, start_year, end_year)
```
remicousin, line 83: compute_annual_seasonal_avg takes a monthly ds. At this point, ds is still the daily data. Maybe what main should do is:
1/ read data according to scenario, model, variable
2/ select data in years
3/ compute monthly averages (have the output be xarray and not df -- the df conversion is really only for writing and plotting)
4/ compute seasonal ones from monthly ones (have the output be xarray and not df -- the df conversion is really only for writing and plotting)
5/ average in space, convert to df and write to csv both outputs of 3 and 4 on one hand
6/ select a month or season from 3 or 4, convert to df and map

I am not entirely sure what end result(s) you are aiming at so the above might not be quite it, but closer.

remicousin (changes requested):
There are still a few problems to fix. Please have a look.

Also, what's up with the previous py files? If you're not going to keep them, you can git rm them.

pepsico/monthly_seasonal_conv_mapping.py
sheenzutshi: ok, I removed them

pepsico/monthly_seasonal_conv_mapping.py
sheenzutshi: right, I will change that

pepsico/monthly_seasonal_conv_mapping.py
sheenzutshi: fixed

remicousin:
@sheenzutshi If you applied fixes, commit and push them so I can review them

pepsico/annual_month_maps.py
sheenzutshi: Added this

pepsico/annual_month_maps.py
sheenzutshi: Ok, I changed the file path

pepsico/monthly_seasonal_conv_mapping.py
sheenzutshi: I have pushed the adjusted code based on these comments. Not sure if it is exactly what you were asking. I broke the code down into more functions so it was more xarray outputs instead of df. Also, I have monthly annual and seasonal annual maps based on a specific month a user inputs in the main method.

pepsico/monthly_seasonal_conv_mapping.py
```diff
@@ -0,0 +1,199 @@
...
+        'JJA': [5, 6, 7],   # Jun, Jul, Aug
+        'JAS': [6, 7, 8],   # Jul, Aug, Sep
+        'ASO': [7, 8, 9],   # Aug, Sep, Oct
+        'SON': [8, 9, 10],  # Sep, Oct, Nov
+        'OND': [9, 10, 11], # Oct, Nov, Dec
+        'NDJ': [10, 11, 0]  # Nov, Dec, Jan
+    }
+    # Select data for the specific season
+    seasonal_data = ds.sel(T=ds['T.month'].isin(season_indices[season]))
+    seasonal_avg = seasonal_data.mean(dim='T')
```
remicousin, line 129: nice, but you had done all the work with your rolling function already. Only you are not using it because you stuck the spatial average in it. If you were taking the spatial average out of that, then you could use the same function to compute seasonal stuff.

pepsico/monthly_seasonal_conv_mapping.py
```diff
@@ -0,0 +1,199 @@
...
+    return rolling_seasonal_avg
+
+
+def create_seasonal_dataframe(rolling_seasonal_avg):
+    # Create a DataFrame for seasonal averages
+    seasonal_avgs_df = rolling_seasonal_avg.to_dataframe().reset_index()
+
+    # Assign the season column with correct length
+    seasonal_avgs_df['season'] = ['DJF', 'JFM', 'FMA', 'MAM', 'AMJ', 'MJJ',
+                                   'JJA', 'JAS', 'ASO', 'SON', 'OND', 'NDJ']
```
remicousin, line 54: I think DJF is likely last, and not first. Since you do a rolling of 3 that is centered, and since data always starts in January (whether you select years or not), the first center will be Feb, and so the first season would be JFM and so on and so forth, the last one would be DJF. Unless groupby reorders them? Please double check.

remicousin (changes requested):
Nice! Don't hate me: I dropped a few more comments :)

Some are petty.

The 2 I care more about are:
1/ please double check that the correspondence made between numbered months/seasons and their "English" representation are correct
2/ I would factor out the spatial average from your monthly average function. This way you would just have:
-- 1 function to monthly average
-- 1 function to seasonally average

Then either you spatially average and make a csv; either you pick a month/season and make a map.

It's probably an overkill to have all those functions that end up just being one line and they probably could be straight into main, but that's ok, often we don't see how things take shape until we actually write them. What was really important was to have the compute/write/plot functions factored out. Because the app is going to use the compute ones, but not to write to csv, nor to map with px.

remicousin (approved):
I will do the merge into comp

PR #436, opened 2024-07-19
Appstruct

remicousin:
Setting up what's needed to make app(s) in pepsico directory. First commit is a pure copy from enacts then 2nd commit adapts to the pepsico situation. Not sure if anything else is needed. Once that and the data are ready, we'll start working on a maproom.py and a layout.py for pepsico/projections.

(Eventually we'd want the common parts between enacts and pepsico to move one level up and be referenced by each, but I don't think we have time for that now -- or rather no time to test that it doesn't break anything in enacts).


remicousin:
I believe we need the docker piece only once we want that be built somewhere in Maproom, correct?

pepsico/config-dev-sample.yaml
```diff
@@ -0,0 +1,10 @@
+# Configuration overrides for development. Copy this file to config-dev.yaml and edit as needed.
+
+mode: debug  # debug, devel or prod
+dev_server_interface: 127.0.0.1
+dev_server_port: 8050
+dev_processes: 4
+loglevel: info
```
aaron-kaplan, line 7: This doesn't appear to be used anywhere. I was tipped off by the fact that there's no default for it in config-defaults.yaml. Please remove from here and from enacts/config-dev-sample.yaml.

pepsico/globals_.py
```diff
@@ -0,0 +1,22 @@
...
+    if isinstance(v, list):
+        GLOBAL_CONFIG['maprooms'][k] = [pingrid.deep_merge(defaultconfig['maprooms'][k], v[i]) for i in range(len(v))]
+    elif v is not None:
+        GLOBAL_CONFIG['maprooms'][k] = pingrid.deep_merge(defaultconfig['maprooms'][k], v)
+
+
+FLASK = flask.Flask(
+    "pepsicomaprooms",
+    static_url_path=f'{GLOBAL_CONFIG["url_path_prefix"]}/static',
+)
```
aaron-kaplan, line 22: Are there going to be multiple maprooms eventually? If not, we can simplify the setup.

aaron-kaplan (approved):
Looks ok so far, but there's no way to test it until the maproom exists. I would lean towards waiting until you've tested it to merge, but there's no risk, so you can merge if you want.

A Dockerfile is needed in order to deploy the application to our servers.

pepsico/globals_.py
remicousin: For the sake of time, we are making a Global one now, but they might want one per "Region" of interest and I am not sure if they'd want that as a control on the Maproom page, or as as many separate applications (which on another hand could also help with performance...). Then if the collaboration continues and they ask for new Maprooms that don't yet exist in classic style, then more Python ones should come up too.

remicousin:
> Looks ok so far, but there's no way to test it until the maproom exists. I would lean towards waiting until you've tested it to merge, but there's no risk, so you can merge if you want.


Let me add maproom/layout.py without data involved. That we save Sheen some time as the Tutorial is a bit outdated with respect to how things are set up.

remicousin:
Note for later: 
localhost:xxxx/python_maproom/
is ok but it links to
localhost:xxxx/python_maproom/projections
that is 404.

I also had to put the following in my local config file. Not sure why it seems to ignore it when in config-defaults. Same thing for maproom: config entries.

But need to double check on all that later.

datasets:
    shapes_adm:
        - name: National
          color: black
          sql: select adm0_code as key, adm0_name as label,
              ST_AsBinary(ST_Simplify(ST_Union(the_geom), .01))
              from g2015_2012_0
          is_checked: True
        - name: Adm2
          color: black
          sql: select (adm0_code, adm1_code) as key, adm1_name as label,
              ST_AsBinary(ST_Simplify(ST_Union(the_geom), .01)) as the_geom
              from g2015_2014_1
          is_checked: False
        - name: Adm2
          color: grey
          sql: select (adm0_code, adm1_code, adm2_code) as key, adm2_name as label,
              ST_AsBinary(ST_Simplify(ST_Union(the_geom), .01)) as the_geom
              from g2015_2014_2
          is_checked: False

pepsico/projections/maproom.py
```diff
@@ -0,0 +1,255 @@
...
+            Output("lng_input_tooltip", "children"),
+            Output("map", "center"),
+            Input("location", "pathname"),
+    )
+    def initialize(path):
+        scenario = "historical"
+        model = "GFDL-ESM4"
+        variable = "pr"
+        data = xr.open_zarr(
+            f'/Data/data24/ISIMIP3b/InputData/climate/atmosphere/bias-adjusted'
```
aaron-kaplan, line 103: Extract this path to the config file. Also occurs in other places below.

remicousin: I meant to leave that out to Sheen.

remicousin:
I am still poking a bit in the dark to figure out why the URL of the app points to 404. If it's evident to you, please unblock me otherwise I'll keep poking. It doesn't seem right that I have to define maprooms: in both config files for it to work as it is so maybe that's where some confusion lies. Maybe I don't have the right syntax here as it changed to allow multiple maprooms from same py files. I need to search what that syntax is.

Otherwise with enactsmaproom conda env:
CONFIG=config-dev-sample.yaml:myconfig.yaml python app.py

where myconfig just has your regular build env stuff and the db pswd

remicousin:
Looks like I made it through. I was the - thingy for multiple maprooms from one file

aaron-kaplan (approved):

remicousin:
For the moment I saw the countries drawn but that's it, no data yet. It's very slow and it looks like storm took a toll with running it. Not sure yet if it's the shapes, or the data, or both.

aaron-kaplan:
Storm doesn't have a huge amount of memory, and @sheenzutshi has a couple of very memory-intensive processes running on it. So (a) give it a try on shortfin01, and (b) if someone shows me Sheen's script maybe I can make a suggestion about how to reduce its memory usage.

remicousin:
Alright, this just plot the map of one variable/model/scenario at it's last time point. Initial load takes a little bit. Most likely due to the shapes as the data comes right after the map + shapes are eventually plotted. Since that parts is a one-time load open landing, that sounds fine. Interestingly enough... this reveals that our drawing of borders is not taken into consideration with respect to Earth periodicity along longitude, as dash-leaflest let's you zoom out or scroll West/East through "multiple Earths." Anyways... not very important at the moment.

This also reads the one variable I rechunked in time. Now I need to go to the Nigeria office to retrieve my passport. After that I can wrap up the rechunking of all the other data, and wrap up the seasonal calculation. And I can talk with Sheen about she is interested in trying to contribute to in the few days left, if she can at all. Independently, it seems that having a first version of this app before she leaves is attainable.

(And so I'll merge this when I come back since was already approve and unless there is new comments in the meantime)

PR #437, opened 2024-07-22
this is what is currently running to create monthly zarr

remicousin:
I changed the 1M resample input into MS as 1M return a warning about imminent deprecation.

I'll amend the PR on app calculations to expect monthly data as input.


aaron-kaplan (commented):
Do we need to continue storing the daily zarr? If we only intend to use monthly averages, couldn't we calculate those directly from the original files?

I think it would be good to have directory path names indicate clearly which files are downloaded from the provider and which are derived via a  script. `/Data/data24/ISIMIP3b/InputData/climate/atmosphere/bias-adjusted/global` doesn't do that for me.

(I actually think maybe we should have completely separate storage areas for original files and derived files, but that's something I'll have to work out with @jturmelle, so I can't give specific advice here.)

aaron-kaplan:
BTW how long did the original zarrification take? We can parallelize these computations with dask.

remicousin:
The daily to monthly is still going and is still in the historical loop.

I don't quite recall but I think the order for the original zarrification from daily nc to daily zarr was 2-3 days.

The PepsiCo colleagues want to be shared the daily data. This will be a one time transfer (by mail apparently). I am not too sure what format they want exactly except that they want it back to nc, but I don't know if they have requests in terms of time and space breakdown file-wise...
Also Sheen's analytical work has been on the daily data directly so we should keep the daily until she's done.
Once both those items are behind us, we can probably delete the daily zarr data (and we can regenerate if need be in the future).

All the data paths of what we generated have 'zarr' in it. Would that be an indication good enough that they are our derived data? Collocating the zarr with the initial data also gives a clue that one comes from the other.

aaron-kaplan:
Sending this much data by email is out of the question!

remicousin:
by *mail*

it's running on dlupdates as datag since it's the only way (I think) I can
write there.

Rémi Cousin (he/him/...)
***@***.***
Phone: +1 8456804422
Fax: +1 8456804864
http://iri.columbia.edu
61 route 9W, Monell Building
Palisades, NY 10964

aaron-kaplan:
Oh, I forgot "mail' had another meaning.

I see the process on dlupdates. It's already using dask, but as dlupdates
is a small machine (4 cores), it's only using five processes,. Garuda can
write to /Data and has 32 cores.

remicousin:
I'll try to remember that for the next job...

aaron-kaplan:
You didn't specify the chunk size, and it seems to have chosen (T, Y, X) = (1, 36, 72).  That results in over 78,000 rather small files, which makes reading slow.

I think I wrote down some rules of thumb for chunking once, but I can't find it. Do you know where that is?

remicousin:
hmmm... I don't know where that it. I recall the rule of thumb being the tiles being sized around 50x50, and to avoid having tiles of different sizes (due to the edges).

The chunks come from what I set up for the initial [zarrification](https://github.com/iridl/python-maprooms/blob/isimipapp/pepsico/isimiptozarr.py) of the daily data

aaron-kaplan:
> The chunks come from what I set up for the initial [zarrification](https://github.com/iridl/python-maprooms/blob/isimipapp/pepsico/isimiptozarr.py) of the daily data

That says

CHUNKS = {"X": 72, "Y": 36, "T": 365*4+1}

which I think means four years per chunk, but when I open your monthly zarr in xarray it says the chunk size is 1 in the T dimension. I think that means one month?!

My first stab would be to keep the X and Y chunk sizes you have, and use 10 years per chunk for T. Keep in mind that the best chunking for finding the timeseries at a single point or small area is the worst for plotting a global map, and vice versa, so test both, unless one of them isn't relevant for the application.

remicousin:
> > The chunks come from what I set up for the initial [zarrification](https://github.com/iridl/python-maprooms/blob/isimipapp/pepsico/isimiptozarr.py) of the daily data
> 
> That says
> 
> CHUNKS = {"X": 72, "Y": 36, "T": 365*4+1}
> 
> which I think means four years per chunk, but when I open your monthly zarr in xarray it says the chunk size is 1 in the T dimension. I think that means one month?!
> 
> My first stab would be to keep the X and Y chunk sizes you have, and use 10 years per chunk for T. Keep in mind that the best chunking for finding the timeseries at a single point or small area is the worst for plotting a global map, and vice versa, so test both, unless one of them isn't relevant for the application.

I see that the daily data is chunked as I asked:
    pr       (T, Y, X) float32 dask.array<chunksize=(1461, 36, 72), meta=np.ndarray>
it means that the resample operator is rechunking to size 1? It's not mentioned in xarray doc but I could probably discover that looking at their code. Regardless, should I force another chunking before writing? and if so, can we do that in parallel as you suggested because it took like 2 days to write the monthly data. Or maybe I can rechunk from the monthly zarr but that would continue to make objects origins harder to track back...

aaron-kaplan:
Suggestion: start the rechunking (on shortfin01 so it will go faster), writing to a new directory. In the meantime, continue using the existing monthly directory for development purposes.

 Another suggestion: work with a subset of the global dataset during development, expand to global once you've got it working.

I share the feeling that manual data transformation steps are to be avoided when possible. When there's no other way, make sure you keep good notes.

remicousin:
The conversation elsewhere on time cells representation makes me think that here, if we want to use center/width, then we'll also have to work on stuff like resample even though I see there are offsets options.

aaron-kaplan:
I thought we agreed that it would be better (for reproducibility) to compute the rechunked monthly directly from daily, instead of going via the badly-chunked monthly.

remicousin:
oh... I understood that the conclusion was to take good notes. I can put the rechunking bit inside the monthly resampling code. Then, are you suggesting I re run all that from the daily data to regenerate the properly chunked monthly data?

remicousin:
ok. I pushed a commit accordingly. I can delete the 1-size chunk monthly version that is used by none. And take my time to rerun this in targetting that location; then I can delete what's in the _rechunked location (that I currently have the app use).

remicousin:
Is it ok to run this? and secondarily to merge it?

aaron-kaplan (approved):

Issue #405, opened 2024-02-26
Better zarr store update from set of nc files


remicousin:
In https://bitbucket.org/iridl/subx-cloud/src/master/zarrify.py I took the approach of creating an empty store and then writing into regions of it. (The motivation in that case was to parallelize writes.) It might be more complicated that what's needed in this case.

It would be nice to be able to either (a) skip opening any netcdf files whose data are already present in the zarr, for speed, or (b) compare pre-existing zarr data to what's in the netcdf files, for extra safety when you're not in a hurry. But that can wait for a subsequent version.

_Originally posted by @aaron-kaplan in https://github.com/iridl/python-maprooms/pull/404#pullrequestreview-1898957700_

PR #439, opened 2024-08-02
Zarrud

remicousin:
This is a start to address Issue #405 . It's not all fresh so it's probably all what we need. It seems to be working. Note that if you add new nc files to your nc directory , they will be added as along as they are after the last time of the zarr. It will do it even if it creates a gap (e.g. last zarr is in Dec 2023 and new nc in Feb 2024). Then Feb 2024 is the new end so it won't ever check back on January 2024 gap.

The ultiamte goal here is to have something that can be included into update_datalib for automation of zarr updates, and thus of Python Maprooms with new data.

### DIFF
--- a/enacts/enactstozarr.py
+++ b/enacts/enactstozarr.py
@@ -214,11 +214,16 @@ def convert(
     if Path(output_path).is_dir() :
         current_zarr = calc.read_zarr_data(output_path)
         last_T_zarr = current_zarr["T"][-1]
-        T_nc = filename2datetime64(netcdf, time_res=time_res).where(lambda x: x > last_T_zarr.values, drop=True)
-        if size(T_nc) == 0 :
+        T_nc = [
+            filename2datetime64(netcdf[i], time_res=time_res)
+            for i in range(len(netcdf))
+        ]
+        T_nc = np.where(T_nc > last_T_zarr.values, T_nc, np.datetime64('nat'))
+        T_nc = T_nc[~np.isnat(T_nc)]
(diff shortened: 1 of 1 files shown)


aaron-kaplan:
It is technically possible to fill in gaps in the zarr. You call `to_zarr` with `region` instead of `append_dim`. See https://bitbucket.org/iridl/subx-cloud/src/1c3ab3bc4d044b346998266ba27c04b60f8cb6b7/zarrify.py#lines-84 .

It's worth discussing whether we should actually do that or not. Arguably, once we've published the data up to time t, everything up to that time should be immutable from then on. If we eventually get data that fills in a gap, we should publish that as a new version of the dataset, not change the existing dataset, to avoid changing the results of analyses that were made when the gap was there.

We're obviously not immediately going to solve the full problem of keeping track of gaps and the protocol for filling them in. I'm not sure which option is the best in the meantime--leave the gaps there, or backfill when we can. I lean towards filling them in.

Whose ENACTS are you testing this on? How many files does it have? How long does the `glob` take? If it's too slow, we could try a different approach? Look up the final zarr date, calculate the next expected date after that, and look for that one file, which I think will be faster than listing all the files. Do that in a loop until you reach today (or some other limit).

aaron-kaplan:
It would be good to have an option (disabled by default) to verify the existing zarr data, in addition to filling in data that are missing from the zarr.

aaron-kaplan (approved):

xchourio (approved):

PR #446, opened 2024-10-15
Set umask before creating files/directories in /data/aaron/fbf-candidate

aaron-kaplan:
Also moving top-level commands into a __main__ block; that's in a separate commit for your reviewing convenience.


aaron-kaplan:
This fell through the cracks. @jturmelle please review.

jturmelle (approved):
My only comment would be it would helpful to add some help text to the argparse arguments.  For example, you could give a list of available datasets.

PR #467, opened 2025-01-17
himudity range from -/+30

remicousin:


aaron-kaplan:
Are you only plotting anomalies? Since neither precip nor relative humidity can be negative.

remicousin:
> Are you only plotting anomalies? Since neither precip nor relative humidity can be negative.

yes: mapping anomalies but time series of actual fields. At least so far.

PR #468, opened 2025-01-17
Set Ethiopia MAM default forecast version to v3

mm5330:


aaron-kaplan (approved):
Perfect. And I see you made v3 the default, so I'll copy that change to the public server.

PR #469, opened 2025-01-17
Prcp anomaly cs

remicousin:
Gives more appropriate color scales variable-wise (in spite of the name of the PR, I did mean to deal with all variables in it). Something seems off with respect to map_attributes factorization; how min/max are obtained; and how min/max values are passed on. But I can't put my finger on it as of now. Maybe it's ColorScale that needs a bit more functionality. Anyway, it's not important unless someone sees something obviously cleaner to do (that is not a major project).

### DIFF
--- a/pepsico/projections/maproom.py
+++ b/pepsico/projections/maproom.py
@@ -368,6 +368,11 @@ def register(FLASK, config):
             colorscale = CMAPS["prcp_anomaly"]
             map_min = -100
             map_max = 100
+        elif variable in ["prsn"]:
+            colorscale = CMAPS["prcp_anomaly_blue"]
+            map_amp = max(abs(data.min().values), abs(data.min().values))
+            map_min = -1*map_amp
+            map_max = map_amp
         else:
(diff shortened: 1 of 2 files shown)


pepsico/projections/maproom.py
```diff
@@ -361,17 +361,20 @@ def seasonal_change(
...
-            map_min = -100
-            map_max = 100
+            colorscale = CMAPS["prcp_anomaly"].rescaled(-100, 100)
         else:
             assert (data is not None)
-            map_amp = max(abs(data.min().values), abs(data.min().values))
-            map_min = -1*map_amp
-            map_max = map_amp
-        return CMAPS["correlation"].rescaled(map_min, map_max), map_min, map_max
+            map_amp = max(abs(data.min().values), abs(data.max().values))
```
aaron-kaplan, line 369: Simpler: `map_amp = np.max(np.abs(data.values))`?

aaron-kaplan (approved):

PR #474, opened 2025-01-24
validating year entries and disabling button if not

remicousin:
Something should probably be factored out because will likely be reused over and over in all sorts of Maprooms, but I don't quite see what it is as of now.


aaron-kaplan (commented):
This probably wouldn't work well over a slow connection. First of all, with debounce=False it makes a callback to the server on every keystroke. That's a lot of requests in quick succession. Second, if it takes half a second to get a reply to a callback, the user could easily click the submit button before the results of validation come back. This kind of validation should be done client-side.

For more complex validation there are client-side callbacks, but in this case it seems like dbc.Input's min and max attributes should be sufficient. I see that we're already using them, but their values are currently hard-coded. I suggest not setting them in the layout, and instead set them in the `initialize` callback.

remicousin:
> This probably wouldn't work well over a slow connection. First of all, with debounce=False it makes a callback to the server on every keystroke. That's a lot of requests in quick succession. Second, if it takes half a second to get a reply to a callback, the user could easily click the submit button before the results of validation come back. This kind of validation should be done client-side.
> 
> For more complex validation there are client-side callbacks, but in this case it seems like dbc.Input's min and max attributes should be sufficient. I see that we're already using them, but their values are currently hard-coded. I suggest not setting them in the layout, and instead set them in the `initialize` callback.

It's been working smoothly on shortfin but I understand it might not elsewhere.

I don't think we can use dynamically min/max. Let's say the initialization for projection years is 2021-2025, that would imply initializing start's max at 2025 and end's min at 2021. Now let's say next the user wants to do 2030-2035, which is a legit ask, if it starts by switching 2021 to 2030, it will turn red; then the user turns 2025 to 2035, since there is no debounce, it will have not impact on the start control that will remain red. The user might still think it's cool to click Submit button, and that will submit None instead of 2025 because that's how min/max works: it returns None for out of range values.

Unless I am missing something, which is very possible, then I may have to go for the client-side callback.

xchourio (approved):
I’m going to approve it, though I’m concerned about the behavior on 'over a slow connection' mentioned by Aaron. We’ll need to stay attentive to feedback from Pepsico, who I assume will share it with other users from different departments. That should give us a better understanding of how it performs.

aaron-kaplan:
> Let's say the initialization for projection years is 2021-2025, that would imply initializing start's max at 2025 and end's min at 2021. Now let's say next the user wants to do 2030-2035, which is a legit ask, if it starts by switching 2021 to 2030, it will turn red; then the user turns 2025 to 2035, since there is no debounce, it will have not impact on the start control that will remain red.

I see. Use your browser dev tools to simulate a slow connection (e.g. for firefox: https://blog.nightly.mozilla.org/2016/11/07/simulate-slow-connections-with-the-network-throttling-tool/). If the server-side callback seems to work ok under those conditions, go ahead, otherwise we should use a client-side callback, which means learning a little javascript (looks like a pretty simple function, though).

remicousin:
> > Let's say the initialization for projection years is 2021-2025, that would imply initializing start's max at 2025 and end's min at 2021. Now let's say next the user wants to do 2030-2035, which is a legit ask, if it starts by switching 2021 to 2030, it will turn red; then the user turns 2025 to 2035, since there is no debounce, it will have not impact on the start control that will remain red.
> 
> I see. Use your browser dev tools to simulate a slow connection (e.g. for firefox: https://blog.nightly.mozilla.org/2016/11/07/simulate-slow-connections-with-the-network-throttling-tool/). If the server-side callback seems to work ok under those conditions, go ahead, otherwise we should use a client-side callback, which means learning a little javascript (looks like a pretty simple function, though).

ok. So what I experience is that those CallBacks are always fast, only that they are queued against other CallBacks and even with 4G/LTE throttling, even the simple test CallBacks (title and description) can take a little while and so if the validation CallBack waits for them to complete, then there is time to click Submit button before it's disabled (or before red warnings show). And of course if it waits for the ts or map, there is time to go to the bathroom and click the Submit button (depending on the level of throttling). So I am not sure what the conclusion to that is: either client-side callback; or I don't know if there is a way to prioritize CallBack and whether that is desirable anyways.

aaron-kaplan:
I don't think we have control over callback priority. 

Simple example of a client-side callback that validates inputs: https://community.plotly.com/t/validation-for-dcc-input/61634/2

But does it really need to be so complicated? What if we just do the validation as the first step of the Submit callback instead?

aaron-kaplan:
Indeed, the fact that it triggers are multiple callbacks makes it a little more complicated. The approach you outlined sounds reasonable. I find that PreventUpdate/no_update is sometimes confusing because it leaves the previous map/chart/table in place, and a user could mistakenly think that the results being displayed correspond to the current values of the inputs, rather than to some previous set of values that is no longer visible. I would prefer to blank out the results (the error_fig approach).

Client-side validation would also be fine.

remicousin:
So I went for the client side option, mostly to learn something, and because it looks to me that it would be a bit less messy. Now it seems that to do so, I'd need to add
dash_mantine_components
to our conda environment.

So... going ahead with that? Or back to last proposal with error figures?

aaron-kaplan:
You don't need dash-mantine-components. The example happens to use dmc.Number, but you can just as well use dcc.Number.

remicousin:
ah? Nothing happened with dcc and I found dmn in another example so I assumed it was mandatory. I guess I had a syntax error then. ok, will look into it again

pepsico/projections/maproom.py
```diff
@@ -322,6 +322,34 @@ def write_map_title(
         )
 
 
+    APP.clientside_callback(
+        """function(start_year, end_year, start_year_ref, end_year_ref) {
+            if (!(!start_year || !end_year)) {
```
aaron-kaplan, line 327: The double negation is a bit confusing. Can't we rewrite this as follows?
```
if (start_year && end_year)
```
Similarly in line 331.

Also confusing is assigning a new value (especially of a different type) to a parameter. I would prefer to introduce new variables `valid_start` etc. instead of changing the value of `start_year` etc.

remicousin:
Shit I forgot to rebase...

remicousin:
Should I do 

git reset --hard HEAD@{1}

git push origin master

Then return to rebasing

?

remicousin:
hmmm... I can't tell what's wrong in the history... is gitHub rebasing by default? I clicked the squash and merge button I believe. I am going to leave it at that but if need be revisited, need be sooner than later.

aaron-kaplan:
It does look like it rebased. I'm not sure how, because before you merged it said there were conflicts. It's possible that one of the early commits on this branch introduced a conflict and a later commit reverted that change, so squashing them together made the conflict disappear.

Un-pushing a commit you've already pushed to master is frowned upon. Once something is in master, it's generally best to leave it there and just fix whatever's wrong in a subsequent commit.

PR #475, opened 2025-01-24
bringing units to the colorscale

remicousin:
This made me have the colorscale be vertical to accommodate sometimes long units. It would make sense that colorbar could have a label or title in which to put the units ones, but it doesn't have that. The way to tailor further with having maybe the units on only one tick would be to create the entire ticking manually and feed it to colorbar. But that sounds tedious, at least for now.

Also made me revisit a bit the map_attributed factored out function since now it needs data all the time to read units. Or at least that workflow seemed to make more sense to me.

### DIFF
--- a/pepsico/projections/maproom.py
+++ b/pepsico/projections/maproom.py
@@ -17,6 +17,7 @@ from globals_ import FLASK, GLOBAL_CONFIG
 import app_calc as ac
 import numpy as np
 
+xr.set_options(keep_attrs=True)
 
 STD_TIME_FORMAT = "%Y-%m-%d"
 HUMAN_TIME_FORMAT = "%-d %b %Y"
@@ -204,7 +205,7 @@ def register(FLASK, config):
             x=ts["T"].dt.strftime(STD_TIME_FORMAT),
             y=ts.values,
             customdata=ts["seasons_ends"].dt.strftime("%B %Y"),
-            hovertemplate=("%{x|"+start_format+"}%{customdata}: %{y:.2f}" + units),
+            hovertemplate=("%{x|"+start_format+"}%{customdata}: %{y:.2f} " + units),
             name=name,
             line=pgo.scatter.Line(color=color),
             connectgaps=False,
@@ -288,10 +289,6 @@ def register(FLASK, config):
                 start_format = "%b %Y - "
             else:
                 start_format = "%b-"
-            if data_ds["histo"].attrs["units"] == "Celsius" :
-                units = "˚C"
-            else:
-                units = data_ds["histo"].attrs["units"]
             local_graph = pgo.Figure()
             data_color = {
                 "histo": "blue", "picontrol": "green",
@@ -301,11 +298,15 @@ def register(FLASK, config):
             lat_units = "˚N" if (lat >= 0) else "˚S"
             for var in data_ds.data_vars:
                 local_graph.add_trace(plot_ts(
-                    data_ds[var], var, data_color[var], start_format, units
+                    data_ds[var], var, data_color[var], start_format,
+                    data_ds[var].attrs["units"]
                 ))
             local_graph.update_layout(
                 xaxis_title="Time",
-                yaxis_title=f'{data_ds["histo"].attrs["long_name"]} ({units})',
+                yaxis_title=(
+                    f'{data_ds["histo"].attrs["long_name"]} '
+                    f'({data_ds["histo"].attrs["units"]})'
+                ),
                 title={
                     "text": (
                         f'{data_ds["histo"]["T"].dt.strftime("%b")[0].values}-'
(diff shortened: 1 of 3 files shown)


xchourio (approved):

pepsico/projections/maproom.py
```diff
@@ -17,6 +17,7 @@
 import app_calc as ac
 import numpy as np
 
+xr.set_options(keep_attrs=True)
```
aaron-kaplan, line 20: This gives me pause because it's a global setting, and thus will affect not only this maproom, but all maprooms that run in the same process as this one. (For the moment it looks like this is the only one, but the project is set up to maybe have more in the future.) If we're going to change global settings, We should probably do it in app.py rather than in an individual maproom.

BTW the [doc](https://docs.xarray.dev/en/stable/generated/xarray.set_options.html) says the default behavior is "attrs should only be kept in unambiguous circumstances." Do you know what it means by unambiguous, and why it's ambiguous in this case?

pepsico/projections/maproom.py
```diff
@@ -347,19 +348,19 @@ def seasonal_change(
         data = data - ref
         if variable in ["hurs", "huss", "pr"]:
             data = 100. * data / ref
-            data["units"] = "%"
+            data.attrs["units"] = "%"
         return data.rename({"X": "lon", "Y": "lat"})
 
 
-    def map_attributes(variable, data=None):
+    def map_attributes(data, to_dash_leaflet=False):
```
aaron-kaplan, line 355: I don't love the `to_dash_leaflet` option. I think it's better to have this function always return the same representation, and have the client translate it to dash-leaflet representation if that's what's needed.

The general principle is to use a uniform representation throughout our application, and when interfacing with other applications or libraries that use a different representation, do the translation at the interface, rather than letting the complexity of multiple representations propagate through our code.

pepsico/projections/maproom.py
remicousin: So the operators that drop the attributes are the mean("T") in seasonal_change (in maproom.py) but also the subtraction (data - ref) operation. The attributes are long_name, standard_name and units. They are the same for both data and ref.  Curiously the mean() applied to groupby in seasonal_data (in app_calc.py) is not dropping the attributes.

I don't quite see how groupby().mean() would be unambiguous and .mean("T") would be, even though I can see that one might want to reconsider both the names and units of the result of an average. As for the subtraction, it should be unambiguous for units given they are the same; possibly ambiguous for the names... But maybe the code doesn't check if units are the same... and if they aren't the same then it can be ambiguous. Or maybe the behavior applies for any possible attribute and maybe the logic is that sampling functions are unambiguous and all other operators are. Then maybe groupby is considered sampling and its behavior is propagated to its methods...

Unclear... I could look in the code is you like. I started by putting the keep_attrs=True where needed, until I hit the subtraction case. I started to look at a function for subtracting instead of using - sign then figured that at this rate, keep_attrs=True was going to be needed everywhere...

pepsico/projections/maproom.py
remicousin: taken out

pepsico/projections/maproom.py
aaron-kaplan: I have a feeling that using keep_attrs=True everywhere would result lots of invalid and unintended attributes all over the place.

Also, I looked at the implementation of xarray.set_options and it's clearly not thread-safe. The authors didn't consider that people might use the library in a multi-threaded application such as a web server.

There might be a funny way to write subtraction that gives it a keep_attrs option, but if not, you can always just manually copy the desired attributes from one DataArray to the other. Please do that.

remicousin: How is that (last commit)?

aaron-kaplan: good.

pepsico/projections/maproom.py
```diff
@@ -338,28 +338,31 @@ def seasonal_change(
...
         data = ac.seasonal_data(
             ac.read_data(scenario, model, variable, region, unit_convert=True),
             start_month, end_month,
             start_year=start_year, end_year=end_year,
-        ).mean(dim="T")
-        data = data - ref
+        ).mean(dim="T", keep_attrs=True)
+        #Tedious way to make a subtraction only to keep attributes
+        data = xr.apply_ufunc(
+            np.subtract, data.load(), ref.load(), keep_attrs="drop_conflicts"
```
aaron-kaplan, line 349: It doesn't work without the `.load()`?

aaron-kaplan (approved):

pepsico/projections/maproom.py
remicousin: no. It returns:

  File "/home/remic/miniconda3/envs/enactsmaproom/lib/python3.9/site-packages/xarray/core/computation.py", line 692, in apply_variable_ufunc
    raise ValueError(
ValueError: apply_ufunc encountered a dask array on an argument, but handling for dask arrays has not been enabled. Either set the ``dask`` argument or load your data into memory first with ``.load()`` or ``.compute()``

aaron-kaplan: Thanks, I didn't know that. Looks like you could use `dask="allowed"` instead of multiple calls to `.load`. I don't care if you change it or not.

https://docs.xarray.dev/en/stable/generated/xarray.apply_ufunc.html

remicousin: I applied the dask option rather. Also note that there is [ufuncs.subtract](https://docs.xarray.dev/en/latest/generated/xarray.ufuncs.subtract.html) too but it dones't look like I could apply keep_attrs to that, does it?

remicousin: Resolving this and merging. Please reopen if it turns out there is something about ufuncs.subtract

Issue #364, opened 2023-07-13
Fill false is enough for all Maprooms


remicousin:
as in [here](https://github.com/iridl/python-maprooms/pull/280/commits/368e6a82333fa6df029b6f63b3b70debed5aba18)

remicousin:
Resolved by PR #477

PR #477, opened 2025-01-30
sufficient set up per CSC case

remicousin:
To resolve issue #364


aaron-kaplan:
I don't know the context, and the linked issue doesn't illuminate it at all (it just has a link to a commit that no longer exists).

remicousin:
Right, the Issue is useless now for some reason.
I think it's just that
fill = False
is equal to:
fill = True AND Opacity = 0

xchourio (approved):

aaron-kaplan (approved):

PR #478, opened 2025-01-31
Downloadcsv

remicousin:
Series of commits to provide a download button to download csv identical to data in ts graph.

### DIFF
--- a/pepsico/projections/layout.py
+++ b/pepsico/projections/layout.py
@@ -283,7 +283,10 @@ def map_layout():
 
 
 def results_layout():
-    return dbc.Tabs([dbc.Tab(
-        [dbc.Row([dbc.Col(dbc.Spinner(dcc.Graph(id="local_graph"),)),])],
-        label="Local History and Projections",
-    )], className="mt-4")
+    return dbc.Tabs([dbc.Tab([
+        html.Div([
+            dbc.Button("Download data in graph", id="btn_csv", size="sm"),
+            dcc.Download(id="download-dataframe-csv"),
+        ], className="d-grid justify-content-md-end"),
+        html.Div([dbc.Spinner(dcc.Graph(id="local_graph"))]),
+    ], label="Local History and Projections")])
--- a/pepsico/projections/layout.py
+++ b/pepsico/projections/layout.py
@@ -283,22 +283,7 @@ def map_layout():
 
 
 def results_layout():
-    return dbc.Tabs(
-        [
-            dbc.Tab(
-                [
-                    dbc.Row(
-                        [
-                            dbc.Col(
-                                dbc.Spinner(
-                                    dcc.Graph(id="local_graph"),
-                                )
-                            ),
-                        ]
-                    )
-                ],
-                label="Local History and Projections",
-            )
-        ],
-        className="mt-4",
-    )
+    return dbc.Tabs([dbc.Tab(
+        [dbc.Row([dbc.Col(dbc.Spinner(dcc.Graph(id="local_graph"),)),])],
+        label="Local History and Projections",
+    )], className="mt-4")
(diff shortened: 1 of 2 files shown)


pepsico/projections/maproom.py
```diff
@@ -164,6 +164,42 @@ def pick_location(n_clicks, click_lat_lng, region, latitude, longitude):
...
+        })
+        error_msg = None
+        missing_ds = xr.Dataset()
+        try:
+            if any([var is None for var in data_ds.data_vars.values()]):
+                data_ds = missing_ds
+                error_msg="Data missing for this model or variable"
+            else:
+                data_ds = pingrid.sel_snap(data_ds, lat, lng)
+        except KeyError:
```
aaron-kaplan, line 193: As far as I can see, KeyError can only be raised in line 192. The purpose of the try/except would be clearer if you narrowed its scope to that one line.

pepsico/projections/maproom.py
```diff
@@ -164,6 +164,42 @@ def pick_location(n_clicks, click_lat_lng, region, latitude, longitude):
...
+            ),
+            "ssp585" : ac.read_data(
+                "ssp585", model, variable, region, unit_convert=True,
+            ),
+        })
+        error_msg = None
+        missing_ds = xr.Dataset()
+        try:
+            if any([var is None for var in data_ds.data_vars.values()]):
+                data_ds = missing_ds
```
aaron-kaplan, line 189: This is redundant, as it will happen again in line 199.

pepsico/projections/maproom.py
```diff
@@ -164,6 +164,42 @@ def pick_location(n_clicks, click_lat_lng, region, latitude, longitude):
...
+                error_msg="Data missing for this model or variable"
+            else:
+                data_ds = pingrid.sel_snap(data_ds, lat, lng)
+        except KeyError:
+            data_ds = missing_ds
```
aaron-kaplan, line 194: Redundant

pepsico/projections/maproom.py
```diff
@@ -176,6 +212,59 @@ def plot_ts(ts, name, color, start_format, units):
...
+            file_name = (
+                f'{data_ds["histo"]["T"].dt.strftime("%b")[0].values}-'
+                f'{data_ds["histo"]["seasons_ends"].dt.strftime("%b")[0].values}'
+                f'_{variable}_{model}_{abs(lat)}{lat_units}_{abs(lng)}{lng_units}'
+                f'.csv'
+            )
+            df = data_ds.to_dataframe()
+        else :
+            file_name = "nodata.csv"
+            df = pd.DataFrame()
```
aaron-kaplan, line 264: Seems like it would be more appropriate to display an error message rather than downloading a file that we know will be empty. (Not a big deal.)

pepsico/projections/layout.py
```diff
@@ -283,22 +283,10 @@ def map_layout():
...
-                            ),
-                        ]
-                    )
-                ],
-                label="Local History and Projections",
-            )
-        ],
-        className="mt-4",
-    )
+    return dbc.Tabs([dbc.Tab([
```
aaron-kaplan, line 286: I don't understand why we're using Tabs if there's only one tab. Do you plan to add more?

pepsico/projections/maproom.py
remicousin: I didn't realize I could return None. That is done. As for the error message: it's printed on the empty graph, and the Download button is disabled, so we should be good here.

pepsico/projections/layout.py
remicousin: Some other Maprooms have more than one. This is just repeating the Structure. I could remove it but I'd still want its label to indicate the "Local" section without jumping directly to the figure title, and without clogging further the figure title with information should there be no intermediary. So might as well keep the lonely Tab?

PR #480, opened 2025-02-05
turns out this is where logos are meant to come from

remicousin:


aaron-kaplan:
This might be how it works for the existing maprooms (I haven't checked), but it's not a good way to do it. This repo shouldn't contain partner-specific resources, just the code that runs at all sites. I'll come back and make a suggestion later.

remicousin:
Maybe it's not. I took from Guyana case and it's not showing since there is not Guyana logo there. Just assumed it was lost in transition.

KMD case works
http://kmddl.meteo.go.ke:8081/python_maproom/crops_climate_suitability/
and points to /python_maproom/static/
so still in non partner specific location.

aaron-kaplan:
The [python_maproom_kmd](https://bitbucket.org/iridl/python_maproom_kmd/src/master/) repo contains the logo file, and the Dockerfile [copies that image into the container](https://bitbucket.org/iridl/python_maproom_kmd/src/9f9b15f2cf8ff07cf664af749c00963504760feb/Dockerfile#lines-4).

PR #492, opened 2025-02-21
Switch from conda to pixi

aaron-kaplan:
[Pixi](https://github.com/prefix-dev/pixi) is an up-and-coming competitor to conda that uses the same packages from conda-forge (and potentially PyPI) but is much faster, easier to boostrap (it's a native executable, not a python program, so you don't need to install python to install conda to install python), has lockfile functionality that's built in (in contrast to conda-lock which is a separate package) and better, and generally seems to be a lot more stable and robust.

The immediate impetus for this change is that I had to build a new image while on the NiMet call and it took something like 20 minutes. Now with pixi it takes 1m10s.

I'm submitting this as a draft because I haven't been able to test it yet. I went through six different partners' python_maproom_* repos and couldn't find one with a config-iri.yaml that worked with the current version of enactsmaproom. We should set up a test country whose config we keep in this repo for testing and development purposes.


aaron-kaplan:
One difference from conda is that pixi creates the python environment in the project directory, whereas conda puts it in `$CONDA_HOME/envs`. There are pros and cons to this. For shared environments on a server we should probably use [detached environments](https://pixi.sh/dev/reference/pixi_configuration/#detached-environments).

aaron-kaplan:
I added this file exclusion to my CrashPlan settings: `.*/\.pixi`. Is there a way to tell if it's working?

aaron-kaplan:
I know how to test fbfmaproom, so I've switched my packaging improvements to that for now. Once I'm satisfied with the setup there, I'll come back and apply it to enactsmaproom. https://github.com/iridl/python-maprooms/pull/493

jturmelle:
The easiest way would be to try a restore and see if it’s there.  I think you only need to say \.pixi though.

PR #493, opened 2025-02-22
Update fbfmaproom2 image to CentOS Stream 9 and pixi

aaron-kaplan:
For an explanation of pixi, see https://github.com/iridl/python-maprooms/pull/492, which does the same for enactsmaproom. Unlike that PR, which I haven't been able to test yet, this one is tested and ready to go.

If the Stream 10 part looks ok but there are reservations about pixi, I can back out the pixi commit and put that in a separate PR.

Tested as follows, on docker.iri:
```
docker run \
  --rm \
  -it \
  --publish 8000:8000 \
  -v /data/aaron/fbf-candidate:/data/aaron/fbf-candidate \
  -v $(pwd):/config \
  -e CONFIG=/config/fbfmaproom-sample.yaml:/config/fbfmaproom-override.yaml \
  sha256:abcde
```
with `abcde` replaced by the hash of the image I built from this branch.


jturmelle (commented):
When I tried to build the CentOS10Stream Docker container on a very old server (2011), I couldn't build it.  We need to add comments somewhere that servers need to have chips that are at least the Haswell generation, which is probably > 2015, to be safe.  For example, this won't be able to run on our dlcompute servers or our older frontsquid server iridls1.  Not that I think this is a problem, just something we need to be make note of.

```
 > [common 2/3] RUN dnf4 install -y httpd:
0.840 Fatal glibc error: CPU does not support x86-64-v3

```

fbfmaproom/pixi.toml
```diff
@@ -0,0 +1,47 @@
...
+
+[tasks]
+
+[system-requirements]
+linux = "3.10.0"
+
+[dependencies]
+# rasterio build fails with python 3.10
+# https://github.com/rasterio/rasterio/issues/2333
+python = "3.9.*"
```
jturmelle, line 15: I'm confused by this because it looks like python3.12 is installed with CentOS10-Stream?  Does pixi download and install python3.9?

jturmelle (approved):
I don't completely understand the GitHub workflows, and don't completely understand pixi yet, but I get the gist of it and everything seems alright. But if we implement this, we'll have to take iridls1 out of the server mix.  We still have 2 other real servers with sufficient CPUs.

fbfmaproom/pixi.toml
aaron-kaplan: Yes, both conda and pixi manage the installation of python itself as well as python packages.

Conda is a python program, which makes bootstrapping complicated--you have to install python in order to run conda in order to install python. Pixi is a single-file executable compiled from rust, which makes bootstrapping easier.

aaron-kaplan:
> if we implement this, we'll have to take iridls1 out of the server mix.

The current PR only affects fbfmaproom2, which doesn't run on the squid server, so that's not an issue. When we get around to upgrading the squid image, we can keep that in mind.

aaron-kaplan:
On the other hand, we're planning on deploying fbfmaproom2 to partner servers in the coming year, so we might have to hold off if any of them are running on older processors.

I think I'll back off to Stream 9. It's easier than doing the legwork to find out who has what CPU, and I'm hoping to move to more minimalist images long before Stream 9 EOL.

jturmelle:
> Now using Stream 9. @jturmelle could you test this on the old server?

Yes, it worked.  Although it occurs to me you said you were going to change this from apache to gunicorn, so maybe it wasn't going to be an issue anyway.

aaron-kaplan:
I think conda packages are dynamically linked against glibc, meaning even if we switch to gunicorn we'll be using the system glibc, so if it doesn't support our CPU then we're out of luck.

aaron-kaplan:
...and while I'm excited to switch to gunicorn, it requires some testing and performance tuning, and it's hard to justify spending time on that right now, so I'm forcing myself to stop here for now.

aaron-kaplan:
Regarding the github action, the miniconda stuff I deleted is complicated, but the new one is pretty simple ([link to the new version](https://github.com/iridl/python-maprooms/blob/afd436de372e320373df043be2ec0da6ccede959/.github/workflows/fbf-test.yaml)). There are three steps: check out the project, create a python environment from the lock file using pixi, and run the tests. `pixi run x` just runs shell command `x` in the pixi environment.

aaron-kaplan:
I deployed to prod and the container failed to start. log says
```
2904 2025] [mpm_prefork:notice] [pid 1:tid 1] AH00163: Apache/2.4.62 (CentOS Stream) mod_wsgi/4.7.1 Python/3.9 configured -- resuming normal operations
3170 2025] [core:notice] [pid 1:tid 1] AH00094: Command line: '/usr/sbin/httpd -D FOREGROUND'
5962 2025] [wsgi:alert] [pid 15:tid 15] (1)Operation not permitted: mod_wsgi (pid=15): Couldn't create monitor thread in daemon process 'fbfmaproom'.
5960 2025] [wsgi:alert] [pid 11:tid 11] (1)Operation not permitted: mod_wsgi (pid=11): Couldn't create monitor thread in daemon process 'fbfmaproom'.
6052 2025] [wsgi:alert] [pid 11:tid 11] (1)Operation not permitted: mod_wsgi (pid=11): Couldn't create deadlock thread in daemon process 'fbfmaproom'.
6052 2025] [wsgi:alert] [pid 15:tid 15] (1)Operation not permitted: mod_wsgi (pid=15): Couldn't create deadlock thread in daemon process 'fbfmaproom'.
6177 2025] [wsgi:alert] [pid 11:tid 11] (1)Operation not permitted: mod_wsgi (pid=11): Couldn't create worker thread 0 in daemon process 'fbfmaproom'.
6181 2025] [wsgi:alert] [pid 15:tid 15] (1)Operation not permitted: mod_wsgi (pid=15): Couldn't create worker thread 0 in daemon process 'fbfmaproom'.
...
```
It worked when I tested it on docker.iri, which is centos 7.9.2009 and docker 25.0.1. It doesn't work on dlcomputegeo1, which is centos 7.6.1810 and docker 1.13.1. @jturmelle why does docker.iri have a different version of docker?

aaron-kaplan:
I can reproduce the failure without interrupting the production service by running this on the command line on geo1:
```
docker run iridl/fbfmaproom:2.52.0
```

aaron-kaplan:
`aureport -f` reports no failures.

jturmelle:
> I deployed to prod and the container failed to start. log says
> 
> ```
> 2904 2025] [mpm_prefork:notice] [pid 1:tid 1] AH00163: Apache/2.4.62 (CentOS Stream) mod_wsgi/4.7.1 Python/3.9 configured -- resuming normal operations
> 5962 2025] [wsgi:alert] [pid 15:tid 15] (1)Operation not permitted: mod_wsgi (pid=15): Couldn't create monitor thread in daemon process 'fbfmaproom'.
> ...
> ```
> 
> It worked when I tested it on docker.iri, which is centos 7.9.2009 and docker 25.0.1. It doesn't work on dlcomputegeo1, which is centos 7.6.1810 and docker 1.13.1. @jturmelle why does docker.iri have a different version of docker?

I'm not sure why they're different, I'm looking into the issue now.

aaron-kaplan:
I meant to do aureport --failed. I'm not sure how to interpret the output of that, but I can't find anything in audit.log that seems relevant.

aaron-kaplan:
Running it with `docker run --privileged` eliminates the errors. Of course, we don't want to do that in production.

jturmelle:
Maybe we should just upgrade the OS and docker version on geo1.

aaron-kaplan:
I think the old version of docker is causing other problems too (can't pull certain images), so yeah, but we have to be careful about it because it could easily stop working entirely. Maybe snapshot dlcomputemon3-vm and upgrade that first, because it's easy to roll back?

strace shows a file access failing with "Permission denied." Then I ran bash in the image and got this interesting result:
```
(fbfmaproom2:prod) pwd
/app
(fbfmaproom2:prod) ls
Dockerfile  __about__.py  activate  data-conversion-scripts  fbf-update-data.py  fbfmaproom-sample.yaml  fbfmaproom.wsgi  forecasts  pixi.lock  pyproject.toml           tests
README.md   __pycache__   assets    docker                   fbflayout.py        fbfmaproom.py           fbftable.py      pingrid    pixi.toml  release_container_image
(fbfmaproom2:prod) ls /app
ls: cannot access '/app': Operation not permitted
(fbfmaproom2:prod) ls /
ls: cannot access '/': Operation not permitted
```
Happens even if I run the container as root. But if I run it with --privileged, I can `ls /`. So something wrong with permissions on the root of the container's filesystem.

jturmelle:
> Maybe snapshot dlcomputemon3-vm and upgrade that first, because it's easy to roll back?
> 
Good idea.  I’ll give it a try.

aaron-kaplan:
Let's upgrade to CentOS 9 via ansible, not do something ad hoc. There's no rush, I've already rolled back the fbfmaproom change.

jturmelle:
> Let's upgrade to CentOS 9 via ansible, not do something ad hoc. There's no rush, I've already rolled back the fbfmaproom change.

OK, I thought it was a priority.  I've just upgraded dlcomputemon3-vm to the final version of CentOS7 and rebooted.  I hadn't upgraded docker yet though.  If it's not working, I have a snapshot I can revert to.

jturmelle:
Just tested the playbook and it ran with no real changes and all the tests came back positive..

aaron-kaplan:
That's good. The problem persists, but that's ok for now.

PR #494, opened 2025-02-22
pixi for fbf

aaron-kaplan:


aaron-kaplan:
Combined into https://github.com/iridl/python-maprooms/pull/493

PR #495, opened 2025-02-26
Mali Maproom Creation

nitinmagima:
Hi Aaron,

I've created the Mali Maproom based in Niger.  The forecast needs to be updated since it's being cut off a bit so I'll update that later. Overall it's working, however, I'm unable to convert the bad years properly, can you tell me what I'm doing wrong? I've added the data and the code I'm using to do this as well.


fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -1906,6 +1906,160 @@ countries:
...
+                  where adm0_name = 'Mali'
+              vuln_sql: select (adm0_code, adm1_code, adm2_code) as key, year, sum(vulnerability) as vuln
+                  from public.fbf_vulnerability where adm0_name = 'Mali' group by 1, 2
+
+        datasets:
+            defaults:
+              predictors:
+                - prcp-v1
+                - chirps-precip-jas
+              predictand: bad-years-emdat
```
aaron-kaplan, line 1942: This has to refer to  one of the obs datasets, but there's no dataset with this name. In line 2027 you've named it `bad-years`.

fbfmaproom/fbf-update-data.py
```diff
@@ -341,6 +341,34 @@
         "niger/wrsi-jas",
         "http://iridl.ldeo.columbia.edu/SOURCES/.USGS/.EROS/.FEWS/.dekadal/.WAF/.Millet/.do/X/0.125/15.875/RANGE/Y/11.125/16.875/RANGE/T/(Jul-Sep)/seasonalAverage/T/(months%20since%201960-01-01)/streamgridunitconvert/",
     ),
+    (
+        'mali/chirps-precip-jun',
```
aaron-kaplan, line 345: Please name this `mali/chirps-precip-jun-jas`, to indicate that the T values have been faked (shiftGRID) to make it look like JAS. See e.g. https://github.com/iridl/python-maprooms/blob/master/fbfmaproom/fbfmaproom-sample.yaml#L885 for example.

Likewise for the other time-shifted datasets below.

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -1906,6 +1906,160 @@ countries:
                 issue_months: [0, 1, 2, 3, 4, 5]
                 start_year: 1991
 
+    mali:
+        logo: Mali_IRI_120x48.png
```
aaron-kaplan, line 1910: Do you have a logo file? I have to add that to the server (separately, through a different mechanism).

fbfmaproom/data-conversion-scripts/mali/bad_years.py
```diff
@@ -0,0 +1,13 @@
...
+import xarray as xr
+
+bad = pd.read_csv('bad_years_emdat.csv')
+
+years = [ cftime.Datetime360Day(y, 8, 16) for y in bad.iloc[:,0].to_list() ]
+ranks = bad.iloc[:,1].to_list()
+
+ds = xr.Dataset(data_vars={"bad": xr.DataArray(ranks, coords={"T": years})})
+
+ds.to_zarr('/data/aaron/fbf-candidate/mali/bad-years-emdat.zarr', mode='w')
```
aaron-kaplan, line 13: Sorry I couldn't remember how to do this on the call yesterday. Had to refresh my memory.

There's actually a generic script [data-conversion-scripts/zarrify-bad-years.py](https://github.com/iridl/python-maprooms/blob/master/fbfmaproom/data-conversion-scripts/zarrify-bad-years.py), so we don't need to write a new script for each country any more.

That script was added last September, but you don't have it in this branch, so you must not have pulled the latest changes from github since some time before that. In future, whenever you create a new branch you should do it as follows (`#` indicates comments):
```
git fetch              # gets the latest version from github, sets origin/master to point to that
git checkout origin/master
git checkout -b mali_v1   # creates a new branch called mali_v1 starting from the latest version
```
In this case, where you've already created the branch from an old version, I suggest the following to update it:
```
git fetch
git checkout mali_v1
git merge origin/master
```

The script expects `bad_years_emdat.csv` to be at `/data/aaron/fbf-candidate/original_data/mali/bad_years_emdat.csv` (on the server). Then it looks like you would run it like this:
```
activate enactsmaproom
cd data-conversion-scripts
python zarrify-bad-years.py mali/bad_years_emdat.csv 7.5
```
where 7.5 is the center of the target season, as you defined it in line 2057 of the config file.

That was a bit complicated. Let me know if you want to get on a call and go through it together.

aaron-kaplan (commented):
Nice work. See comments below.

aaron-kaplan:
Superseded by https://github.com/iridl/python-maprooms/pull/499

PR #496, opened 2025-02-27
Back to CentOS 7

aaron-kaplan:
The CentOS 9 container wouldn't run on geo1, so back to CentOS 7 until we can upgrade the servers. I'm keeping pixi.

### DIFF
--- a/fbfmaproom/Dockerfile
+++ b/fbfmaproom/Dockerfile
@@ -1,12 +1,11 @@
-# centos:stream9 as of 2025-02-26
-ARG CENTOS_HASH="sha256:f9ac4692a0505202cb05380e3c69461320358db735288c798689502d903dbe37"
-
-# Dependencies that are used in both build stage and final stage
-FROM quay.io/centos/centos@${CENTOS_HASH} as common
+FROM centos:7.9.2009 as common
 
 ARG PIXI_VERSION="v0.41.4"
 
-RUN dnf install -y httpd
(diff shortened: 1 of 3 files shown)


jturmelle (approved):
That’s interesting  the centos7 version still works.

jturmelle (approved):
Do we need to, or should we, update geo1 to the most recent version of centos7?

aaron-kaplan:
I'm actually not sure what "the most recent version of centos7" means. Is
installing a newer version of the same release the same as installing an
older version and then running yum update, or is it different? We should
keep our servers up to date with the latest bug and security patches, but
not make optional upgrades needlessly.

jturmelle:
> I'm actually not sure what "the most recent version of centos7" means. Is installing a newer version of the same release the same as installing an older version and then running yum update, or is it different? We should keep our servers up to date with the latest bug and security patches, but not make optional upgrades needlessly.

Yes, I just meant a yum update, to bring it to the final release of CentOS7.  But if it's working and we're upgrading to CentOS9, then I won't bother.

aaron-kaplan:
> That’s interesting the centos7 version still works.

Indeed, it confirms that the relevant difference is the base image, not what version of CentOS and Docker I used to build the derived image.

aaron-kaplan:
FYI I had forgotten to commit CentOS-Vault.repo. Pushed that directly to master after merging: 487b589f6fcb5773b705f30cab141df1c383b882

jturmelle:
I noticed that last night.  I was looking all over for that file.

PR #497, opened 2025-02-27
Create conda environment descriptors from pixi.lock

aaron-kaplan:
I've switched to pixi for managing this project's dependencies, but I don't want to tell all the developers to switch to pixi until (a) I've had more experience with it and I'm confident that it's going to work out, and (b) all our projects are moved over, so people don't have to deal with both, and remember which project uses which. In the meantime, I've written this simple script that translates the pixi lock file into conda lock files. This way I can use pixi's more convenient tools for adding and updating dependencies, and others can continue using conda for creating environments.


jturmelle (approved):

aaron-kaplan:
@jturmelle could you update the fbfmaproom environment on mako to use the packages in this new conda-linux-64.lock?

jturmelle:
> @jturmelle could you update the fbfmaproom environment on mako to use the packages in this new conda-linux-64.lock?

It's building now: fbfmaproom2.52

aaron-kaplan:
Please call it fbfmaproom2. People shouldn't have to check what the latest
version number is in order to activate the environment. My thinking is that
the globally-installed environment should always match what's in github
master. Developers who need to use an earlier version or are working on a
future version will create their own environment.

-A

aaron-kaplan:
Unfortunately you can't rename a conda environment, because the path is
hard-coded in the executables. You'll have to delete and recreate it.

jturmelle:
OK, I didn't know I couldn't rename the environment.  Nitin is working on mako right now, so it may need to wait.

PR #505, opened 2025-03-06
Regwatbal

remicousin:
Same as the other two but for water balance. Done up to the layout config step

### DIFF
--- a/enacts/config-test-enacts.yaml
+++ b/enacts/config-test-enacts.yaml
@@ -91,9 +91,7 @@ maprooms:
               map_max: 40
 
     wat_bal:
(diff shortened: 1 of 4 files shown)


aaron-kaplan:
Code looks ok. Need a config file to test with. Maybe merge this branch into one of the others that have a test config?

remicousin:
ok. I rebased to have and then add to the test config file, and finished the set up via testing. There were the config entries in layout to port to callback and some debugging.

aaron-kaplan (commented):
As for CSC, mousing over regions doesn't display place names. Looks ok other than that.

PR #513, opened 2025-03-14
Splitenactsread

remicousin:
The first commit splits the reading business from the Maprooms for ENACTS data only.

The second commit introduces the case of synthesizing daily ENACTS data if no data to read. Introducing a new test config file for onset and CSC (because monthly uses dekadal and water_balance needs also fake soil data). But the whole can be tested on the Senegal test case to test the 1st commit.

At the moment, I can't guarantee tmin < tmax since they are generated independently... Something to think about how much it matters (the functions that care about it might be dealing with the case...)

Next steps would be a function to make dekadal out of daily and test the case and apply to monthly; then fake soil data to apply to wat_bal; then consider what to do with respect to shapefiles to be entirely independent.

Then I guess the synthetic test config file case could replace the data reading one...

### DIFF
--- a/enacts/calc.py
+++ b/enacts/calc.py
@@ -7,7 +7,7 @@ import datetime
 
 np.random.seed(123)
 
-def read_enacts_zarr_data(
+def read_enacts_data(
     variable="precip", time_res="dekadal", ds_conf=None, as_array=True
 ):
     """ Read ENACTS zarr data and return `xr.Dataset` or `xr.DataArray`
--- a/enacts/calc.py
+++ b/enacts/calc.py
@@ -32,7 +32,7 @@ def read_enacts_data(
     --------
     xr.open_zarr
     """
-    if ds_conf[time_res] is None:
+    if ds_conf is None:
         # Center mu, amplitude amp of the base sinusoid
         # and amplitude of noisy anomalies to apply to it
         characteristics = {
--- a/enacts/calc.py
+++ b/enacts/calc.py
@@ -7,17 +7,53 @@ import datetime
 
 np.random.seed(123)
 
-def read_enacts(variable, ds_conf="dekadal"):
+
+def find_enacts_zarr_path(variable, ds_conf, time_res="daily"):
+    """ Find path and name of ENACTS zarr variable if any
+
+    Parameters
+    ----------
+    variable : str
+        string representing ENACTS variable (precip, tmin or tmax)
+    ds_conf : dict
+        dictionary indicating ENACTS zarr path (see config)
+    time_res : str, optional
+        "daily" or "dekadal" resolution of the desired variable
+        defautl is "daily"
+    
+    Returns
+    -------
+        path to feed `xr.open_zarr` or None
+    
+    See Also
+    --------
+    xr.open_zarr
+    """
+    try:
+        data_path = ds_conf[time_res]['vars'][variable][1]
+        if data_path is None:
+            data_path = ds_conf[time_res]['vars'][variable][0]
+        zarr_path = f"{ds_conf[time_res]['zarr_path']}{data_path}"
+        var_name = ds_conf[time_res]['vars'][variable][2]
+    except KeyError:
+        zarr_path = None
+        var_name = None
+    return zarr_path, var_name
+
+
+def read_enacts(variable, ds_conf=None, time_res="daily"):
     """ Read ENACTS zarr data and return `xr.DataArray`
 
     Parameters
     ----------
     variable : str
         string representing ENACTS variable (precip, tmin or tmax)
-    ds_conf: str or dict, optional
-        string "daily" or "dekadal" indicating time resolution of synthetic data
-        or dictionary indicating ENACTS zarr path (see config)
-        default is synthetic dekadal data
+    ds_conf : dict, optional
+        dictionary indicating ENACTS zarr path (see config)
+        default is None to produce synthetic data
+    time_res : str, optional
+        "daily" or "dekadal" resolution of the desired variable
+        defautl is "daily"
     
     Returns
     -------
@@ -27,7 +63,8 @@ def read_enacts(variable, ds_conf="dekadal"):
     --------
     xr.open_zarr
     """
-    if not isinstance(ds_conf, dict):
+    data_path, var_name = find_enacts_zarr_path(variable, ds_conf, time_res=time_res)
+    if data_path is None:
         # Center mu, amplitude amp of the base sinusoid
         # and amplitude of noisy anomalies to apply to it
         characteristics = {
@@ -75,12 +112,8 @@ def read_enacts(variable, ds_conf="dekadal"):
         ).interp(X=np.arange(-55, -51, 0.0375), Y=np.arange(2, 6.5, 0.0375))
         var_name = variable
     else:
-        data_path = ds_conf['vars'][variable][1]
-        if data_path is None:
-            data_path = ds_conf['vars'][variable][0]
-        xrds = xr.open_zarr(f"{ds_conf['zarr_path']}{data_path}")
-        var_name = ds_conf['vars'][variable][2]
-    return xrds[array]
+        xrds = xr.open_zarr(data_path)
+    return xrds[var_name]
 
 # Growing season functions
 
--- a/enacts/calc.py
+++ b/enacts/calc.py
@@ -30,7 +30,7 @@ def read_enacts_data(
     
     See Also
     --------
-    read_zarr_data
+    xr.open_zarr
     """
     if ds_conf[time_res] is None:
         # Center mu, amplitude amp of the base sinusoid
@@ -83,30 +83,9 @@ def read_enacts_data(
         data_path = ds_conf[time_res]['vars'][variable][1]
         if data_path is None:
             data_path = ds_conf[time_res]['vars'][variable][0]
-        xrds = read_zarr_data(f"{ds_conf[time_res]['zarr_path']}{data_path}")
+        xrds = xr.open_zarr(f"{ds_conf[time_res]['zarr_path']}{data_path}")
         array = ds_conf[time_res]['vars'][variable][2]
     return xrds[array] if as_array else xrds
-
-
-def read_zarr_data(zarr_path):
-    """Read and return data in zarr format.
-
-    Parameters
-    ----------
-    zarr_path : str
-        String of path to zarr folder.
-    Returns
-    -------
-    zarr_data : Dataset
-        Data from zarr folder as multidimensional xarray dataset.
-    See Also
-    --------
-    Notes
-     -----
-    """
-    zarr_data = xr.open_zarr(zarr_path)
-    return zarr_data
-
 
 # Growing season functions
 
--- a/enacts/calc.py
+++ b/enacts/calc.py
@@ -26,10 +26,10 @@ def get_data(variable, time_res, ds_conf):
     
     See Also
     --------
-    read_enacts, synthetize_enacts
+    read_enacts, synthesize_enacts
     """
     if ds_conf[time_res] == "FAKE" :
-        return synthetize_enacts(variable, time_res)
+        return synthesize_enacts(variable, time_res)
     else:
         return read_enacts(variable, ds_conf[time_res])
 
@@ -62,7 +62,7 @@ def read_enacts(variable, dst_conf):
     return xrds[var_name]    
 
 
-def synthetize_enacts(variable, time_res):
+def synthesize_enacts(variable, time_res):
     """ Synthetize ENACTS data as `xr.DataArray`
 
     Parameters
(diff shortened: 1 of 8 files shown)


enacts/monthly/maproom.py
```diff
@@ -91,7 +82,7 @@ def get_shapes(query):
         Input("location", "pathname"),
     )
     def initialize(path):
-        rr_mrg = read_data("precip")
+        rr_mrg = calc.read_enacts_zarr_data(**READ_PARAMS)
```
aaron-kaplan, line 85: I think I would prefer to make `variable` mandatory, rather than defaulting to `precip`. This call might be misunderstood to be fetching all the data (all variables). What's your justification for defaulting to `precip`?

enacts/monthly/maproom.py
```diff
@@ -170,25 +163,26 @@ def update_map(variable, month):
...
     def pick_location(click_lat_lng):
         if click_lat_lng == None:
-            rr_mrg = read_data("precip")
+            rr_mrg = calc.read_enacts_zarr_data(**READ_PARAMS)
             return [
                 ((rr_mrg["Y"][int(rr_mrg["Y"].size/2)].values)),
                 ((rr_mrg["X"][int(rr_mrg["X"].size/2)].values)),
             ]
-        return click_lat_lng                           #  in the data to where the user clicked on the map.
+        return click_lat_lng # in the data to where the user clicked on the map.
```
aaron-kaplan, line 171: I don't understand this comment (the "in the data" part). Do you? Seems like it might have been the second line of a two-line comment, and the first line got deleted.

aaron-kaplan (commented):
A clear improvement. Minor comments below.

enacts/calc.py
```diff
@@ -1,13 +1,98 @@
...
+    
+    Returns
+    -------
+        `xr.Dataset` or `xr.DataArray` of ENACTS `variable` at `time_res` time steps
+    
+    See Also
+    --------
+    read_zarr_data
+    """
+    if ds_conf[time_res] is None:
```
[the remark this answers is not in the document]
remicousin: Right... that's a leftover of my misunderstanding of what the absence of config entries returned... Should I use a try/except instead of if/else?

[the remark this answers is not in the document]
aaron-kaplan: Don't use try/except. The simplest thing would be to change this to
```
if ds_conf is None:
```
but if you want to be able to use real data for one variable and fake data for another in the same configuration, you could make `ds_conf` mandatory, but potentially empty, and keep this test as-is.

enacts/monthly/maproom.py
remicousin: So... probably just my ignorance. I wanted arguments to be keyword arguments so that I can define them as a dictionary so that I can pass them on as parameters through the **{} calling method. If keyword, I have to give it a default value, don't I? And in that case, precip is used everywhere and was thus the more efficient/logical default.

enacts/calc.py
```diff
@@ -1,13 +1,98 @@
 import numpy as np
 import pandas as pd
 import xarray as xr
+import datetime
 
 # Date Reading functions
+
+np.random.seed(123)
+
+def read_enacts_zarr_data(
```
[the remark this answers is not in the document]

enacts/calc.py
```diff
@@ -1,13 +1,98 @@
...
+    else:
+        data_path = ds_conf[time_res]['vars'][variable][1]
+        if data_path is None:
+            data_path = ds_conf[time_res]['vars'][variable][0]
+        xrds = read_zarr_data(f"{ds_conf[time_res]['zarr_path']}{data_path}")
+        array = ds_conf[time_res]['vars'][variable][2]
+    return xrds[array] if as_array else xrds
+
+
 def read_zarr_data(zarr_path):
```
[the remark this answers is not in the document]

enacts/calc.py
```diff
@@ -1,13 +1,98 @@
...
+
+def read_enacts_zarr_data(
+    variable="precip", time_res="dekadal", ds_conf=None, as_array=True
+):
+    """ Read ENACTS zarr data and return `xr.Dataset` or `xr.DataArray`
+
+    Parameters
+    ----------
+    variable : str, optional
+        string reprensenting ENACTS variable (precip (default), tmin or tmax)
```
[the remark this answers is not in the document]

enacts/config-test-synt.yaml
```diff
@@ -0,0 +1,59 @@
...
+          color: black
+          sql: select adm1_code as key, adm0_name as label, adm0_code, ST_AsBinary(the_geom) as the_geom
+              from g2015_2014_1 where adm0_code=86
+          is_checked: True
+    daily: null
+
+maprooms:
+    # Onset
+    onset:
+    - default_search_month: Jan
```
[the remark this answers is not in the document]
remicousin: done

enacts/monthly/maproom.py
remicousin: done

remicousin:
In the past days, I came to thinking that instead of using synthetic data, we could use "compressed" data. That is to say take 10 years or coarse daily ENACTS data, compute its SVD or Fourier Transform, keep only enough modes or harmonics to keep enough variance (possibly keep less and add random noise when recomposing the data). Save that in gitHub: recompose data with given modes/harmonics (add random noise), interpolate to finer grid if need be.

This may make for more robust fake data. My playing with the entirely synthetic is borderline irrealistic and while it still allows to compute onset, cessation and CSC (which surpirses me a bit), the next fancy thing it will need to calculate might blow up...

The question is how big the compressed data would be... And I assume there is what's needed in xarray/scipi/numpy/etc. to compute either SVD or Fourier, or something that can serve the same purpose, or enough to write them...

enacts/calc.py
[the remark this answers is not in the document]
remicousin: done. Then I guess the data reading case could face abruptly if the config is not as it should but maybe that is something to live with until better days.

enacts/calc.py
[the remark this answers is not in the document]
aaron-kaplan: I suggested `read_enacts`, dropping the `_data`, because I don't think it adds anything. Take it or leave it, I won't insist.

enacts/monthly/maproom.py
aaron-kaplan: Positional arguments can be provided as keyword arguments.
```
>>> def foo(a, b):
...     return b - a
... 
>>> foo(b=2, a=1)
1
```

aaron-kaplan:
Interesting idea about using real but highly-compressed data. Something to keep in our back pocket in case we need it.

enacts/calc.py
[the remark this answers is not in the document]
remicousin: I had to revisit this as ds_config, who was GLOBAL_CONFIG['datasets'] can not be None as of now because it also has the shapes information. I had it be GLOBAL_CONFIG['datasets']['daily'] or ['dekadal'] but I also need that time resolution information for the synthetic case to know what to return and so I had these be strings rather than dict in the config is test against that...
It's kind of messy right now but it should clear up once there is also synthetic or "fake" shapefile data?

enacts/calc.py
```diff
@@ -1,27 +1,88 @@
...
 import xarray as xr
+import datetime
 
 # Date Reading functions
-def read_zarr_data(zarr_path):
-    """Read and return data in zarr format.
+
+np.random.seed(123)
+
+def read_enacts(variable, ds_conf="dekadal", as_array=True):
```
aaron-kaplan, line 10: I think it would be simpler to have `ds_conf` be the particular variable's config rather than the set of configs. So it would be called like
```
var = 'precip'
read_enacts(var, GLOBAL_CONFIG['datasets']['daily']['var']
```

BTW `as_array=False` doesn't seem useful. If I want a Dataset I can just call .to_dataset().

remicousin: That's not going to work that straightforwadly as the reading currently needs the following configuration where your var is actually inside "vars" what would be just fine but is also needed zarr_path that is tuck under dekadal/daily... for no (any more) good reason it seems.

I guess part of the decoupling between data reading and app also concerns the config file so maybe we should just revisit it now. You were alluding to it earlier.

    dekadal:
        zarr_path: /data/remic/mydatafiles/
        vars:
            precip: 
                - ANACIM/anacim_v3/ALL_block2/rr_mrg_dekad/
                - null
                - precip
            tmax: 
                - ANACIM/anacim_v3/ALL_block2/tx_mrg_dekad/
                - null
                - temp
            tmin: 
                - ANACIM/anacim_v3/ALL_block2/tn_mrg_dekad/
                - null
                - temp

remicousin: well it's not as much the reading/app that are tangled into the config file as the app/zarrification

aaron-kaplan: > your var is actually inside "vars" what would be just fine but is also needed zarr_path that is tuck under dekadal/daily... for no (any more) good reason it seems.

I'm not sure what you mean about "no good reason." It seems appropriate that `zarr_path` isn't defined on each individual var, because the same value is shared by all of them. If you mean that it should be at the top level rather than inside `daily` and `dekadal`, I agreee with that.

Giving every function access to the entire configuration can make the code hard to maintain as it grows, because when you change some element of the configuration you have to find and change all the places it's used. In this case I might make `zarr_path` an extra argument rather than passing in the entire config.

I probably obsess about configuration mechanisms more than is useful. The application is still small and there's currently no big risk of it getting unmaintainable; and however we choose to structure it now, we can change our minds later.

enacts/calc.py
```diff
@@ -82,7 +80,7 @@ def read_enacts(variable, ds_conf="dekadal", as_array=True):
             data_path = ds_conf['vars'][variable][0]
         xrds = xr.open_zarr(f"{ds_conf['zarr_path']}{data_path}")
         array = ds_conf['vars'][variable][2]
```
aaron-kaplan, line 82: `array` is an unhelpful name here. I assumed it was a DataArray, but then saw it being used as `xrds[array]`, and had to go look at the config to confirm that it's actually the name of the variable in the zarr file. Something like `zarr_varname` might be better; alternatively, change the zarrification script to always put the application's standard names (e.g. `tmax` and `tmin` instead of `temp`) in the zarrs, so we don't even need to look at `var_config[2]` when reading.

aaron-kaplan (approved):

enacts/calc.py
remicousin: ok. How about last commit.
(config files is the first thing that is going to face 3rd party users/developers so they better be nice :) )

enacts/calc.py
```diff
@@ -7,17 +7,53 @@
...
+    --------
+    xr.open_zarr
+    """
+    try:
+        data_path = ds_conf[time_res]['vars'][variable][1]
+        if data_path is None:
+            data_path = ds_conf[time_res]['vars'][variable][0]
+        zarr_path = f"{ds_conf[time_res]['zarr_path']}{data_path}"
+        var_name = ds_conf[time_res]['vars'][variable][2]
+    except KeyError:
```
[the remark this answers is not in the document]
remicousin: how about now?

enacts/calc.py
```diff
@@ -1,27 +1,101 @@
...
-    zarr_data = xr.open_zarr(zarr_path)
-    return zarr_data
+    try:
+        data_path = ds_conf['vars'][variable][1]
+        if data_path is None:
+            data_path = ds_conf['vars'][variable][0]
+        zarr_path = f"{ds_conf['zarr_path']}{data_path}"
+        var_name = ds_conf['vars'][variable][2]
+    except KeyError as e:
+        raise Exception(f'configuration of data reading raised error {e}')
```
aaron-kaplan, line 36: What do you intend to accomplish by catching and reraising, as opposed to just letting the KeyError be raised?

aaron-kaplan (commented):
This is not better. If you misspell "daily" in the config file, you still get no indication that anything is wrong, it just silently substitutes fake data for the real data. Depending on how realistic the fake data are, that could be both hard to catch and potentially damaging.

Plus the logic for deciding between real and fake data is now duplicated in every maproom instead of centralized in one place.

Here's a suggestion that's simpler than the thing I mentioned in my previous review that was complicated to explain. Make `daily` mandatory in the config file. If it's set to `FAKE` then use fake data, otherwise interpret it as a zarr file path. (At first I was going to suggest `null` instead of `FAKE` as the special value, but I think we might want to support sites that have only daily and no dekadal, or vice versa.)

enacts/calc.py
```diff
@@ -1,28 +1,126 @@
...
+    data_path = dst_conf['vars'][variable][1]
+    if data_path is None:
+        data_path = dst_conf['vars'][variable][0]
+    zarr_path = f"{dst_conf['zarr_path']}{data_path}"
+    var_name = dst_conf['vars'][variable][2]
+    xrds = xr.open_zarr(zarr_path)
+    return xrds[var_name]    
 
 
+def synthetize_enacts(variable, time_res):
```
aaron-kaplan, line 65: It's spelled (and pronounced) "synthesize," with an s. ("Synthetic" with a t is correct, however.)

aaron-kaplan (approved):
That's better.

PR #523, opened 2025-03-21
shading years or ref and proj periods

remicousin:
Ultimately, would want to allow the user to move the edges of the rectangle to update those periods. But one step at a time...

### DIFF
--- a/pepsico/projections/maproom.py
+++ b/pepsico/projections/maproom.py
@@ -7,6 +7,7 @@ from . import layout
 import plotly.graph_objects as pgo
 import xarray as xr
 import pandas as pd
+from dateutil.relativedelta import *
 import dash_leaflet as dlf
 import psycopg2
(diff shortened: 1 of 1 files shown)


aaron-kaplan:
I did a superficial review and it looks fine. I don't fully understand the meaning of the graph; I'm assuming Xandre has more context, so I won't dig into it unless you ask me to.

remicousin:
Adding some explanations... In this app, the map is seasonal anomaly between years of projected variable against years of historical reference years; while the ts is the entire period seasonal time series. The PR adds a visual to the ts to show what years are part of the reference period and what years part of the projected period making the anomalies map.

(and ultimately I'd like this visual helper to be also interactive so that the user could manipulate it to change years of respective periods as an alternative to change the corresponding text box -- but not part of this PR).

aaron-kaplan (approved):

PR #527, opened 2025-03-28
daily to dekadal functions

remicousin:
Functions to compute dekadal totals from daily data and their tests. Initially needed to generate synthetic dekadal data from synthetic daily data to test Maprooms.

In the process, I (eventually!) discovered how to make use of the Intervals objects that groupby functions return. It turns out they are pandas.Interval, but you need surgical striking to hit the pandas object with its attributes: too soon, it's still a DataArray and too late it's an ndarray... tells so much about my understanding of how xarray, numpy, pandas... intricate...

So there are 3 functions:

- groupedby_dekads derived from xr.groupby_bins returns a grouped object to which can be applied group methods (sum, mean, etc)
- swap_interval assigns a new coord at the left, mid or right point of the interval dim that a grouped object returns after being applied a method
- daily2dekad_sum applies all the above to return dekadal totals with an additional dim at the beginning of the dekad (what we used in zarrification scripts

For the very short-term purpose of synthetic data into Maprooms, I will need a mean version for temperature. It's trivial.
I didn't deal with units since we might use pint some day.

For the medium term, swap_interval can be applied to any sort of dim (not just time). There could be simplification/generalization to make after checking what xarray ojects return Interval dims and whether they name it always the same or what-have-you. Right now I assumed some behavior.

For longer term, that could open the door for clean library to deal with time: the Intervals contain all the information we need and can be retrieved. I imagine pair-wise operations could be applied on the Interval dims directly, what would force the Intervals to be consistent throughout the pair (Ingrid only relies on centers). Time-wise operations could probably be automatized to start from Interval, move to point-wise for the sake of the operation, and returned to appropriate Interval accordingly. But we have other discussion on time elsewhere and maybe this is not new, or needs be completed by what was said there.


remicousin:
I mean this PR is really just to have acceptable dekadal data to feed the synthetic Maprooms. We can move the broader time dimension where it belongs.

enacts/tests/test_calc.py
```diff
@@ -5,6 +5,65 @@
...
+    precip = xr.DataArray(values, dims=["T"], coords={"T": t})
+    dekad_precip = calc.groupby_dekads(precip).sum()
+    dekad_precip_topoint = calc.swap_interval(
+        dekad_precip, "T_bins", to_point="left"
+    )
+
+    assert (dekad_precip_topoint["T"] == pd.to_datetime([
+        "2000-05-01", "2000-05-11", "2000-05-21",
+        "2000-06-01", "2000-06-11", "2000-06-21"
+    ])).all()
```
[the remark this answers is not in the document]
aaron-kaplan: You can resolve it by using backslashes instead of parentheses to break it into multiple lines. But in this case it might be more convenient to use `pd.testing.assert_array_equal` intead of  `assert`.

enacts/calc.py
```diff
@@ -594,6 +594,141 @@ def _cess_date(dry_thresh, dry_spell_length_thresh, sm_func, time_coord):
...
+    if (daily_data[time_dim][-1] + np.timedelta64(1, "D")).dt.day == 1 :
+        dekad_bins = xr.concat(
+            [dekad_bins, daily_data[time_dim][-1] + np.timedelta64(1, "D")],
+            dim=time_dim,
+        )
+    assert dekad_bins.size > 1, "daily_data must span at least one full dekad"
+    return daily_data.groupby_bins(daily_data[time_dim], dekad_bins, right=False)
+
+
+def daily2dekad_sum(daily_data, time_dim="T"):
```
aaron-kaplan, line 699: The `getattr` function I mentioned elsewhere could also be used here to make the aggregation (mean or sum) a parameter.

aaron-kaplan (commented):
The Interval class is interesting. But if you have to convert it to points in order to do anything useful with it, then it's of limited utility.

enacts/calc.py
```diff
@@ -594,6 +594,141 @@ def _cess_date(dry_thresh, dry_spell_length_thresh, sm_func, time_coord):
...
+    -------
+    grouped : xr.DataArrayGroupBy or xr.DatasetGroupBy
+        `daily_data` grouped by dekads
+
+    See Also
+    --------
+    xarray.groupby_bins
+    """
+    dekad_bins = daily_data[time_dim].where(
+        lambda x : ((x.dt.day == 1) | (x.dt.day == 11) | (x.dt.day == 21)),
```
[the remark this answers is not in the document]

enacts/calc.py
```diff
@@ -594,6 +594,141 @@ def _cess_date(dry_thresh, dry_spell_length_thresh, sm_func, time_coord):
...
+        drop=True,
+    )
+    if (daily_data[time_dim][-1] + np.timedelta64(1, "D")).dt.day == 1 :
+        dekad_bins = xr.concat(
+            [dekad_bins, daily_data[time_dim][-1] + np.timedelta64(1, "D")],
+            dim=time_dim,
+        )
```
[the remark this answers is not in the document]
remicousin: was not needed, or resolved by the sub-day precision discrepancy mentioned above.

enacts/calc.py
```diff
@@ -594,6 +594,141 @@ def _cess_date(dry_thresh, dry_spell_length_thresh, sm_func, time_coord):
...
+    assert dekad_bins.size > 1, "daily_data must span at least one full dekad"
```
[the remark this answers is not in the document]
remicousin: I don't think it's related to the noon vs midnight issue and I don't think we can get away from it. If time_dim is 2 Jan 2000 to 19 Jan 2000, you can not form 1 dekad. You need 2 edges to form 1 bin. So I renamed objects between bins and edged for clarity. groupby_bins' bins parameter is either a number of bins or edges so it's difficult to be consistent both with meaning and choices made from the function this is all derived from

enacts/tests/test_calc.py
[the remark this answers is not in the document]
remicousin: It was an assert in one of the function that was causing this, not the tests' asserts. It's now corrected (or maybe removed as no more necessary).
I guess I didn't notice as I was running tests function by function and may have not noticed it when I ran the tests that triggered it.

enacts/calc.py
```diff
@@ -594,6 +594,141 @@ def _cess_date(dry_thresh, dry_spell_length_thresh, sm_func, time_coord):
 
 
 # Time functions
+
+
+def swap_interval(data, interval_dim, to_point="mid", assigned_coord=None):
+    """ Assign coordinate of left, mid or right point of interval dimension
+    and swap to it
```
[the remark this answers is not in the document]
remicousin: Function that builds the new coordinate has split and namings revisited. I am trying to name after the xr functions used to try and remain in one same nomenclature. But that may not be advisable.

remicousin:
> The Interval class is interesting. But if you have to convert it to points in order to do anything useful with it, then it's of limited utility.

All the commits but last address the comments of this PR. The last one adds a suite of functions that can replace all others, including the older seasonal functions for which cases confirming tests have been made too.

You'll have to make a concession somewhere. With Intervals, you'll have to convert them to something to make maths. Without, you need at least 2 coords to describe one time dimension...

enacts/tests/test_calc.py
```diff
@@ -5,6 +5,298 @@
...
+    grouped = calc.groupby_dekads(precip).sum()
+    mid_dim = calc.interval_to_point(grouped["T_bins"])
+    left_dim = calc.interval_to_point(grouped["T_bins"], to_point="left")
+    right_dim = calc.interval_to_point(grouped["T_bins"], to_point="right")
+
+    assert mid_dim.name.endswith("_mid")
+    assert left_dim.name.endswith("_left")
+    assert right_dim.name.endswith("_right")
+    for t in range(mid_dim.size) :
+        assert mid_dim.values[t] == mid_dim["T_bins"].values[t].mid
```
aaron-kaplan, line 57: Could you explain? I thought `T_bins` was an array of intervals, but you're comparing them to points here? And why aren't you making any assertions about left_dim and right_dim? (Also why are they all called `*_dim`, when they're not dimensions but DataArrays?)

enacts/calc.py
```diff
@@ -596,6 +596,51 @@ def _cess_date(dry_thresh, dry_spell_length_thresh, sm_func, time_coord):
...
+    Notes
+    -----
+    Should work for any type of array, not just time.
+    xr.groupby_bins against dim renames the Interval dim_bins,
+    not sure if xr.groupby does the same,
+    and what other Xarray functions return Intervals but, depending,
+    could generalize the returned array name
+    """
+    return xr.DataArray(
+        data = [getattr(interval.values[t], to_point) for t in range(interval.size)],
```
aaron-kaplan, line 632: PEP8 says no spaces around the equals sign when it's used in a kwarg.

(PEP8 is the style guide for contributions to python itself, but is also generally used for python programs in general.)

enacts/calc.py
```diff
@@ -718,7 +658,7 @@ def groupby_dekads(daily_data, time_dim="T"):
 
     See Also
     --------
-    xarray.groupby_bins
+    xarray.groupby_bins, xarray.DataArrayGroupBy, xarray.DatasetGroupBy
```
aaron-kaplan, line 699: `DataArrayGroupBy` and `DatasetGroupBy` aren't in the `xarray` package, they're in `xarray.core.groupby`.

enacts/calc.py
```diff
@@ -594,6 +594,423 @@ def _cess_date(dry_thresh, dry_spell_length_thresh, sm_func, time_coord):
...
+        a dictionary of keyword arguments to pass to `method`
+        Default is an empty dict thus applying method default arguments
+    time_dim : str, optional
+        name of daily time dimenstion, default is "T"
+
+    Returns
+    -------
+    dekad_data : xr.DataArray or xr.Dataset or DataArrayGroupBy or DatasetGroupBy objects
+        daily data grouped by dekad if no `method` provided. Otherwise daily data
+        reduced to dekadal data according to `method`
```
aaron-kaplan, line 737: This function seems complicated and unnecessary. It's no harder to write
```
groupby_dekads(daily_data).sum()
```
than to write
```
daily_to_dekad(daily_data, method="sum")
```

enacts/calc.py
```diff
@@ -746,6 +746,271 @@ def daily_to_dekad(daily_data, method=None, method_kwargs={}, time_dim="T"):
...
 
 
+def resample_interval_to_daily(time_series, extent=None, time_dim="T_bins"):
+    """ Resample any (interval-based) time series to daily
+
+    Parameters
+    ----------
+    time_series : xr.DataArray or xr.Dataset
+        data depending on time intervals greater or equal then a day
+    extent : str, optional
```
aaron-kaplan, line 756: Make this a boolean instead of a string, and call it `is_extensive` (or `is_intensive`, your choice).

Also, maybe it's just me but I would be helped by definitions or examples (temperature, rainfall). I can never remember which is which.

enacts/calc.py
```diff
@@ -746,6 +746,271 @@ def daily_to_dekad(daily_data, method=None, method_kwargs={}, time_dim="T"):
...
+    pandas.Interval, interval_to_point, swap_interval_to_assigned_point,
+    xarray.resample
+
+    Notes
+    -----
+    The day is considered the smallest unit or interval of time in the sense that a
+    time dimensions expressed as time points is considered equivalent to intervals of
+    length 1 day. There may be generalization to make to adapt to the actual smallest
+    time unit of this ecosystem which is the ns.
+    Extent proterty could be inferred in more cases, e.g. Kelvin is intensive.
```
aaron-kaplan, line 781: "property" misspelled

enacts/calc.py
```diff
@@ -746,6 +746,271 @@ def daily_to_dekad(daily_data, method=None, method_kwargs={}, time_dim="T"):
...
+
+    Returns
+    -------
+    time_series : xr.DataArray or xr.Dataset
+        `time_series` resampled to daily
+
+    See Also
+    --------
+    pandas.Interval, interval_to_point, swap_interval_to_assigned_point,
+    xarray.resample
```
aaron-kaplan, line 773: `xarray.resample` doesn't exist. You mean `xarray.DataArray.resample` and `xarray.Dataset.resample`? BTW how does this differ from those? Might be good to add the answer to the documentation.

enacts/calc.py
```diff
@@ -746,6 +746,271 @@ def daily_to_dekad(daily_data, method=None, method_kwargs={}, time_dim="T"):
...
+
+    Parameters
+    ----------
+    time_series : xr.DataArray or xr.Dataset
+        data depending on time intervals greater or equal then a day
+    extent : str, optional
+        indicate the "extensive" or "intensive" property of `time_series` .
+        Upsampling to daily requires intensive data.
+        If extensive, make intensive by dividing by length of intervals in days
+        Default is None in which case: if units end with "/day", considers intensive,
```
aaron-kaplan, line 760: Why would we use this function if the input was already in units of */day?

aaron-kaplan (commented):
I've reviewed all but the last commit ("another suite of functions..."). Will have to continue with that next week.

enacts/calc.py
```diff
@@ -746,6 +746,271 @@ def daily_to_dekad(daily_data, method=None, method_kwargs={}, time_dim="T"):
...
+        ], dim=time_dim_left)
+        time_series = (
+            time_series.resample({time_dim_left: "1D"}).ffill()
+            # once filled, can drop the open right point of last interval
+            .isel({time_dim_left : slice(0, -1)})
+        )
+    return time_series
+
+
+def regroup(time_series, group="1D", method=None, method_kwargs={}, time_dim="T"):
```
aaron-kaplan, line 827: Trying to understand why you chose the name "regroup," and how that differs from resampling. It seems like grouping, not regrouping. I think what you had in mind was that the input consists of intervals, which are sort of like groups; but I find that confusing.

enacts/calc.py
```diff
@@ -746,6 +746,271 @@ def daily_to_dekad(daily_data, method=None, method_kwargs={}, time_dim="T"):
...
+        elif group == "8day" :
+            # last period of year used to adjusting
+            bins = edges_base.where((edges_base.dayofyear % 8) == 1).dropna()
+        elif group == "dekad" :
+            bins = edges_base.where(
+                (edges_base.day == 1)
+                | (edges_base.day == 11)
+                | (edges_base.day == 21)
+            ).dropna()
+        elif group == "16day" :
```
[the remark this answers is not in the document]
aaron-kaplan: Now I see there's a test that uses 7D as the resampling period. Couldn't 8day and 16day be handled the same way?

enacts/tests/test_calc.py
```diff
@@ -5,6 +5,33 @@
 import data_test_calc
 
 
+def test_groupby_dekads_perfect_partition():
+    t = pd.date_range(start="2000-05-01", end="2000-06-30", freq="1D")
+    values = 1 + np.arange(t.size)
+    precip = xr.DataArray(values, coords={"T": t})
+    grouped = calc.groupby_dekads(precip)
+
+    assert (grouped.sum() == [ 55, 155, 286, 365, 465, 565 ]).all()
```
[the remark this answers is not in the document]
remicousin: done throughout test_calc, not just the new tests.

enacts/calc.py
```diff
@@ -594,6 +594,423 @@ def _cess_date(dry_thresh, dry_spell_length_thresh, sm_func, time_coord):
...
+            interval.name.replace("_bins", f'_{to_point}')
+            if interval.name.endswith("_bins")
+            else "_".join(interval.name, f'_{to_point}')
+        ),
+        attrs = interval.attrs if keep_attrs else {},
+    )
+    return data
+
+
+def swap_interval_to_assigned_point(
```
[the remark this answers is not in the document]
remicousin: Right... that's why I also put assign in the name. Maybe it should have been assign_and_swap, but then it's harder to concisely include interval and point in the name... anywho... done.

enacts/calc.py
remicousin: exit daily_to_dekad

enacts/calc.py
remicousin: see extended Notes: e.g. monthly precipitation can be expressed in mm/day -- and you'd still want to resample that to daily

enacts/tests/test_calc.py
remicousin: T_bins is an array of intervals. .mid is an attribute of a pd.Interval but only each value of T_bins is of type pd.Interval to which .mid can be applied. So both mid_dim and T_bins... .mid are points and I am testing that my function does the right point conversion. I could do the same test for left and right... I guess I figured if it works for one attribute...

I guess I called them _dim by usage extension... since all of this is to convert coordinates (or dims... the distinction between coords and dims in xarray API/doc is sometimes confusing too). So... lots of confusion.

remicousin:
The above address all but "another suite..." commit

enacts/calc.py
remicousin: It's **regroup** because it's a variation of **group**by_bins dedicated to time groups we care about as **re**sample is a groupby dedicated to time objects :)

"It seems like grouping, not regrouping." My English is not rich enough to grasp the subtleties between those 2 (as much as I don't really see a difference between sampling and resampling -- other than some emphasis effect).

xarray's resample only takes "frequencies" as input. Even if there are many choices, there are none to make those we want (dekads, pentads, 8-day, 16-day -- and who knows... something new could come up in the future). Another option could have been to add a frequency string to resample, which seems to mean to add to pandas (or does it? I don't know if Xarray are reimplementations or wrappers... maybe it depends...). Either way I've not added to either xr or pd so far.
There is not resampler_bins like for groupby (which ironically is of interest because _bins allows edges as inputs to define said bins). Well resampler [doc](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.resample.html#id2) claims it can take a Resampler object for specialized resampling liek groupby but the [source](https://docs.xarray.dev/en/stable/generated/xarray.groupers.Resampler.html#xarray.groupers.Resampler) code of that is eloquent (maybe they are waiting for my contribution here?).

Another difference is that resample returns points, not intervals. Even though it offers to reduce against the new sampling, doing the reduction causes to then ignore those intervals by "anchoring" somewhere (at day level it seems), often the beginning or end through indication in the frequency string, and loose information. The commit is part of an interval-based implementation. The alternative is probably a pair of coords with the start and end of the time intervals.

enacts/calc.py
```diff
@@ -746,6 +746,271 @@ def daily_to_dekad(daily_data, method=None, method_kwargs={}, time_dim="T"):
...
+            f"group must be int, array, or str of form nD, pentad, 8day, dekad,"
+            f" 16day, nM,d-d Mmm or d Mmm - d Mmm"
+        )
+    if (not isinstance(group, int)):
+        assert (bins.size > 1), (
+            "data must span at least one full group (need 2 edges to form 1 bin)"
+        )
+    regrouped = time_series.groupby_bins(time_series[time_dim], bins, right=False)
+    if method is not None :
+        regrouped = getattr(regrouped, method)(**method_kwargs)
```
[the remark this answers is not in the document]
remicousin: done. Could potentially break down further and remove the groupby_bins part and really have a function making bins (or edges) of interest to feed into groupby_bins. While the purely bins definitions could be factored out, the beginning of regroup deals with converting any interval to daily, which I am not sure is a definitive solution in that scheme (see Notes).

enacts/calc.py
```diff
@@ -746,6 +746,271 @@ def daily_to_dekad(daily_data, method=None, method_kwargs={}, time_dim="T"):
...
+            bins = edges_base.where((edges_base.dayofyear % 16) == 1).dropna()
+        elif group.endswith("M") :
+            bins = edges_base.where(edges_base.day == 1).dropna()
+            bins = np.array([bins[t] for t in range(0, bins.size, int(group[:-1]))])
+        elif "-" in group :
+            # e.g. "29 Feb - 30 Mar" or "2-29 Mar"
+            # This case usage is to keep only the season of interest given as input.
+            # Thus not to form a partition of time as in other cases and could be
+            # moved to another function. Or if kept could include the selection of
+            # said season of interest. (Or just leave as is).
```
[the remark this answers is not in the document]
remicousin: I'll leave it for now, until there is a sense we are going somewhere.

remicousin:
> I'm having trouble figuring out where you're going with this last commit. I'm a little sleepy, so maybe I'll have better luck another time; but also some high-level guidance might help.
> 
> Is this meant to be a piece of the general-purpose time library you mentioned in the last paragraph of the PR description? If so, then before reviewing a detailed implementation I would like to see a high-level summary of the proposed design. What class of problem are you trying to solve, how do resampling and regrouping contribute to solving those problems, and then what is the resampling and regrouping API you propose. Perhaps do some example calculations, just on paper, using the API you wish you had.

Yes, it's an experiment for a general-purpose time library. I still want to finish clean up the existing Python code  we have (splitting data reading / apps etc.) before I engage in that (or actually I could work on both in parallels if I don't dive into servers logs analyses). But I got carried away with the idea and stuck that in here. So maybe see it as an experiment. I can still write high level considerations in [here](https://docs.google.com/document/d/1pRB3gGiDC6AWGiQBrGDG1PB5GkgYOQnl3c2c2rQ7EHU/edit?tab=t.0).

enacts/tests/test_calc.py
aaron-kaplan: Indeed, I seem to have missed the `.mid` at the end of the line.

In xarray, a coordinate is a DataArray that contains the labels for a dimension. A dimension isn't represented by any xarray class, it's just a description of how the data in an array are shaped.

enacts/calc.py
aaron-kaplan: You resolved this comment without addressing it?

enacts/calc.py
aaron-kaplan: Sampling maps a continuous signal to a sequence of discrete values. Resampling maps a sequence of discrete values to another sequence of discrete values.  It's called resampling because the input sequence is typically the result of some earlier sampling step.

Similarly, grouping takes samples as input and yields groups of samples as output. Regrouping means taking groups as input and yielding different groups as output.

> resample is a groupby dedicated to time objects

No it isn't. Groupby is *part* of the implementation of resampling (particularly downsampling). The other part is aggregation (sum or mean) over the groups to yield a new sequence of samples.

> Another difference is that resample returns points, not intervals.

That's only an observation about xarray's current implementation, not about resampling in the abstract. I think what you're proposing would best be implemented as changes to xarray to make it able to represent time coordinates using intervals instead of points.

aaron-kaplan:
I would prefer to have one PR with a minimal implementation of just what we need now to make synthetic data for testing, and a different branch where we experiment with generalizations. The short-term minimal stuff might belong in a testutils.py rather than calc.py so we don't build too much on top of it before we decide it's the right approach.

enacts/calc.py
remicousin: boh...? I though I had...

enacts/calc.py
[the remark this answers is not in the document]
remicousin: doesn't belong in this PR anymore

enacts/calc.py
remicousin: moved to another branch

remicousin:
Move the experimental "regroup" with pd.Interval to another branch.

xchourio (approved):

PR #528, opened 2025-03-28
Splitshpandmapping

remicousin:
This is a bunch of factorization for us to get ready to split shapes reading from apps and apply the case of fake shapes. But I am not sure at what level we'll fork the fake case... at sql/db? at GeoJSON? create dummy geom from scratch?

The commits kind of make sense in chronological order except I forgot to add the new file which should have been in the 2nd commit, and later commits update the former so... might make more sense to look at the final thing. Commits' comments tell you the story.

### DIFF
--- a/enacts/calc.py
+++ b/enacts/calc.py
@@ -144,8 +144,7 @@ def sql2GeoJSON(shapes_sql, db_sql):
 
     See Also
     --------
-    psycopg2.connect, psycopg2.sql, pandas.read_sql, shapely.wkb,
-    shapely.MultiPolygon, shapely.geometry.mapping
+    sql2geom, geom2GeoJSON
 
     Examples
     --------
@@ -160,6 +159,73 @@ def sql2GeoJSON(shapes_sql, db_sql):
     Notes
     -----
     Description of Return is approximative as I don't know exactly what the output is
+    """
+    return geom2GeoJSON(sql2geom(shapes_sql, db_sql))
+
+
+def geom2GeoJSON(df):
+    """ Form a GeoJSON dict from a geometric object
+
+    Parameters
+    ----------
+    df: geometric object
+        shapely geometric object
+    
+    Returns
+    -------
+    features: dict
+        dictionary with features as key and GeoJSON of `geom` as value
+
+    See Also
+    --------
+    sql2geom, shapely.MultiPolygon, shapely.geometry.mapping
+
+    Notes
+    -----
+    Description of Return is approximative as I don't know exactly what the output is
+    """
+    df["the_geom"] = df["the_geom"].apply(
+        lambda x: x if isinstance(x, MultiPolygon) else MultiPolygon([x])
+    )
+    shapes = df["the_geom"].apply(shapely.geometry.mapping)
+    for i in df.index: #this adds the district layer as a label in the dict
+        shapes[i]['label'] = df['label'][i]
+    return {"features": shapes}
+
+
+def sql2geom(shapes_sql, db_sql):
+    """ Form a geometric object from sql query to a database
+
+    Parameters
+    ----------
+    shapes_sql: str
+        sql query
+    db_sql: dict
+        dictionary with host, port, user and dbname information
+    
+    Returns
+    -------
+    geometric object
+        shapely geometric object
+
+    See Also
+    --------
+    psycopg2.connect, psycopg2.sql, pandas.read_sql, shapely.wkb,
+
+    Examples
+    --------
+    shapes_sql: select id_1 as key, name_1 as label,
+        ST_AsBinary(the_geom) as the_geom from sen_adm1
+    db_sql:
+        host: postgres
+        port: 5432
+        user: ingrid
+        dbname: iridb
+
+    Notes
+    -----
+    Description of Returns is approximative
+    as I don't know exactly what the output is
     """
     with psycopg2.connect(**db_sql) as conn:
         s = sql.Composed(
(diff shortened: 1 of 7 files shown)


xchourio (approved):

enacts/calc.py
```diff
@@ -121,6 +126,126 @@ def synthesize_enacts(variable, time_res):
...
+
+def sql2GeoJSON(shapes_sql, db_sql):
+    """ Form a GeoJSON dict from sql request to a database
+
+    Parameters
+    ----------
+    shapes_sql: str
+        sql request
+    db_sql: dict
+        dictionary with host, port, user and dbname information
```
aaron-kaplan, line 138: This is not SQL, so I don't understand why the variable name is `db_sql`. `db_config`?

enacts/maproom_utilities.py
```diff
@@ -0,0 +1,50 @@
+import dash_leaflet as dlf
```
aaron-kaplan, line 1: File name: mapping_utilities.py? Everything in this repo is for maprooms, e.g. calc.py and pingrid.py.

enacts/flex_fcst/maproom.py
```diff
@@ -42,52 +40,6 @@ def register(FLASK, config):
...
-                },
-            ),
-            name=adm_name,
-            checked=is_checked,
-        )
-
-
     #Should I move this function into the predictions.py file where I put the other funcs?
     #if we do so maybe I should redo the func to be more flexible since it is hard coded to read each file separately..
     def read_cptdataset(lead_time, start_date, y_transform=config["y_transform"]):
```
aaron-kaplan, line 45: Just noticing that the logic for reading CPT datasets is written here, not imported from cptio. Not something to chnage in this PR, but potentially something to add to the todo list.

enacts/calc.py
```diff
@@ -121,6 +126,126 @@ def synthesize_enacts(variable, time_res):
...
+    ----------
+    shapes_sql: str
+        sql query
+    db_sql: dict
+        dictionary with host, port, user and dbname information
+    
+    Returns
+    -------
+    geometric object
+        shapely geometric object
```
aaron-kaplan, line 209: No it isn't, it's a pandas dataframe (hence the name `df`), one column of which contains shapely objects.

Incidentally, it's a plain old pandas dataframe, not a [geopandas](https://geopandas.org/en/stable/) dataframe. I remember suggesting to Drew that geopandas might be more appropriate, and that we eventually decided not to switch, but I don't have a clear memory of  the reason. It might have been that it just didn't seem like it helped enough to be worth adding another dependency to the conda environment.

enacts/calc.py
```diff
@@ -121,6 +126,126 @@ def synthesize_enacts(variable, time_res):
...
+    db_sql:
+        host: postgres
+        port: 5432
+        user: ingrid
+        dbname: iridb
+
+    Notes
+    -----
+    Description of Returns is approximative
+    as I don't know exactly what the output is
```
aaron-kaplan, line 228: See above.

enacts/calc.py
```diff
@@ -121,6 +126,126 @@ def synthesize_enacts(variable, time_res):
...
+        shapely geometric object
+
+    See Also
+    --------
+    psycopg2.connect, psycopg2.sql, pandas.read_sql, shapely.wkb,
+
+    Examples
+    --------
+    shapes_sql: select id_1 as key, name_1 as label,
+        ST_AsBinary(the_geom) as the_geom from sen_adm1
```
aaron-kaplan, line 218: I'm not sure why we use `ST_AsBinary` and then use shapely to convert to GeoJSON, instead of just using [`ST_AsGeoJson`](https://postgis.net/docs/ST_AsGeoJSON.html). Possibly because some of it was copy-pasted from fbfmaproom, where shapes are rasterized into the tiles, not sent to the browser as GeoJSON. Something for me to look into.

enacts/calc.py
```diff
@@ -121,6 +126,126 @@ def synthesize_enacts(variable, time_res):
     var_name = variable
     return xrds[var_name]
 
+
+def sql2GeoJSON(shapes_sql, db_sql):
```
aaron-kaplan, line 130: Do you have a justification for putting this shape stuff in calc rather than in maproom_utilities?

(Also the data reading stuff arguably doesn't belong in calc either.)

enacts/calc.py
aaron-kaplan: Even in those cases we get the result in AsBinary format (don't we?), and could choose to use AsGeoJson instead.

I'm not asking for a change right now, in any case.

enacts/maproom_utilities.py
aaron-kaplan: I don't know if it will be useful outside of maprooms, but it doesn't cover all of maprooms, it's specific to the map-drawing part.

enacts/calc.py
```diff
@@ -121,6 +126,113 @@ def synthesize_enacts(variable, time_res):
...
+    ----------
+    shapes_sql: str
+        sql query
+    db_config: dict
+        dictionary with host, port, user and dbname information
+    
+    Returns
+    -------
+    df : pandas.DataFrame
+        a pd.DF including "the_geom" : a shapely geometric object
```
aaron-kaplan, line 201:
```
a pd.DataFrame with columns "label" (dtype=string), "key" (string or int depending on the table), and "the_geom" (shapely.Geometry)
```
FWIW the fact that `key` is sometimes a string and sometimes a number is a bad design that tends to lead to bugs. But that's not at issue right now.

enacts/maproom_utilities.py
remicousin: more will come later in maproom_utilities.py

PR #529, opened 2025-03-28
adding v4 of the jja forecast for Gautemala jja

dantesab9:
Hi Aaron, can we have this new version in the public sever? It is missing some years but the forecast can be calculated for 2025 so it is helpful to have it available.


aaron-kaplan (approved):

aaron-kaplan:
Done.

PR #506, opened 2025-03-07
Updated fbf-update-data.py and fbfmaproom-sample.yaml for Madagascar NDJ

nitinmagima:
Hi Aaron,

The forecast generated with the new link much faster. However I was using the bad years from DJF over here and for some reason that's not working, any idea why? I've attached the the error below for you to see.

@dantesab9  I've pushed the updated config file to Madagascar's repository if you need to discuss it next week with the MET department and Azhar

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/magima/python-maprooms/fbfmaproom/fbfmaproom.py", line 1160, in table_cb
    main_df, summary_df, thresholds = generate_tables(
  File "/home/magima/python-maprooms/fbfmaproom/fbfmaproom.py", line 390, in generate_tables
    basic_ds = fundamental_table_data(
  File "/home/magima/python-maprooms/fbfmaproom/fbfmaproom.py", line 550, in fundamental_table_data
    obs_ds = select_obs(country_key, obs_keys, target_month0)
  File "/home/magima/python-maprooms/fbfmaproom/fbfmaproom.py", line 522, in select_obs
    ds = ds.where(lambda x: (x["time"].dt.month == target_month) & (x["time"].dt.day == target_day), drop=True)
  File "/home/magima/.conda/envs/fbfmaproom2/lib/python3.9/site-packages/xarray/core/common.py", line 1103, in where
    cond = cond(self)
  File "/home/magima/python-maprooms/fbfmaproom/fbfmaproom.py", line 522, in <lambda>
    ds = ds.where(lambda x: (x["time"].dt.month == target_month) & (x["time"].dt.day == target_day), drop=True)
  File "/home/magima/.conda/envs/fbfmaproom2/lib/python3.9/site-packages/xarray/core/dataset.py", line 1439, in __getitem__
    return self._construct_dataarray(key)
  File "/home/magima/.conda/envs/fbfmaproom2/lib/python3.9/site-packages/xarray/core/dataset.py", line 1350, in _construct_dataarray
    _, name, variable = _get_virtual_variable(self._variables, name, self.dims)
  File "/home/magima/.conda/envs/fbfmaproom2/lib/python3.9/site-packages/xarray/core/dataset.py", line 186, in _get_virtual_variable
    raise KeyError(key)
KeyError: 'time'
Traceback (most recent call last):
  File "/home/magima/python-maprooms/fbfmaproom/fbfmaproom.py", line 1160, in table_cb
[...]

PR #530, opened 2025-04-03
Updating the bad year versions for the NDJ MAdagascar tool

dantesab9:
Hi Aaron, we added an updated bad year set for NDJ season (added to origin al data and then zarrified) and fixed the yaml file accordingly to show the NDJ specific sets (two versions now).

### DIFF
--- a/fbfmaproom/fbfmaproom-sample.yaml
+++ b/fbfmaproom/fbfmaproom-sample.yaml
@@ -420,7 +420,7 @@ countries:
                 predictors:
                     - prcp-v1
                     - enacts-precip-mon
-                predictand: bad-years-v3
+                predictand: bad-years-ndj-v2
             observations:
                 enacts-precip-mon:
                     label: ENACTS MON precip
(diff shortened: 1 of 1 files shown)


aaron-kaplan (approved):

aaron-kaplan (changes requested):
There are git conflicts, meaning the version to which you made these changes was not the latest version. I'll look over the conflicts and get back to you on how to resolve them.

aaron-kaplan:
Nitin's PR #506 added the Madagascar NDJ configuration. It consists of two commits (see https://github.com/iridl/python-maprooms/pull/506/commits). The current PR from Dante includes the first of those two commits but not the second. Nitin's second commit on that branch changed some of same lines that this PR changes, so the two conflict.

When you start to work on something, the first step should always be
```
git checkout master
git pull
git checkout my-new-branch
```
(replacing `my-new-branch` with something descriptive). That pulls the latest version from github into your local copy. If you don't do that, you end up editing an old version rather than the latest one.

Since this change is small, I think the easiest way to fix it is to close this PR and start over.

dantesab9:
Thank you Aaron, that makes sense and noted!

PR #536, opened 2025-04-11
Add zambia config

jpngfile:


fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -3771,3 +3771,55 @@ countries:
                 length: 3.0
                 issue_months: [5, 6, 7, 8]
                 start_year: 1983
+    zambia:
+        logo: Zambia_IRI_120x48.png
```
aaron-kaplan, line 3775: @nitinmagima do we have a logo file? If not, we'll use the generic IRI one.

fbfmaproom/fbfmaproom.py
```diff
@@ -47,7 +47,7 @@
 SERVER.register_error_handler(ClientSideError, pingrid.client_side_error)
 
 month_abbrev = ['jan', 'feb', 'mar', 'apr', 'may', 'jun', 'jul', 'aug', 'sep', 'oct', 'nov', 'dec']
-abbrev_to_month0 = dict((abbrev, month0) for month0, abbrev in enumerate(month_abbrev))
+abbrev_to_month0: Dict[str, int]  = dict((abbrev, month0) for month0, abbrev in enumerate(month_abbrev))
```
aaron-kaplan, line 50: @jpngfile if you're motivated to set up type checking in Github Actions, I would enthusiastically support that.

aaron-kaplan: (Before embarking on that, talk to me. Topics to discuss: which type checker, managing the pixi/conda environment.)

fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -3771,3 +3771,55 @@ countries:
...
+        center: [27, -15]
+        zoom: 6
+        rotation: 0
+        marker: [25, -17]
+        shapes:
+            - name: National
+              sql: select adm0_code as key, adm0_name as label, ST_AsBinary(the_geom) as the_geom
+                  from iridb.g2015_2012_0 where adm0_name = 'Zambia'
+              vuln_sql: select adm0_code as key, year, sum(vulnerability) as vuln
+                  from public.fbf_vulnerability where adm0_name = 'Zambia' group by 1, 2
```
aaron-kaplan, line 3786: This presumably doesn't work, because there are no rows for Zambia in the fbf_vulnerability table. That will cease to be a problem once the vulnerability layer is removed, but it would be good to merge this sooner than that. See `ethiopia` for a dummy query you can use instead.

aaron-kaplan (changes requested):
This currently doesn't work, probably because the forecast in /data/aaron/fbf-candidate/original-data/zambia is Riz' version, which uses the wrong season (NDJ). Please fix that, and send a pull request for the forecast configuration in python_maproom_zambia.

Please add a README to /data/aaron/fbf-candidate/original-data/zambia explaining what the different bad years csv files represent, where they came from, and what command was used to convert them to zarr. (@nitinmagima this probably requires input from you.)

fbfmaproom/fbfmaproom-sample.yaml
aaron-kaplan: To use the IRI one, put `IRI_128x128.png`

fbfmaproom/fbfmaproom-sample.yaml
aaron-kaplan: The vulnerability layer has been removed, so now all we need to do is remove `vuln_sql`.

PR #541, opened 2025-04-24
Splitcptread

remicousin:
This was a bit tedious because while pyCPTv2 data files can be known and read entirely from their parent path, the tsv case needs starts and leads/targets informed by the config files, read one pair of S L/target at a time, and can take L/target as one or the other.
This creates a chicken and egg issue with respect to reading data and initializing the Start and Target menus. Given that we don't expect new tsv files being produced, I tried to address this following 2 principles:

- make the least changes to the tsv case
- make the tsv case "transparent" meaning that all of it can be removed without altering anything else

There are 2 commits only because I had to change branches to work on something else so they are arbitrary, better to see the thing as a whole.

Functions Notes give some more details of the management of the 2 cases. I am also squeezing comments in the PR for clarification.

On top of the tsv/nc considerations, there are 2 cases: single or multiple targets. It matters both for data reading and app design. If multiple, the L dimension is introduced in the dataset and a target menu is displayed in the app. If single, there's no L dimention, and no target menu (only start).

Tested both single/multiple target cases but only for nc files. I don't have tsv files for Senegal. We could generate some or write a config file for tsv files we do have. This could be done now, or if/when ever tsv files cases arise in the future.

### DIFF
--- a/enacts/flex_fcst/maproom.py
+++ b/enacts/flex_fcst/maproom.py
@@ -13,7 +13,7 @@ from . import predictions
 from . import pycpt
 import maproom_utilities as mapr_u
 import dash_leaflet as dlf
-from globals_ import FLASK, GLOBAL_CONFIG
+from globals_ import GLOBAL_CONFIG
 import calc
 
 def register(FLASK, config):
@@ -302,7 +302,7 @@ def register(FLASK, config):
                 + "<br>%{x:.1f} "
                 + obs.attrs["units"],
                 name="obs (empirical)",
-                line=pgo.scatter.Line(color="blue"),
+                line=pgo.scatter.Line(color="forestgreen"),
             )
         )
         cdf_graph.update_traces(mode="lines", connectgaps=False)


enacts/config-test-enacts.yaml
```diff
@@ -76,6 +76,24 @@ datasets:
     # Soil data
     taw_file: /data/remic/mydatafiles/soilgrids/Senegal/GYGA_ERZD_wat_cap_abs_anacim_enacts3.nc
 
+    # Forecast data
+    fcst_data:
+        #forecast_path: /data/remic/mydatafiles/ANACIM/cpt_demo/UCSBV3
+        forecast_path: /data/remic/mydatafiles/ANACIM/cpt_demo/UCSBV3/Jul-Sep
```
remicousin, line 82: Use one or the above to test the nc multiple/single cases. Everything else below pertains to tsv case.

enacts/config-test-enacts.yaml
```diff
@@ -131,4 +149,6 @@ maprooms:
           min_wet_days: 60 #days
           wet_day_def: 1 #mm
 
-    flex_fcst: null
+    flex_fcst:
+      - y_transform: false
+        variable: Precipitation
```
remicousin, line 154: I also split within the config file data reading part and purely app part

enacts/flex_fcst/cpt.py
```diff
@@ -59,6 +319,60 @@ def read_file(
     return read_ds
 
 
+def read_cptdataset(lead_time, is_target, start_date, fcst_conf, y_transform=False):
```
remicousin, line 322: not new but taken away from maproom.py and adapted to take fcst_conf as input

enacts/flex_fcst/cpt.py
```diff
@@ -101,62 +416,153 @@ def starts_list(
...
+    If single target, `fcst_ds` remains 1D against S only and Ts remain coordiantes
+    of S.
+    `obs` is in any case always appended and sorted into a natural 1D time series
+    """
     data_path = Path(data_path)
     children = list(data_path.iterdir())
-
     if 'obs.nc' in map(lambda x: str(x.name), children):
-        fcst_mu, fcst_var, obs = read_pycptv2dataset_single_target(data_path)
+        fcst_ds, obs = read_pycptv2dataset_single_target(data_path)
```
remicousin, line 460: I had to change the rationale throughout from outputting tuples of xr.DataArrays to xr.Datasets of xr.DataArrays because:

1. so the outputs in the both nc/tsv cases be the same (tsv can also outputs hcst)
2. because I need to make ds-wise manipulations in the nc case where the Ts can either remain coordinates (single target) or become variables (multiple targets)

enacts/flex_fcst/cpt.py
```diff
@@ -101,62 +416,153 @@ def starts_list(
...
+        )
+        obs_slices = [obs_slices]
+        if len(children) > 1 :
+            for target in children[1:] :
+                new_fcst_ds, new_obs = read_pycptv2dataset_single_target(
+                    target, expand_L=expand_L
+                )
+                fcst_ds, new_fcst_ds = xr.align(fcst_ds, new_fcst_ds, join="outer")
+                fcst_ds = fcst_ds.fillna(new_fcst_ds)
+                obs_slices.append(new_obs)
```
remicousin, line 474: It turns out that we must never had run into the case of multiple lead times for multiple targets as the previous code was not covering that case. We only got the case of multiple targets with 1 same lead time covered... multiple targets and leads requite L to become a new dimension, not just another coordinate. Since we read and append the data target after target, it creates gappy S, L squares that must be either filled in or appended as we go. If we were reading and appending S after S, it would be more simple, but it would raise other problems (it's not how we organized the data files).

enacts/flex_fcst/maproom.py
```diff
@@ -439,21 +245,22 @@ def local_plots(marker_pos, start_date, lead_time):
         ).rename("obs_ppf")
         # Obs quantiles
         obs_quant = obs.quantile(quantiles, dim="T")
-
         # Forecast CDF
-        fcst_q, fcst_mu = xr.broadcast(quantiles, fcst_mu)
+        fcst_q, fcst_mu = xr.broadcast(quantiles, fcst_ds["deterministic"])
```
remicousin, line 249: until then, could work with the whole fcst_ds. From then, need to get the individual variables of interest

enacts/flex_fcst/maproom.py
```diff
@@ -657,7 +470,7 @@ def make_map(proba, variable, percentile, threshold, start_date, lead_time):
             mapr_u.make_adm_overlay(
                 adm_name=adm["name"],
                 adm_geojson=calc.sql2GeoJSON(adm["sql"], GLOBAL_CONFIG["db"]),
-                adm_clor=adm["color"],
+                adm_color=adm["color"],
```
remicousin, line 473: typo from other splitting PRs that applied to all Maprooms but could not test on fcst until now.

enacts/config-test-enacts.yaml
aaron-kaplan: Put that comment in the file.

aaron-kaplan (commented):
A comment before making a detailed review: do we need to continue supporting the tsv case? Can we not just say that if you want to upgrade your maprooms you have to also upgrade PyCPT?

remicousin:
> A comment before making a detailed review: do we need to continue supporting the tsv case? Can we not just say that if you want to upgrade your maprooms you have to also upgrade PyCPT?

I am very fine with that. In theory, there could be projects funding a new forecast Maprooms without updating the partner to upgraded PyCPT... In practice, they aren't going to be any (such) projects.

remicousin:
We might want to wrap this one in a way or another: we would need flex_fcst MR to port PepsiCo classic forecast MRs to that; and that would lay the ground for such MRs for IRI forecasts as well, if ever part of the shutdown mix.

aaron-kaplan:
Also note this branch now has conflicts.

remicousin:
> I vote to remove the tsv version. Ethiopia AA is now using the latest version of PyCPT. I don't know if there are other programs still using an older version, but if there are, they're not funded.

done

remicousin:
> Also note this branch now has conflicts.

rebased

enacts/flex_fcst/cpt.py
```diff
@@ -1,162 +1,228 @@
...
+                monthly_path,
+                'MME_forecast_prediction_error_variance_*.nc',
+                expand_L=expand_L,
+            ))
+    fcst_mu = xr.concat(mu_slices, "S")
+    fcst_var = xr.concat(var_slices, "S")
+    fcst_ds = fcst_mu.merge(fcst_var)
+    fcst_ds = fcst_ds.sortby(fcst_ds["S"])
     obs = xr.open_dataset(data_path / f"obs.nc")
     obs_name = list(obs.data_vars)[0]
```
aaron-kaplan, line 187: 186-187 can be written more succinctly and more readably as
```
obs = xr.open_dataarray(data_path / "obs.nc")
```

enacts/flex_fcst/cpt.py
```diff
@@ -1,162 +1,228 @@
...
     obs = obs[obs_name]
-    return fcst_mu, fcst_var, obs
+    return fcst_ds, obs
 
 
 def open_mfdataset_nodask(filenames):
     return xr.concat((xr.open_dataset(f) for f in filenames), 'T')
```
aaron-kaplan, line 193: Unused function?

enacts/flex_fcst/cpt.py
```diff
@@ -1,162 +1,228 @@
...
+    To use in read_pycptv2dataset expecting multiple targets, expand_L must be True
+    in which case T coordinates become variables and L is expanded to all variables.
+    If single target, expand_L muse be False and Ts remain coordinates of S only.
+    """
     filenames = path.glob(filepattern)
     slices = (xr.open_dataset(f) for f in filenames)
     ds = xr.concat(slices, 'T').swap_dims(T='S')
+    if expand_L :
+        L = (((
+            ds.isel(S=[0])["Ti"].dt.month - ds.isel(S=[0])["S"].dt.month
```
aaron-kaplan, line 225: Why `S=[0]` rather than `S=0`?

Unrelated question on the same line: do I understand correctly that you're defining L to be the interval between the start date and the beginning of the season, rather than between the start date and the midpoint of the season as in the DL? Seems like that might cause confusion for people (like us) who use both. But if it makes the maproom simpler I'm not necessarily opposed.

enacts/flex_fcst/cpt.py
```diff
@@ -1,162 +1,228 @@
...
+
+    See Also
+    --------
+    read_pycptv2dataset_single_target
+
+    Notes
+    -----
+    Whether `data_path` contains multiple targets or not is determined by whether the
+    observation obs.nc file(s) is a direct child of `data_path` or under target-wise
+    folders that have no restriction on how they are named.
```
aaron-kaplan, line 110: Seems like their names should be predetermined, not arbitrary, so we can just open the one we need instead of opening all of them and then picking the one we need (see my comment at the top of this review). Just a thought for another day.

enacts/flex_fcst/cpt.py
aaron-kaplan: Took me a minute to understand, but I think I see what you're saying. You get five years of May starts from one directory, and five years of June starts from another, and then you have to interleave them to put them on a common S grid. I think you reimplemented [`combine_by_coords`](https://docs.xarray.dev/en/stable/generated/xarray.combine_by_coords.html#xarray.combine_by_coords)?

I don't think the organization of the data files is necessarily a barrier to doing this differently. You could open all the files and then combine them all at once, or you could list them all and then read them in the order you want. But this seems fine (modulo using `combine_by_coords` if that's really what's happening).

It's probably more efficient to load them all and then align/fill/combine them all at once, instead of incrementally merging one at a time. If it's not slow then this is fine.

aaron-kaplan (commented):
I've reviewed cpt.py but not yet maproom.py. Will continue with that another time.

Your approach is to open all the files in the dataset every time, but I think the maproom probably only ever needs one S/L pair at a time, so you may be doing unnecessary work. If it's fast enough then it's fine. If it's slow, there's an optimization we can make.

enacts/flex_fcst/cpt.py
aaron-kaplan: > S=[0] rather than S=0 because read_pycptv2dataset_single_target applies append to the output of open_var and to do so, S need be an array, and not a scalar.

Assuming by "append" you mean `concat`, it works fine with scalar coordinates:
```
>>> da = xr.DataArray([1, 2], coords={'X': [100, 200]})
>>> da.isel(X=0)
<xarray.DataArray ()> Size: 8B
array(1)
Coordinates:
    X        int64 8B 100
>>> xr.concat([da.isel(X=0), da.isel(X=1)], dim='X')
<xarray.DataArray (X: 2)> Size: 16B
array([1, 2])
Coordinates:
  * X        (X) int64 16B 100 200
```

Your argument for integer lead time seems solid, so just make sure it's clearly documented. Maybe it would be better to use a different name, to remind us that it's not the same quantity as Ingrid's `L`. E.g. `L_start`?

(Every time I look at forecast date math I worry about eventually having to generalize it to the subseasonal case, but I think I can stop worrying about that now, since it seems like we're never going to get there.)

remicousin: open_var currently uses append, but I'll consolidate the discussion on whether to read all or one single pair of L/S down below.

Renames L Li to match the Ts nomenclatura. There are actually Ingrid datasets where L is the difference between start and first edge of forecast season. But anyway, that should be enough of an indication, and I think more important than documenting is to keep the information together with the target (Ts) which is done.

remicousin:
combine_by_coords doesn't work, or rather doesn't work in the current set up that nests the looping by targets and then S within those targets. This is because once you read the data for one targetA, you have say data for (S1, L=2) and (S2, L=1) so combine_by_coords makes a square array of S and L where the diagonal is NaN and the anti-diagonal is the data. When comes targetB, say it has data for (S2, L=2) but that cell already exists and there is no way to tell combine_by_coords to picj data over missing.

Now, as you pointed out, the figures, be they map or graphs, only use one (S, L) pair at a time so there is no need to combine all the data together, to then select the S and L of interest. However, there is a need to get all the S to initialize the corresponding menu, and then for a given S, the generate all the (L, T) to make the other target menu.

So could instead write something that gets all the S, and then all it's (L, T), and then write something else than gets data for a given (S, L)

remicousin:
> `Li` is perfect.
> 
> I was suggesting that we load all the slices for all the targets and then combine_by_coords all at once, rather than adding them up incrementally. Whatever.

something like the last commit?

aaron-kaplan:
Something like that, yes. But see comments in-line.

enacts/flex_fcst/pycpt.py
```diff
@@ -0,0 +1,225 @@
...
+            children[0], expand_L=expand_L
+        )
+        obs_slices = [obs_slices]
+        if len(children) > 1 :
+            for target in children[1:] :
+                new_mu_slices, new_var_slices, new_obs = read_pycptv2dataset_single_target(
+                    target, expand_L=expand_L
+                )
+                for nms in new_mu_slices:
+                    mu_slices.append(nms)
```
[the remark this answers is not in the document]
remicousin: before the loop, I was getting nested lists with append, which in turn made the combination fail. I don't know why += does it differently

PR #542, opened 2025-04-26
Replace apache with gunicorn

aaron-kaplan:
gunicorn is way simpler to install compared to mod_wsgi, which has to be compiled from source for the particular combination of python version and apache version you're using. This makes the container build fast enough that we can consider doing it in CI.

This is the first application I'm doing this with. If this works out, I'll gradually switch all our python images over.

Also moving from centos 7 to 9 in the process.

I probably won't deploy this right away, as I don't want to complicate debugging of the network issues we've been having.


aaron-kaplan:
@jturmelle any thoughts on this? Bad experiences with gunicorn, etc.?

I did some research into a few alternatives, can share that if you care.

jturmelle (commented):
I don't have anything against gunicorn. I've used it in the past for development, but never wanted to take the time to change our mod_wsgi configurations which were working well.  Do we need to make any changes to haproxy to accommodate this? https://docs.gunicorn.org/en/stable/deploy.html

aaron-kaplan:
We still have squid between haproxy and the application, so I can't think of a reason any changes to haproxy would be needed.

I did confirm that haproxy protects against "slow loris" attacks, which is the main reason they give for putting gunicorn behind nginx. (Squid probably does too, I just didn't find it documented.)

jturmelle (approved):
OK, I have no concerns about using it if you're comfortable with it.

PR #544, opened 2025-04-29
Intervals exp

remicousin:
Also contains daily2dekadal branch. Will rebase on master once daily2dekadal has merged so that it is easier to read.

This is an experiment using pd.Interval and groupby_bins to transform time series described by pd.Interval to explore ideas on how to improve time representation in Xarray and consort.


remicousin:
daily2dekadal has been merged into master and this has been rebased on master so that it now contains only this generalization/concept/experiment

PR #546, opened 2025-05-01
New forecast for Guatemala ASO and bad year update for Madagascar NDJ (old)

dantesab9:
Hi, I added a new version (v3) of the forecast for gutemala-aso but also updated bad years for madagascar-ndj. Sorry to mix topics in this branch, I will start from the master branch next time. I realize the Madagascar update is older and may be in conflict at this point.

### DIFF
--- a/fbfmaproom/fbfmaproom-sample.yaml
+++ b/fbfmaproom/fbfmaproom-sample.yaml
@@ -2608,7 +2608,7 @@ countries:
         datasets:
             defaults:
                 predictors:
-                    - pnep-v2
+                    - pnep-v3
                     - rain
                 predictand: bad-years
             observations:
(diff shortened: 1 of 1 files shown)


fbfmaproom/fbfmaproom-sample.yaml
```diff
@@ -502,20 +502,20 @@ countries:
                     lower_is_worse: yes
                     format: number3
                 enso_state: *enso_3mo
-                bad-years-v2:
+                bad-years-ndj-v2:
                     label: Drought humanitarian ranking v2
-                    description: Drought ranking of expert committee humanitarian data review (v2)
-                    path: madagascar/bad-years-djf.zarr
+                    description: Drought ranking of expert committee humanitarian data review (v2) collected on March 2025
+                    path: madagascar/bad-years-ndj-v2.zarr
```
aaron-kaplan, line 508: This looks nonsensical--it says you're replacing a djf zarr with an ndj zarr. Looking at the git history, I think that's because you branched this from a version that was never merged, or was corrected before it was merged. Please close this and start again from master.

It would also be best if you could put the Guatemala change in one PR and the Madagascar one in a different one. That way if there's a problem with one of them (as there is here), it doesn't have to hold up the other.

PR #547, opened 2025-05-02
Fix Djibouti logo; rearrange title and logo to save space

aaron-kaplan:
The Djibouti logo wasn't appearing because the filename in the config file didn't match what we had on the server. (FYI the files on the server are managed in a different repo that only DL team has access to. The plan is to move each country's logo file into that country's python_maproom repo when we install the application on their server.)

While I was at it, I also decided to stack the title on top of the logo to save some space. Before:

image

After:

image


nitinmagima (approved):
Thanks, Aaron, it looks great!

PR #549, opened 2025-05-08
adding v4 for Guatemala-JJA and V3 for Guatemala-ASO

dantesab9:
Here it goes again, updates only for Guatemala tools.


aaron-kaplan:
There's no JJA configuration in this PR. Look at the "Files Changed" tab--it only adds the new ASO.

I'll merge this now; you can submit another PR for JJA.

PR #551, opened 2025-05-15
Splitpepsico

remicousin:
Factored out the things constantly used across python maprooms. Meant to avoid constant copy and paste. Can potentially be moved one level up to apply also to enacts if ever helpful to do so. The fact that there can a control triggers a reinitialization of the map with new set of admin to use makes it a bit messier and the only difference with enacts. Can revisit and clean that up later if need be.

### DIFF
--- a/pepsico/maproom_utilities.py
+++ b/pepsico/maproom_utilities.py
@@ -16,7 +16,7 @@ STD_TIME_FORMAT = "%Y-%m-%d"
 HUMAN_TIME_FORMAT = "%-d %b %Y"
 
 
-def get_geom(level, conf):
+def get_geom(level, conf, shapes_adm_name):
     """ Form a geometric object from sql query or synthetic
 
     Parameters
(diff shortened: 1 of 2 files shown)


pepsico/maproom_utilities.py
```diff
@@ -0,0 +1,413 @@
...
+                "fill": True,
+                "color": adm_color,
+                "weight": adm_weight,
+                "fillOpacity": 0,
+            },
+        ),
+        name=adm_name,
+        id=adm_id,
+        checked=is_checked,
+    )
```
remicousin, line 261: down to here is copied from enacts where we split the geometry-wise functionality. With the addition of the multiple admin sets in pepsico case.

aaron-kaplan (approved):
I skimmed. I don't feel like it's a priority to review carefully.

xchourio (approved):

PR #555, opened 2025-07-11
factoring out common layout functionality

remicousin:
The idea is that layout.py will contain only the specifics of a given app, using building blocks from layout_utilities.py that can be added to, with optional arguments in current functions, or new functions.
Will avoid copy-and-pasting all this stuff from Maproom to Maproom. Probably better engineering if we hand this over to someone.

Could be applied to enacts.

### DIFF
--- a/pepsico/layout_utilities.py
+++ b/pepsico/layout_utilities.py
@@ -1,0 +1,194 @@
+import dash_bootstrap_components as dbc
+from dash import dcc
+from dash import html
+import dash_leaflet as dlf
+
+
+IRI_BLUE = "rgb(25,57,138)"
+IRI_GRAY = "rgb(113,112,116)"
+LIGHT_GRAY = "#eeeeee"
+
+
+def app_layout_1(navbar, description, map, local):
+    """
+    Layout with controls in navbar at the top,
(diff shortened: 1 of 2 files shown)


pepsico/layout_utilities.py
```diff
@@ -0,0 +1,194 @@
...
+import dash_leaflet as dlf
+
+
+IRI_BLUE = "rgb(25,57,138)"
+IRI_GRAY = "rgb(113,112,116)"
+LIGHT_GRAY = "#eeeeee"
+
+
+def app_layout_1(navbar, description, map, local):
+    """
```
xchourio, line 13: Just a comment—when I saw the '_1', I initially thought there were other types of layouts, like a '_2', etc. Later I understood it was more to differentiate it from the main app layout definition. If we’re not going to have others, we could maybe give it a more descriptive name like "head, etc". In any case, it's not crucial—just something to keep in mind in case we need to create other versions of app_layout_* later on.

remicousin:
What else to do?

PepsiCo will likely need additional Python MRs. Now with both RSF awarded, more likely we'll convert forecast MRs (PepsiCo and Climate Group ones) to Python.

If you have visibility on more useful/urgent stuff to do now, please share.

PR #556, opened 2025-07-17
function to count extreme events

remicousin:
I tried to think of a vectorial solution but couldn't. Probably because there is a "memory" aspect of the analysis: once a day is part of an extreme event, it can't be part of another one.

### DIFF
--- a/pepsico/app_calc.py
+++ b/pepsico/app_calc.py
@@ -1,7 +1,6 @@
 import xarray as xr
 import numpy as np
 from scipy.stats import norm
-import operator as op
 
 
 #This is what we should need for the app
@@ -660,7 +659,7 @@ def number_extreme_events_within_days(
     number_extreme_events_within_days(rain_daily_data_in_mm, "gt", 80, 2)    
     """
     count = 0
-    dd = daily_data
+    dd = daily_data.copy()
     #Start with shortest events
     for w in range(1, window+1):
         for t in range(len(dd[dim])-(w-1)):


pepsico/app_calc.py
```diff
@@ -143,3 +144,56 @@ def strftimeb2int(strftimeb):
...
+        for t in range(len(daily_data[dim])-(w-1)):
+            #Assert new event
+            new_event = getattr(op, operator)(
+                daily_data.isel({dim: slice(t, t+w)}).sum(dim), threshold
+            )
+            #Mask days having formed a new event so that they won't account again
+            #for following events (t loop) or longer events (w loop)
+            daily_data[{dim: slice(t, t+w)}] = (
+                daily_data[{dim: slice(t, t+w)}].where(~new_event)
+            )
```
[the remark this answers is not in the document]
remicousin: That's correct, [2nd, 3rd, 4th]  is not an extreme event because [1st, 2nd, 3rd]. Similarly, if all the rain fell in 1st, then  [1st, 2nd, 3rd] is not an event, but [2nd, 3rd, 4th] could be. That's why I had to use loops rather than rolling or something alike.

pepsico/app_calc.py
```diff
@@ -143,3 +144,56 @@ def strftimeb2int(strftimeb):
...
+    #Start with shortest events
+    for w in range(1, window+1):
+        for t in range(len(daily_data[dim])-(w-1)):
+            #Assert new event
+            new_event = getattr(op, operator)(
+                daily_data.isel({dim: slice(t, t+w)}).sum(dim), threshold
+            )
+            #Mask days having formed a new event so that they won't account again
+            #for following events (t loop) or longer events (w loop)
+            daily_data[{dim: slice(t, t+w)}] = (
```
aaron-kaplan: I strongly discourage you from modifying `daily_data`. The caller has no reason to suspect that calling this function will modify the contents of the array he provides as input. At the very least, such behavior would need to be documented, but people don't always read the documentation, so it's better just not to do it.

A less surprising way of implementing this algorithm would be to make a "deep copy" (see https://docs.xarray.dev/en/stable/generated/xarray.DataArray.copy.html) of `daily_data` and modify the copy rather than the original. But I doubt this is the best algorithm anyway, as I'll comment elsewhere.

pepsico/app_calc.py
```diff
@@ -143,3 +144,50 @@ def strftimeb2int(strftimeb):
...
+    count = 0
+    dd = daily_data
+    #Start with shortest events
+    for w in range(1, window+1):
+        for t in range(len(dd[dim])-(w-1)):
+            #Assert new event
+            new_event = dd.isel({dim: slice(t, t+w)}).sum(dim) > threshold
+            #Mask days having formed a new event so that they won't account again
+            #for following events (t loop) or longer events (w loop)
+            dd[{dim: slice(t, t+w)}] = dd[{dim: slice(t, t+w)}].where(~new_event)
```
aaron-kaplan: You're still modifying the input array. `dd` points to the same DataArray object as `daily_data` (see line 183), so when you modify the contents of that object, the change will be visible to the caller.

Changing line 183 to `dd = daily_data.copy(deep=True)` would fix it.

PR #557, opened 2025-07-29
Added OND forecast version 3 for all Ethiopia instances

mm5330:


aaron-kaplan:
Nans shouldn’t cause any outright breakage, or if they do, that can and
should be fixed. If there are regions for which a significant fraction is
masked, I would be concerned that the spatial average (of the remaining
non-Nan pixels) could be misleading.

<https://github.com/iridl/python-maprooms/pull/557#issuecomment-3140515952>
>
>

PR #559, opened 2025-08-04
Update README.md

mm5330:


fbfmaproom/README.md
```diff
@@ -39,17 +39,6 @@ cd python-maprooms/fbfmaproom
 module load python/miniconda3.9.5
```
jturmelle, line 39: This should be python/miniconda3

fbfmaproom/README.md
```diff
@@ -9,7 +9,7 @@ For now, these instructions are specific to developers working on an IRI server.
 Choose port on which to run your development server. You must choose one that isn't in use by another developer on the same server. Log into the server (shortfin01 in the example) and forward your chosen port (1234 in the example).
```
jturmelle, line 9: This should make clear that the port should not be 1234, and it should be > 1000.   There should also be a mechanism to prevent each other from using the same id (for example, everyone using port 1001).   For example, everyone might be able to user their userid as the port number.   I think this would work...

`
port=$(id -u)
ssh -L $port:localhost:$port mako.iri.columbia.edu
`

change shortfin01 to mako

mm5330: that makes sense - my userid isn't a number >1000 (it's 502), would it still work?

mm5330: gotcha. addressed this in my latest commit

jturmelle (approved):
I think it addresses my comments and should work as a step-by-step, but it's still assuming you know a lot of background details.  But that's a different issue.

PR #560, opened 2025-08-04
New versions for JJA (v6) and ASO(v4) added for Guatemala AA

dantesab9:
Hi @aaron-kaplan , I added the new versions of the forecast for Guatemala. I am keeping the interim versions you made to update for the new MeteoFrance in case we keep working on it. The new versions have been limited to 3 months lead time, so some months in both tool instances will be empty in the new versions (that I checked in stagingtoo) , moving forward I wanted to check if you advise to cut the issue months from the tool instance too or keeping them is fine.

### DIFF
--- a/fbfmaproom/fbfmaproom-sample.yaml
+++ b/fbfmaproom/fbfmaproom-sample.yaml
@@ -3094,6 +3094,19 @@ countries:
                         pct: quantile
                     colormap: pne_25
                     is_poe: no
+                pnep-v6:
+                    label: Forecast prob non-exc v6
+                    description: NextGen forecast probability of non-exceedance of selected percentile v6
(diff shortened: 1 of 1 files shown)


aaron-kaplan:
> The new versions have been limited to 3 months lead time, so some months in both tool instances will be empty in the new versions (that I checked in stagingtoo) , moving forward I wanted to check if you advise to cut the issue months from the tool instance too or keeping them is fine.

I suppose it depends on whether we/they still care about the old forecasts that have longer lead times. If they still want to be able to look at those, we need to keep all six leads in the configuration. If they don't, we can reduce it to three and avoid confusion about why some of them are blank.

This branch is forked from an old version. Please check out master, do `git pull`, create a new branch, make and commit your change there, and submit a new pull request.

PR #561, opened 2025-08-04
New versions for JJA (v6) and ASO(v4) added for Guate_AA, new attempt

dantesab9:
Hi @aaron-kaplan , new attempt here. Thanks for the review, I added  the version to be displayed as default this time and seems like there's no conflicts.

### DIFF
--- a/fbfmaproom/fbfmaproom-sample.yaml
+++ b/fbfmaproom/fbfmaproom-sample.yaml
@@ -3444,7 +3444,7 @@ countries:
         datasets:
             defaults:
                 predictors:
-                    - pnep-v4
+                    - pnep-v6
                     - rain
                 predictand: bad-years
             observations:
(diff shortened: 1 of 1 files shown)


aaron-kaplan:
BTW there's a `guatemala-may` config in fbfmaproom-sample.yaml that we never deployed to the public site. Do we still need it?

dantesab9:
Yeah, it is ready for deployment to public. datasets **guatemala/prcp-jja-v6** and **guatemala/prcp-aso-v4** would need to be synced (I can send a separate email for that.) About the guatemala-may config, we don't need it anymore, should I delete it in a new branch? Btw, is it ok if I delete closed branches I opened?

aaron-kaplan:
No need for a separate email. Yes, if you could submit a PR that removes guatemala-may that would be great.

What are we doing with prcp-jja-v5? That's still sitting in fbf-candidate, not synced to fbf.

dantesab9:
Ok I will remove may. Let's keep prcp-jja-v5 a little longer in candidate in case it is brought back by partners.

PR #564, opened 2025-08-13
Local Shapefile Support

nitinmagima:
Local shapefile support - Reads GADM administrative boundaries from local files instead of requiring database connection


aaron-kaplan:
Please review your PR before submitting it. This PR includes your entire data directory, including forecast data in both netcdf and zarr formats.

I'm going to close this PR and delete the branch, otherwise the repo will carry those data files forever, increasing its size. Please submit a PR that changes only the python and yaml files. Don't include shapefiles either. I'll have to figure out a different way to share those.

PR #567, opened 2025-08-27
Spells

remicousin:
Little suite of spell related functions needed by PepsiCo. They all are meant to then feed "seasonal" functions to be applied to multiple years for same season of the year. The core of the word comes from enacts' [longest_run_length](https://github.com/iridl/python-maprooms/blob/64c842a2fa896bdf498d831c3a9b03cf4128a873/enacts/calc.py#L443).

The "public" functions are a bit dummy, but the "private" one definitely should not be callable because it returns something a bit funky. Not sure if there's a better way to engineer this.

Tests produced (also heavily inspired from enacts')

### DIFF
--- a/pepsico/app_calc.py
+++ b/pepsico/app_calc.py
@@ -214,7 +214,7 @@ def daily_tobegroupedby_season(
 
     Parameters
     -----------
-    daily_data : DataArray
+    daily_data : xr.DataArray
         Daily data to be grouped.
     start_day : int
         Day of the start date  of the season.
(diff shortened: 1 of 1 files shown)


aaron-kaplan:
Recording the highlights of our conversation...

`_cumsum_flagged_diff[n]` represents the length of the spell that starts at position `n+1`, except for `n == 0`, where it represents either the length of the spell that starts at position 0 or the spell that starts at position 1. It seems like this loss of information is acceptable if all we need is to find the number and length of spells (there can't be a spell starting at 0 and a different spell starting at 1), but if we ever want to know the dates of the spells, we would have to switch to a different implementation.

An alternative would be padding the sequence with zeros at the left end. The current implementation is an optimization that avoids the cost of padding (= copying) the array.

It might be worth putting some of this into a comment in the code.

PR #595, opened 2026-02-17
Rain events

remicousin:
This is another one for the count of rain events. However, while it seems to be working, I interrupted the app after nearly 1h still spinning to produce a map or a graph. So indeed my algo is too slow. If you want to publish your work in a PR, I could take over and try and find a faster solution.

In the meantime, review here can remain pending until the function is substituted.


remicousin:
This was addressed in another PR.

PR #601, opened 2026-02-25
new implementation of neewd

remicousin:
Aaron: FYI since you contributed to this calculation initially.

New implementation divided the length to run the same test by ~3. If approved, will replace in existing function and see the impact on the app.

Xandre, I know you had worked on an alternative implementation, if you share your work, I can have a look at it.

### DIFF
--- a/pepsico/app_calc.py
+++ b/pepsico/app_calc.py
@@ -113,7 +113,7 @@ def seasonal_data(monthly_data, start_month, end_month, start_year=None, end_yea
 
 def seasonal_wwc(
     labelled_season_data, variable, frost_threshold, wet_threshold, hot_threshold,
-    warm_nights_spell, dry_spell,
+    warm_nights_spell, dry_spell, rain_event_amount, rain_event_window,
 ):
     # Boolean variables need the additional where to return NaNs from False/0 to Nans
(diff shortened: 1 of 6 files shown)


pepsico/app_calc.py
```diff
@@ -738,3 +738,18 @@ def number_extreme_events_within_days(
...
+
+def neewd(daily_data, threshold, window, dim="T"):
+    acc = daily_data.isel({dim: 0}, drop=True)
+    count = (acc > threshold) * 1
+    w = xr.ones_like(acc, dtype=int)
+    for i in range(1, daily_data[dim].size):
+        data_i = daily_data.isel({dim: i}, drop=True)
+        acc_new = (acc + data_i).where(
+            ((acc <= threshold) and (w < window)), other=data_i
+        )
```
xchourio: There are xarrays involved in this operation, to avoid possible 'ambiguous' warnings (or sometimes errors) it is better to use **&** instead of **and**

xchourio (approved):

remicousin:
Alright. First of all, things are working well with pixi and without dask. Then for this specific function, I tried different things but eventually landed on this proposal. It comes out in an acceptable amount of time now. Can do some more performance optimization later in other PRs.

xchourio (approved):

PR #612, opened 2026-03-31
Topixi

remicousin:
FYI for @xchourio as I believe you made a move to pixi for your pepsico app -- but you're not using pingrid I beleive so it's likely another can of worms.

Can be run with 
CONFIG=toto pixi run python app.py
where toto is config-iri.yaml plus your local settings
I only look into the projections app (not wwc) that is lighter in data handling.

Below what I write to Aaron to describe where I'm at and that prompted the share via PR

I'm stalling. I can either get the map or the graph but not both.

To recap...

I could not build the env from scratch adding one package after another so instead I moved on to starting from fbfmaproom toml.

Then I got a bit confused by how to make sure I really refresh the env and in the following, I am systematically removing the lock and .pixi/

Before that though, I did resolved one problem if going to dash prior to 2.4 that had 2 changes in dash.callback_context: the ctx shortcut and how to actually get the triggered id (even though I did otherwise than using the split trick they recommend and just used normal dict manip)

https://dash.plotly.com/determining-which-callback-input-changed

So with the proper removal of all things pixi before testing, I tested 9 configurations:

dash = "2.3.1.*" # from fbf pixi
dash = "2.7.0.*" # from enacts conda lock
dash = "*" # latest

dash-leaflet-iri = "*" # from fbf pixi
dash-leaflet = "0.1.23.*" # from enacts conda lock
dash-leaflet = "*" # latest

It looks like dash-leaflet-iri = "0.1.23"

The graph is wrong if dash is not the latest (3.2.0). The graph draws a line from 1 to n (size of the x-axis) and the Tabs label is not showing. There are recent logged changes in dbc that report on changes to Tabs.
https://www.dash-bootstrap-components.com/changelog/
As I wrote this, I've just tried returning to dbc found in conda lock (1.2.1). It fixed the Tab label but fixed nothing else (and also broke nothing additional).

The map doesn't show if dlf is *; it's grey with a marker if dash is latest (3.2.0). So it works only if dash is 2.7.0 or 2.3.1 and dlf is iri / 0.1.23. But those lower versions of dash break the graph.

There are only errors to be found (and sometimes no error no where) when inspecting elements. I researched them but could not figure anything out of them.

So I am not sure how to proceed next...

### DIFF
--- a/pepsico/maproom_utilities.py
+++ b/pepsico/maproom_utilities.py
@@ -317,8 +317,8 @@ def picked_location(data, initialization_cases, click_lat_lng, latitude, longitu
         dash_ctx_triggered["value"] == None
         or dash_ctx_triggered["prop_id"] in initialization_cases
     ):
-        lat = data["Y"][int(data["Y"].size/2)].values
-        lng = data["X"][int(data["X"].size/2)].values
+        lat = data["Y"].isel(Y=int(data["Y"].size/2), drop=True).values.item()
+        lng = data["X"].isel(X=int(data["X"].size/2), drop=True).values.item()
     else:
         if dash_ctx_triggered["prop_id"] == "map.click_lat_lng":
             lat = click_lat_lng[0]
--- a/pepsico/maproom_utilities.py
+++ b/pepsico/maproom_utilities.py
@@ -312,14 +312,15 @@ def picked_location(data, initialization_cases, click_lat_lng, latitude, longitu
     latitude: str
         Input from latitude pick a point control
     """
+    dash_ctx_triggered = dash.callback_context.triggered[0]
     if (
-        dash.ctx.triggered_id == None
-        or dash.ctx.triggered_id in initialization_cases
+        dash_ctx_triggered["value"] == None
+        or dash_ctx_triggered["prop_id"] in initialization_cases
     ):
         lat = data["Y"][int(data["Y"].size/2)].values
         lng = data["X"][int(data["X"].size/2)].values
     else:
-        if dash.ctx.triggered_id == "map":
+        if dash_ctx_triggered["prop_id"] == "map.click_lat_lng":
             lat = click_lat_lng[0]
             lng = click_lat_lng[1]
         else:
--- a/pepsico/pixi.lock
+++ b/pepsico/pixi.lock
@@ -1,0 +1,3880 @@
+version: 6
+environments:
+  default:
+    channels:
+    - url: https://conda.anaconda.org/conda-forge/
+    - url: https://conda.anaconda.org/iridl/
+    options:
+      pypi-prerelease-mode: if-necessary-or-explicit
+    packages:
+      linux-64:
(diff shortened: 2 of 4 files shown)


remicousin:
I'm stalling. I can either get the map or the graph but not both.

To recap...

I could not build the env from scratch adding one package after another so instead I moved on to starting from fbfmaproom toml.

Then I got a bit confused by how to make sure I really refresh the env and in the following, I am systematically removing the lock and .pixi/

Before that though, I did resolved one problem if going to dash prior to 2.4 that had 2 changes in dash.callback_context: the ctx shortcut and how to actually get the triggered id (even though I did otherwise than using the split trick they recommend and just used normal dict manip)
https://dash.plotly.com/determining-which-callback-input-changed

So with the proper removal of all things pixi before testing, I tested 9 configurations:

dash = "2.3.1.*" # from fbf pixi
dash = "2.7.0.*" # from enacts conda lock
dash = "*" # latest

dash-leaflet-iri = "*" # from fbf pixi
dash-leaflet = "0.1.23.*" # from enacts conda lock
dash-leaflet = "*" # latest

It looks like dash-leaflet-iri = "0.1.23"

The graph is wrong if dash is not the latest (3.2.0). The graph draws a line from 1 to n (size of the x-axis) and the Tabs label is not showing. There are recent logged changes in dbc that report on changes to Tabs.
https://www.dash-bootstrap-components.com/changelog/
As I wrote this, I've just tried returning to dbc found in conda lock (1.2.1). It fixed the Tab label but fixed nothing else (and also broke nothing additional).

The map doesn't show if dlf is *; it's grey with a marker if dash is latest (3.2.0). So it works only if dash is 2.7.0 or 2.3.1 and dlf is iri/0.1.23. But those lower versions of dash break the graph.

There are only errors to be found (and sometimes no error no where) when inspecting elements. I researched them but could not figure anything out of them.

So I am not sure how to proceed next...

I run it with CONFIG=toto pixi run python app.py

in pepsico folder with toto adapted version of config-dev-sample and tested only the projection app.

aaron-kaplan:
I installed with no package version constraints, and I see the graph but not the map. There's an error in the log that indicates a problem with the interface between pandas and the db. That problem was the reason I had pinned pandas. The solution is to go through SQLAlchemy instead of using psycopg2 directly. psycopg2 is only used in one place (sql2geom), so this shouldn't be too hard. I'll give it a shot.

That may not fix everything, particularly the tabs, but I'll try one thing at a time.

remicousin:
Let me know if you want me to take over from there (transitioning to
Alchemy)


*The Data Library will soon shut down. Click here
<https://iri.columbia.edu/resources/data-library/sunset/> to learn more.*

Rémi Cousin (he/him/...)
***@***.***
Phone: +1 8456804422
Fax: +1 8456804864
http://iri.columbia.edu
61 route 9W, Monell Building
Palisades, NY 10964

PR #613, opened 2026-04-16
Tempterc

remicousin:
Adapting the Precip Maproom case to Temp. Adding a config entry for that. Adding Aaron in review just because I am not quite sure this is how the config is meant to be passed to layout...


xchourio (approved):

aaron-kaplan:
> Adding Aaron in review just because I am not quite sure this is how the config is meant to be passed to layout...

It's ok.

PR #614, opened 2026-04-17
Pepsico pixi

aaron-kaplan:
This seems to work. There are now lots of OpenCV errors in the log. I'll look into that next week.

### DIFF
--- a/pepsico/layout_utilities.py
+++ b/pepsico/layout_utilities.py
@@ -151,7 +151,7 @@ def map(
                     ),
                 ],
                 id="map",
-                center=None,
+                center=(0,0),
                 zoom=default_zoom,
                 style={"width": "100%", "height": "50vh"},
             ), type="dot"),
--- a/pepsico/pixi.toml
+++ b/pepsico/pixi.toml
@@ -1,0 +1,21 @@
+[workspace]
+authors = ["Aaron Kaplan <aaron@iri.columbia.edu>"]
+channels = ["conda-forge"]
+name = "pepsico"
+platforms = ["linux-64"]
+version = "0.1.0"
+
+[tasks]
+
+[dependencies]
+dash-leaflet = ">=1.0.15,<2"
+xarray = ">=2026.4.0,<2027"
+opencv = ">=4.13.0,<5"
+psycopg2 = ">=2.9.11,<3"
+rasterio = ">=1.5.0,<2"
+shapely = ">=2.1.2,<3"
+pyyaml = ">=6.0.3,<7"
+dash-bootstrap-components = ">=2.0.4,<3"
+scipy = ">=1.17.1,<2"
+zarr = ">=3.1.6,<4"
+sqlalchemy = ">=2.0.49,<3"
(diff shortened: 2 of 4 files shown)


pepsico/layout_utilities.py
```diff
@@ -151,7 +151,7 @@ def map(
                     ),
                 ],
                 id="map",
-                center=None,
+                center=(0,0),
```
aaron-kaplan, line 154: Pick a better default if you know of one.

pepsico/pixi.toml
```diff
@@ -0,0 +1,21 @@
+[workspace]
+authors = ["Aaron Kaplan <aaron@iri.columbia.edu>"]
```
aaron-kaplan: I usually change this to `"IRI <help@iri.columbia.edu>"`, but maybe it's time to pick something else.

aaron-kaplan:
I may have lost the plot here. Was the goal to remove dask? I included it because the application wouldn't start without it. What is the symptom that we thought might be caused by dask?

remicousin:
I dug up the email that triggered the plot.

aaron-kaplan:
Thanks. I think having an up-to-date pixi environment would be good regardless of what happens on that email thread, so I suggest you review/test this PR now, and we can make further modifications later as needed.

remicousin (approved):

aaron-kaplan:
@xchourio ok with you?

pepsico/pixi.toml
```diff
@@ -0,0 +1,21 @@
...
+authors = ["IRI <help@iri.columbia.edu>"]
+channels = ["conda-forge"]
+name = "pepsico"
+platforms = ["linux-64"]
+version = "0.1.0"
+
+[tasks]
+
+[dependencies]
+dash-leaflet = ">=1.0.15,<2"
```
xchourio, line 11: The minimum version I require is 1.1.3, which is not available on conda-forge, but in any case, if I install it from PyPI it works:

[pypi-dependencies]
dash-leaflet = ">=1.1.3,<2"
dash-bootstrap-components = "==2.0.4"
dash-extensions = "==2.0.4"
dash-auth = ">=2.3.0,<3"

xchourio (approved):

pepsico/pixi.toml
xchourio: in the worst-case scenario—if the other apps must be below 1.1.3—we can still separate the apps (which is what I currently do).

aaron-kaplan: It looks like dash-leaflet updates haven't been published to conda-forge since January 2024. The conda package is managed by a conda-forge community member, not by the author of dash-leaflet, and he probably lost interest.

> in the worst-case scenario—if the other apps must be below 1.1.3—we can still separate the apps (which is what I currently do).

Not sure what you mean by "the other apps." I think it would be best if the Pepsico apps used a common environment. I don't think they need to be in the same environment as ENACTS or CCSR.

@remicousin @xchourio can you work together to find a configuration that works for all the pepsico apps? Let me know if you would like me to merge this PR now, or if you want to change it before merging.

remicousin: I would rather you merge this and we'll take it from there. Xandre's app is, to-date, not part of the gitHub suite for some reason so there's little to say about harmonization as of now.

PR #615, opened 2026-04-21
adapted to weekly subseasonal fcst and applied to precip

remicousin:
The PR is called "bi" because I thought I would start with the biweekly but the biweekly data files are in home jyuan so I don't have an easy way to know where the data files are exactly. So  I did the weekly instead. Some of the dates manipulation for subseasonal were already written for the CPT versions.

### DIFF
--- a/ccsr/config-defaults.yaml
+++ b/ccsr/config-defaults.yaml
@@ -29,11 +29,23 @@ maprooms:
           forecast_path: /Data/data21/iri/FD/NMME_Seasonal_Forecast/Precipitation_ELR
           variable: Precipitation
           var: precip
+          forecast: Seasonal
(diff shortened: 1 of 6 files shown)


xchourio:
Just a comment for the biweekly: is there any reason the data is in the home directory? If there’s an operational script associated with it, it’s best for it to be in a standard path. Even though the path are define and read it from the YAML, I’m concerned it could become a snowball.

xchourio (approved):

remicousin:
Even for weekly, the data is not really in data/jyuan: these are symbolic links to crunch. All those files are generated by Jing. They should care better and I am tired to care in their place. A priori, all these are going to go to forecast.ccsr if they want to use the python-DL to share the data. I don't think I want to deal with a temporary transition to dlupdates.

PR #619, opened 2026-04-28
Ccsrpixi

remicousin:
First 2 commits reproduce Aaron's work to convert PepsiCo to Pixi. 3rd commit reproduce my work to adapt PepsiCo to pixi environment (minus the parts related to data reading). The environment is the exact same as PepsiCo (so no dask).

Symptom at this stage is that tiles fail with a index out of bounds error (always the same regardless of what triggers the error):
IndexError: index 3 is out of bounds for axis 1 with size 3

As I was seeking the problem, a print resolved the issue (see last commit). Before I got there, the error was triggered by idxmax in terc_fcst/maproom.py line 237. Same error occurred with argmax, with max when trying to recompute idxmax by other means, printing fcst_ds["prob"] right before idxmax.

Run with:

CONFIG=config-iri.yaml:myconfig.yaml pixi run python app.py
and visit
python_maproom/terc_fcst_precip/
Comment out the print to see the tiles show up, and break it back by removing the print (or just its .data)

Note that the local graph works (default location is in the ocean but even if the tiles don't show, you can click on land and have the expected graph).


aaron-kaplan:
The reason the print statement changes the behavior is that `.data` causes the array data to be loaded in from disk. Replacing the print with `fcst_ds["prob"].load()` has the same effect.

By trial and error, moving the call to `load` to different places, I found that if I put it just before `fcst_ds = fcst_ds.sel(lead=int(lead_time))` (terc_fcst_calc.py line 53) the tile generation call succeeds, and if I put it just after that statement, it fails.

This looks like a bug in xarray. If you can extract a simple, self-contained program that demonstrates the bug without requiring someone to set up the entire application, we can report it.

remicousin:
Thanks. I moved up optional L selection based on past experience that sooner, the faster the app, and redid the trial and fail to see where to add .load().

aaron-kaplan:
I said the .load has to come before the .sel. I realize that results in loading more data than you need, but if the .sel comes first then you hit the bug, so the choice is between loading too much data and fixing the bug in xarray.

remicousin:
Yes but since I moved the .sel, that's where I found the .load() need be to prevent the bug.

aaron-kaplan:
I see. I misunderstod--thought you were saying "I fail to see where to add .load()."

That's interesting. I think it means the bug is not in .sel itself, but in some interaction between .sel and one of the other operations. I think it would be worthwhile to spend a little time trying to reproduce this in a self-contained program.

remicousin:
I will try.

remicousin:
Aaron already chipped in to identify what might be an xarray bug. Otherwise, same changes as for pepsico conversion to pixi.

Aaron: review as you wish.

remicousin:
> I see. I misunderstod--thought you were saying "I fail to see where to add .load()."
> 
> That's interesting. I think it means the bug is not in .sel itself, but in some interaction between .sel and one of the other operations. I think it would be worthwhile to spend a little time trying to reproduce this in a self-contained program.

Should the self-contained program relies on the env set up here? or an env with just the latest xarray?

aaron-kaplan:
Just the latest xarray, ideally.

remicousin:
@aaron-kaplan it seems I ran into the fix of the problem. While trying to isolate and reproduce the bug, I ran right away into xr.opend_dataset failing to open a file, not finding the engine to apply. Adding netcf4 package passed the reading. In ccsr's pixi, netcdf4 was not installed but scipy was. Xarray [doc](https://docs.xarray.dev/en/stable/user-guide/io.html#netcdf) reports that scipy engine supports only nc3. Adding nc4 as package in pixi, engine in xr.open_dataset, and removing the load() commands behaves.

I presume that with scipy engine, xr.open_dataset was not failing but was not doing the right thing. Possibly there was a warning but I might never have seen it as there are many warnings in the pixi version.

Finally, the same Xarray doc suggests that h5 might be faster than nc4? Shall I try that or bad idea?

aaron-kaplan:
Good find. Regarding h5, I say if it ain't broke don't fix it. IIUC what speed problems you have are related to calculation, not reading data from disk.

xchourio:
What does your myconfig.yaml file look like? I couldn’t find it among the files, so I thought I could use the sample by adjusting the port. Even though I can deploy part of the maproom, it fails with errors like _S = get_issues(fcst_conf)[0].strftime(S_format) IndexError: list index out of range_, so I must be missing the path configuration adjustment somewhere... I’m testing it from the forecast.ccsr server

remicousin:
myconfig.yaml is a copy of the sample with my specifics. The problem is that the data is not on forecast.ccsr, it's on our own /Data so you need to run this on mako.

xchourio:
tested on mako ..... : At runtime I got header incompatibility warnings and the map did not render. After some investigation, it seems that the NumPy version is too new and incompatible with the precompiled binaries used by netCDF4. I tried downgrading the version, but then the issue was that the .nc file being read is “classic” (version 3), for which the compatible reader is “scipy”. I adjusted that and it worked correctly. I also tested selecting several points and the response time was good.

If a clean installation is done from this pixi.toml, 
are  you be able to run the maproom without the previously reported warnings and issues?

remicousin:
I am confused: my very last commit is to exactly move away from scipy (v3) to nc4...??

xchourio (approved):

remicousin:
So @xchourio is that still not working for you with the pixi/toml here?

PR #620, opened 2026-05-06
Added JAS V5 for Djibouti

nitinmagima:
Added JAS V5 for Djibouti and made it the default forecast.


aaron-kaplan:
This has been deployed.

PR #623, opened 2026-06-29
enabling and applying reading geom from shp

remicousin:
This works ok as long as the level 1 admin is not involved. If it is, the app just works on it forever. I am creating a PR nonetheless because I am not sure I will have time to get back to it. Simplify didn't help and actually made things worse. So possibly the solution is to find another source that is smaller.

### DIFF
--- a/pepsico/README.md
+++ b/pepsico/README.md
@@ -2,42 +2,20 @@
 
 # Installation and Run Instructions
 
-## Creating a conda environment with this project's dependencies
+## Get Shapefiles
 
-* see enacts' README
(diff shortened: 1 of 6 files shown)


xchourio (approved):
I'm going to approve it............ It reminded me of a similar issue I had with the counties. In that case, using a lighter shapefile solved the problem.

Later I'll try to explore a possible solution using shapefiles from Natural Earth (https://www.naturalearthdata.com/). However, I'd need to check how switching datasets would affect the shapefile's internal definitions (IDs, attributes, etc.).

For example, the Natural Earth states/provinces shapefile is about 14 MB:
https://www.naturalearthdata.com/downloads/10m-cultural-vectors/10m-admin-1-states-provinces/

whereas the FAO version is around 200 MB.

PR #624, opened 2026-08-19
Added Yemen forecasts to the config. Commented out bad years for now since the data is not ready.

nitinmagima:
Added Yemen forecasts to the config for five seasons. Commented out bad years for now since the data is not ready.


aaron-kaplan (commented):
I have no technical comments, but a couple procedural ones.

AFAIK we've never used CPC-CMAP before in in AA configuration. Every new dataset we introduce incrementally increases the maintenance burden for the eventual manager of the data library. There should be a process for deciding whether a new dependency is worth it or not.

(The [NeededDS spreadsheet](https://docs.google.com/spreadsheets/d/1clypBjkoeyJ4gmWbR4WoWlfdt8j5HhAIvS1_kXsHC0M/edit?gid=0#gid=0) does happen to list it, but as a dependency of Jing's forecast process, not AA. If there's actually an existing AA configuration that uses it, please point it out.)

Secondly, this is an unprecedented number of seasons, which means a lot of forecast configurations and a lot of data syncing. Do we really need to take on that much operational burden?

DOsgoodCU:
This has gotten more complex as we've been having a few people hustling to do stuff, with a lot of the classic "across IRI" communications needing to happen.  Lets discuss where we stand on thursday and start planning a more focused strategy.

> I have no technical comments, but a couple procedural ones.
> 
> AFAIK we've never used CPC-CMAP before in in AA configuration. Every new dataset we introduce incrementally increases the maintenance burden for the eventual manager of the data library. There should be a process for deciding whether a new dependency is worth it or not.
> 
> (The [NeededDS spreadsheet](https://docs.google.com/spreadsheets/d/1clypBjkoeyJ4gmWbR4WoWlfdt8j5HhAIvS1_kXsHC0M/edit?gid=0#gid=0) does happen to list it, but as a dependency of Jing's forecast process, not AA. If there's actually an existing AA configuration that uses it, please point it out.)
> 
> Secondly, this is an unprecedented number of seasons, which means a lot of forecast configurations and a lot of data syncing. Do we really need to take on that much operational burden?

aaron-kaplan:
Conclusion from today's meeting: leave in CPC and all the seasons for the initial demo, and we'll prune later.

aaron-kaplan:
This has been deployed.

